PR #101, opened 2026-05-25
fix(issue): crash/: unbounded EPIPE crash logs bloat ~/.gsd by millions of files (21GB+)

jeremymcs:
## Summary
- Fixed recoverable EPIPE crash logging and bounded crash-log file creation with focused tests and typecheck passing.

## Bugs Addressed
- [x] **Recoverable EPIPE events write crash logs**
- [x] **Crash logs use unbounded per-call filenames**

## Verification
- Completed in the repository worktree before push.

## Related Issue
- Closes #99
- [#99 crash/: unbounded EPIPE crash logs bloat ~/.gsd by millions of files (21GB+)](https://github.com/open-gsd/gsd-pi/issues/99)

## User
- @mattiaverio

## Repo
- `open-gsd/gsd-pi`

## Branch
- `issue/99-crash-unbounded-epipe-crash-logs-bloat-g-1779675494`

Issue #99, opened 2026-05-25
crash/: unbounded EPIPE crash logs bloat ~/.gsd by millions of files (21GB+)


mattiaverio:
## Problem

`~/.gsd/crash/` grows unbounded because `handleRecoverableExtensionProcessError` calls `writeCrashLog` on every EPIPE signal before marking it as recoverable and returning. EPIPE fires routinely when stdout's pipe closes at session end. At high session frequency this creates **millions of files** — one user observed 1.82M files / 21GB.

## Root Cause

**`bootstrap/register-extension.js:25`** — `writeCrashLog` is called for EPIPE before the recoverable-return path:

```js
export function handleRecoverableExtensionProcessError(err) {
    if (err.code === "EPIPE") {
        writeCrashLog(err, "EPIPE");            // ← writes a NEW FILE every EPIPE
        const stdoutGone = process.stdout.destroyed || process.stdout.writableEnded;
        if (stdoutGone) {
            process.exit(0);
        }
        process.stderr.write(`[gsd] swallowed EPIPE ...`);
        return true;   // ← marked as recoverable — crash log was never needed
    }
    ...
}
```

**`bootstrap/crash-log.js:19`** — each call generates a unique millisecond-precision filename:

```js
const ts = new Date().toISOString().replace(/[:.]/g, "-");
const logPath = join(crashDir, `${ts}.log`);   // ← one new file per call
```

EPIPE is a recoverable, high-frequency signal (stdout pipe closed = normal session end). Writing a crash log for it is incorrect by definition — the function marks it `return true` (handled/recoverable) immediately after. A burst of EPIPE events at session teardown (e.g. when Claude Code closes its pipe to GSD) creates thousands of files in milliseconds.

## Expected Behavior

EPIPE should **not** call `writeCrashLog` at all — it is a recoverable, expected signal. The fix at `register-extension.js:24-25` is to remove the `writeCrashLog(err, "EPIPE")` call:

```js
if (err.code === "EPIPE") {
    // EPIPE is recoverable — no crash log needed
    const stdoutGone = process.stdout.destroyed || process.stdout.writableEnded;
    if (stdoutGone) {
        process.exit(0);
    }
    process.stderr.write(`[gsd] swallowed EPIPE (syscall=${err.syscall ?? "?"})\n`);
    return true;
}
```

[...]

PR #164, opened 2026-05-26
fix(bootstrap): exit on EPIPE storm instead of swallowing in a tight loop

evolv3ai:
Follow-up to #101 (issue #99).

## Context

#101 fixed the disk-fill side of the EPIPE crash-storm — EPIPE no longer writes crash artifacts, and other crashes append to one `pid-<pid>.log`. That stops the 681 GB / 6.8M-file flood. 👍

## Remaining gap

The EPIPE branch still has no loop bound:

```ts
if (err.code === "EPIPE") {
  const stdoutGone = process.stdout.destroyed || process.stdout.writableEnded;
  if (stdoutGone) process.exit(0);
  process.stderr.write(`[gsd] swallowed EPIPE (...)\n`);
  return true;          // ← swallow forever if the flags never flip
}
```

When a downstream reader dies but `stdout.destroyed` / `writableEnded` never flip (common with broken pipes in headless/piped invocations), every subsequent write throws EPIPE and is swallowed again — a tight, progress-free CPU spin. The disk no longer fills, but the process is wedged.

## Fix

Add a storm guard: if EPIPE fires **> 100 times within 10 s**, the pipe is gone for good → `process.exit(0)`. Also route stderr writes through a `safeStderr()` wrapper, because `process.stderr.write` can itself throw EPIPE and re-enter this handler (re-looping). Applied to the EPIPE/EIO/ENOENT branches.

36-line change, one file, no behavior change for healthy pipes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-pi/pr/164"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith</code> with what you need. Autofix is enabled.</sup>

### DIFF
--- a/src/resources/extensions/gsd/bootstrap/register-extension.ts
+++ b/src/resources/extensions/gsd/bootstrap/register-extension.ts
@@ -27,6 +27,32 @@ import { initCmuxEventListeners } from "../../cmux/index.js";
 
 export { writeCrashLog } from "./crash-log.js";
 
+// Pipe-closed storm guard. #99/#101 stopped EPIPE from flooding ~/.gsd/crash,
+// but a persistently-broken output pipe whose `destroyed`/`writableEnded` flags
+// never flip is still swallowed on every write — a tight, progress-free CPU
+// spin. If the pipe-closed error fires in a tight loop the pipe is gone for
+// good; exit cleanly instead.
+const EPIPE_STORM_THRESHOLD = 100;
+const EPIPE_STORM_WINDOW_MS = 10_000;
+let epipeCount = 0;
+let epipeWindowStart = 0;
+
+/** Write to stderr without ever re-throwing — stderr can EPIPE too, which would
+ *  re-enter this handler and re-loop. */
+function safeStderr(msg: string): void {
+  try {
+    process.stderr.write(msg);
+  } catch { /* stderr is also broken; nothing we can do */ }
+}
+
+/** A peer closing the read end of a pipe mid-write surfaces differently per
+ *  platform: POSIX throws `EPIPE`; Windows throws `Error: write EOF` (or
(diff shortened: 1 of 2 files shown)


evolv3ai:
Extended the storm guard to also cover the Windows pipe-closed encodings reported in #181.

`handleRecoverableExtensionProcessError` previously matched only POSIX `EPIPE`. On Windows the same logical condition (peer closed the read end of a pipe mid-write) surfaces as `Error: write EOF` / `read EOF` with no `code` set, and a reset socket-backed pipe throws `ECONNRESET` — all now recognized via a `pipeClosedLabel()` helper and routed through the same swallow + storm-exit path. Existing EPIPE log strings are unchanged; matching is on the exact `"write EOF"`/`"read EOF"` message to avoid eating unrelated errors. Added tests for write EOF / read EOF / ECONNRESET, an over-match negative case, and a Windows-EOF storm.

This addresses **part 1** of #181 (recoverable-error recognition) — the Windows `write EOF` after a task completes is now swallowed instead of escaping to the uncaught-exception path. Parts 2 (run lock cleanup before `process.exit(1)` in the uncaught guards) and 3 (register `SIGBREAK` in `auto-supervisor.js`) remain as separate follow-ups, since they touch supervisor lock coordination rather than this storm guard.

jeremymcs (approved):

Issue #141, opened 2026-05-26
[Bug]: Orphaned queued milestone (gsd_milestone_generate_id race + isReusableGhostMilestone too strict) causes endless research-milestone re-dispatch


m1dm4n:
### Bug type

Incorrect behavior or output

### Summary

`gsd_milestone_generate_id` can leave an orphaned `queued` DB row that has no on-disk artifacts. The auto-mode dispatcher then re-dispatches `research-milestone` for that orphan forever, even after a sibling milestone (with the same title) has been fully completed.

### Steps to reproduce

1. `unique_milestone_ids: true` in `PREFERENCES.md`.
2. Start auto-mode at the milestone boundary (`research-milestone` / `discuss-milestone`).
3. Call `gsd_milestone_generate_id` once during the discuss flow — it inserts `M0NN` as `queued` with no content files.
4. The discuss flow re-invokes `gsd_milestone_generate_id` a second time (e.g. after a `/gsd auto` restart that re-enters discuss). Because `unique_milestone_ids` is on, a fresh ID like `M0NN-veveb9` is allocated; the queued `M0NN` row is never reclaimed.
5. Plan + execute + complete `M0NN-veveb9` end-to-end.
6. On the next iteration boundary, observe `orchestrator-dispatch-match research-milestone M0NN` re-fire on the still-empty `M0NN`.

### Expected behavior

When the discuss/plan flow re-runs `gsd_milestone_generate_id` and there is already a `queued`-status DB row for the same conceptual milestone that has zero content files and no worktree, GSD should either:

- (a) Reclaim the existing queued ID instead of allocating a new one (treat queued-empty as reusable in `gsd_milestone_generate_id`'s entry path, not only via `isReusableGhostMilestone`), **or**
- (b) Garbage-collect orphaned `queued` rows when a sibling milestone with the same title reaches `complete`, **or**
- (c) Have `auto-dispatch.js` skip dispatching `research-milestone` for any milestone whose directory does not exist on disk (currently it only checks for the absence of `RESEARCH.md`, not the absence of the milestone directory).

[...]

PR #244, opened 2026-05-29
fix: burn-max cost-spike opt-out (#142) and DeepSeek reasoning_content replay (#209)

### LINKED ISSUE #142: [Bug]: token_profile burn-max cannot opt out of hardcoded 3× rolling-average cost-spike pause
### Bug type

Incorrect behavior or output

### Summary

Auto-mode pauses on a "unit cost spike" (`unit ≥ 3× rolling average`) with no way to opt out — neither `token_profile: burn-max` nor an absent `budget_usd` preference disables the 3× multiplier, and there is no preference key that turns the check off.

### Steps to reproduce

1. Set `token_profile: burn-max` in `.gsd/PREFERENCES.md`. Leave `budget_usd` and `per_unit_cost_cap_usd` unset.
2. Run `/gsd auto` on a milestone with a wide cost distribution across unit types (e.g. a research-heavy slice followed by an execute-heavy one).
3. Wait for any single `execute-task`, `complete-slice`, or `discuss-milestone` to exceed `3.0 × rolling-average-for-that-unit-type` cost.
4. Auto-mode emits `Unit ... cost spike detected ($X.XX vs avg $Y.YY) — pausing auto-mode.` and stops.

### Expected behavior

`burn-max` profile is documented as "Quality-first profile: keep user-selected models, disable downgrade routing" (`preferences-models.js:444-450`). Users who explicitly opt into burn-max have communicated that they do not want cost-based pausing. The cost-spike guard should:

- Skip the 3× rolling-average pause entirely when `prefs.token_profile === "burn-max"`, **or**
- Accept a `prefs.unit_cost_spike_multiplier` preference (e.g. set to `0` or `Infinity` to disable, or a higher multiplier like `10`), **or**
- At minimum, only fire when `budget_usd` is set — using the 3× rolling-average heuristic as a budget-protection mechanism when no budget exists is a defensive over-reach.

Concrete minimal fix: in `auto-budget.js:33-41`, return `"none"` early when the active profile is `burn-max`, **or** plumb a `multiplier` value from `prefs` through both call sites (`auto-post-unit.js:1520`, `auto-verification.js:667`) instead of hardcoding `3.0`.

### Actual behavior

The guard is hardcoded:

```js
// auto-budget.js:33-41
export function getUnitCostSpikeAction(unitCostUsd, rollingAvgUsd, multiplier = 3.0) {
    if (!Number.isFinite(unitCostUsd) || unitCostUsd < 0) return "none";
    if (!Number.isFinite(rollingAvgUsd) || rollingAvgUsd <= 0) return "none";
    if (!Number.isFinite(multiplier) || multiplier <= 0) return "none";
    return unitCostUsd >= (rollingAvgUsd * multiplier) ? "pause" : "none";
}
```

Both call sites pass the literal `3.0`:

```js
// auto-post-unit.js:1520 (the one that actually pauses)
if (getUnitCostSpikeAction(unitCostUsd, rollingAvgUsd, 3.0) === "pause") {
    ...
    ctx.ui.notify(`Unit ${s.currentUnit.id} cost spike detected (${unitCostUsd.toFixed(2)} vs avg ${rollingAvgUsd.toFixed(2)}) — pausing auto-mode.`, "error");
    await pauseAuto(ctx, pi);
    return "dispatched";
}

// auto-verification.js:667 (emits warning only; included for completeness)
if (getUnitCostSpikeAction(unitCostUsd, rollingAvgUsd, 3.0) === "pause") {
    ctx.ui.notify(`...`, "warning");
}
```

`preferences-models.js:344, 376-379, 444-450` defines `burn-max` as a tier-map opt-out:

```js
const VALID_TOKEN_PROFILES = new Set(["budget", "balanced", "quality", "burn-max"]);
...
// burn-max intentionally omits a tier map: it never writes model defaults
// (it preserves the user's explicit model selection), so resolveProfileDefaults
// skips model resolution for this profile.
"burn-max": {},
```

Nothing about `burn-max` touches the cost-spike guard. The user has explicitly chosen a "burn the budget" profile but is still paused by the cost guard. There is no preference key to disable or tune the multiplier.

### GSD version

1.0.2

### Node.js version

v24.14.1

### Operating system

Ubuntu 24.04.4 LTS

### Install method

npm global install

### Affected area

Auto-mode / dispatch loop

### Model and provider

anthropic/claude-sonnet-4-6, anthropic/claude-opus-4-7

```
# Cost-spike anomalies in a single 92-hour project history (selected):
$46.24  vs $3.31 avg execute-task  (14.0×)  → paused
$56.36  vs $3.31 avg execute-task  (17.0×)  → paused
$68.60  vs $3.31 avg execute-task  (20.7×)  → paused
$135.59 vs $3.31 avg execute-task  (40.9×)  → paused
$6.12   vs $1.13 avg complete-slice (5.4×)  → paused
$9.60   vs $2.07 avg discuss-milestone (4.6×) → paused

# Preferences in effect during these pauses:
token_profile: burn-max
budget_usd: <unset>
per_unit_cost_cap_usd: <unset>

# Surfaced as user-visible error notifications:
"Unit execute-task/M0NN/S03/T06 cost spike detected ($40.12 vs avg $3.31) — pausing auto-mode."

# Code references (all in @opengsd/gsd-pi extension source):
auto-budget.js:33-41         getUnitCostSpikeAction — hardcoded 3.0 default, no profile check
auto-post-unit.js:1520       caller hardcoded 3.0, calls pauseAuto on spike
auto-verification.js:667     caller hardcoded 3.0 (warning only — fine to leave)
preferences-models.js:344    VALID_TOKEN_PROFILES includes "burn-max"
preferences-models.js:376-379 burn-max profile config — empty tier map
preferences-models.js:444-450 burn-max default — dynamic_routing disabled, nothing re: cost
```

### Impact

Severity: blocks `burn-max` users from running long auto-mode sessions unattended
Frequency: ~every 5-10 units on a project with diverse unit-type cost distributions
Consequence: defeats the purpose of `burn-max` (which advertises "quality-first, keep user-selected models" but in practice still blocks high-quality models from being used freely); user must `/gsd auto` continue after every pause, sometimes >10× per milestone

### Additional information

Workarounds (all unsatisfying):
- Set `per_unit_cost_cap_usd` to a very high value (e.g. $999) — this is actually a different code path (`auto-post-unit.js:1512`), so it does not affect the spike guard at all.
- There is no `unit_cost_spike_multiplier` preference, no `disable_cost_spike_guard` preference, and no documented way to turn the guard off.

Closely related but distinct from #33 / #39 (synchronous `pauseAuto()` race) — those bugs concern *spurious* pauses on freshly-dispatched units. This bug concerns a *legitimate-by-policy* pause that should not fire under the user's chosen profile.

---
*Auto-generated by `/gsd forensics`. Paths and code snippets refer to `@opengsd/gsd-pi` extension source only; no user project code or secrets included.*

### LINKED ISSUE #209: DeepSeek via OpenRouter/non-deepseek.com baseUrl 400s: reasoning_content field name not enforced on replay
## Problem

When the active model routes DeepSeek through OpenRouter (or any non-`deepseek.com` baseUrl), the provider rejects every follow-up request with:

```
400 Provider returned error
{"error":{"message":"The `reasoning_content` in the thinking mode must be passed back to the API.","type":"invalid_request_error",…}}
```

The unit gets stuck in a dispatch loop because every retry hits the same wall.

## Root Cause

Two coupled defects in `packages/pi-ai/src/providers/openai-completions.ts`.

### Defect 1 — reasoning field name round-trip uses the inbound stream's name

Streamer at **`openai-completions.ts:222-243`** picks the first non-empty reasoning field name as it appears on the wire and stores it verbatim as `thinkingSignature`:

```ts
const reasoningFields = ["reasoning_content", "reasoning", "reasoning_text"];
let foundReasoningField: string | null = null;
for (const field of reasoningFields) {
    if ((choice.delta as any)[field] !== null && (choice.delta as any)[field].length > 0) {
        if (!foundReasoningField) { foundReasoningField = field; break; }
    }
}
…
currentBlock = { type: "thinking", thinking: "", thinkingSignature: foundReasoningField };
```

Replayer at **`openai-completions.ts:593-597`** then writes the assistant message back using *that same* field name:

```ts
const signature = nonEmptyThinkingBlocks[0].thinkingSignature;
if (!compat.stripReasoningContent && signature && signature.length > 0) {
    (assistantMsg as any)[signature] = nonEmptyThinkingBlocks.map((b) => b.thinking).join("\n");
}
```

OpenRouter normalizes DeepSeek streaming chunks to use `reasoning` (its canonical name). So `thinkingSignature = "reasoning"` → next request sends `assistantMsg.reasoning = "…"` → DeepSeek backend behind OpenRouter requires `reasoning_content` → **400**.

DeepSeek-direct "happens to work" because the upstream stream uses `reasoning_content` and the same field name round-trips intact. There is no flag controlling the outgoing field name — only `stripReasoningContent` which removes it entirely.

### Defect 2 — provider name "deepseek" is ignored by `isNonStandard` detection

**`openai-completions.ts:761-770`**:

```ts
const isNonStandard =
    provider === "cerebras" || baseUrl.includes("cerebras.ai") ||
    provider === "xai"      || baseUrl.includes("api.x.ai") ||
    baseUrl.includes("chutes.ai") ||
    baseUrl.includes("deepseek.com") ||
    isZai ||
    provider === "opencode" || baseUrl.includes("opencode.ai");
```

Every other entry checks both `provider === "<name>"` *and* a baseUrl substring. DeepSeek only checks the baseUrl. The block-comment two lines up (line 752: *"Provider takes precedence over URL-based detection since it's explicitly configured"*) is therefore incorrect for DeepSeek: a model registered with `provider: "deepseek"` but a non-`deepseek.com` baseUrl (OpenRouter, Vercel AI Gateway, custom proxy) is treated as a standard OpenAI completions endpoint, gets `supportsStore: true` + `supportsDeveloperRole: true` (both wrong), and never enters the DeepSeek-specific code path.

## Expected Behavior

1. Reasoning replay should use the field name **the destination provider expects**, not whatever the inbound stream emitted. Either:
   - Add a new compat knob `replayReasoningAs?: "reasoning_content" | "reasoning"` to `OpenAICompletionsCompat`, defaulting via auto-detect, *or*
   - When the destination is known to be DeepSeek, force `"reasoning_content"` regardless of inbound field name.

2. `isNonStandard` detection should treat `provider === "deepseek"` as DeepSeek regardless of baseUrl, mirroring how `cerebras`/`xai`/`opencode` are detected.

### Suggested patch (illustrative)

```ts
// Around line 761
const isDeepSeek =
    provider === "deepseek" ||
    baseUrl.includes("deepseek.com") ||
    (baseUrl.includes("openrouter.ai") && model.id.startsWith("deepseek/")) ||
    (baseUrl.includes("gateway.ai.vercel") && model.id.includes("deepseek"));

const isNonStandard =
    provider === "cerebras" || baseUrl.includes("cerebras.ai") ||
    provider === "xai"      || baseUrl.includes("api.x.ai") ||
    baseUrl.includes("chutes.ai") ||
    isDeepSeek ||
    isZai ||
    provider === "opencode" || baseUrl.includes("opencode.ai");

// Then thread `replayReasoningAs: isDeepSeek ? "reasoning_content" : undefined`
// through the compat object, and use it at line 595:
const fieldName = compat.replayReasoningAs ?? signature;
if (!compat.stripReasoningContent && fieldName && fieldName.length > 0) {
    (assistantMsg as any)[fieldName] = nonEmptyThinkingBlocks.map((b) => b.thinking).join("\n");
}
```

### Workaround for users hit by this today

Override `compat` on the model registration to strip reasoning entirely (sacrifices cross-turn thinking context but unsticks the loop):

```ts
{
  provider: "openrouter",
  baseUrl: "https://openrouter.ai/api/v1",
  id: "deepseek/deepseek-v4-pro-…",
  compat: { stripReasoningContent: true },
}
```

## Environment

- GSD version: 1.0.2
- Failing provider/model: `openrouter` / `deepseek/deepseek-v4-pro-20260423`
- Triggered after failover from `openai-codex / gpt-5.5` by `auto-timeout-recovery.ts`
- Unit: `gate-evaluate/M015-…/S21/gates+Q3,Q4,Q8` (also reproduces on `execute-task` and `complete-slice` units)

## Reproduction Context

1. Configure a model with `provider: "openrouter"` (or `provider: "deepseek"`), baseUrl `https://openrouter.ai/api/v1`, model id `deepseek/deepseek-…`.
2. Make a request that produces a thinking block (reasoning model with non-trivial prompt).
3. Continue the conversation (multi-turn tool-use, e.g. a GSD gate-evaluate unit).
4. Second request 400s with `reasoning_content … must be passed back to the API`.

## Forensic Evidence

- Stream record carries `"thinkingSignature":"reasoning"` (not `"reasoning_content"`).
- Same unit redispatched 3+ times in a single auto-mode session (stuck loop).
- Same symptom observed across multiple units in the same auto-mode run (78 stuck-loop anomalies in the affected session; the deepseek-via-openrouter units are the ones that never recover).

---

m1dm4n:
## TL;DR

**What:** Two independent auto-mode/provider bug fixes — `burn-max` can't opt out of the cost-spike pause (#142), and DeepSeek via OpenRouter 400s on reasoning replay (#209).
**Why:** Both stall auto-mode — #142 pauses `burn-max` users on every cost spike; #209 traps DeepSeek units in a dispatch loop.
**How:** A preferences-resolved cost-spike multiplier (`Infinity` for `burn-max`), and replaying DeepSeek reasoning under `reasoning_content`.

## What

One commit per concern:

- **`fix(gsd)` — #142:** `auto-budget.ts` gains `resolveUnitCostSpikeMultiplier(prefs)`; both call sites (`auto-post-unit.ts` pause path, `auto-verification.ts` warning path) thread it instead of the hardcoded `3.0`. A new `unit_cost_spike_multiplier` preference is registered, typed on `GSDPreferences`, and validated.
- **`fix(pi-ai)` — #209:** `openai-completions.ts` replays assistant reasoning under `reasoning_content` when the resolved compat requires it, recognises a DeepSeek model routed through OpenRouter (`deepseek/…`) or the Vercel AI Gateway as needing that replay, and wires `provider: "deepseek"` into `isNonStandard`.

## Why

- **#142:** `burn-max` is documented as the "burn the budget" profile, but the `3× rolling-average` spike guard was hardcoded with no opt-out, repeatedly pausing users who explicitly chose it.
- **#209:** OpenRouter normalises DeepSeek's reasoning to the `reasoning` field; the replay wrote it back under that name, but the DeepSeek backend requires `reasoning_content` → `400 … reasoning_content … must be passed back` → endless re-dispatch.

Closes #142
Closes #209

## How

- **#142** reuses the existing non-finite-multiplier guard in `getUnitCostSpikeAction` (no signature change): `burn-max` resolves to `Infinity` → `"none"`; an explicit positive `unit_cost_spike_multiplier` overrides; otherwise `3.0`.
- **#209** decouples the *reasoning-replay* requirement (broad: includes gateway-routed DeepSeek) from the *thinking-format / endpoint* flags (narrow: direct DeepSeek only), so OpenRouter-routed models keep their `reasoning: { effort }` format while still getting `reasoning_content` on replay.

## Tests

- **#142** — `auto-budget-alerts.test.ts`: `burn-max` disables the pause, an explicit multiplier is honoured, and the default `3×` still pauses. (Fails before the fix on the missing export; passes after.)
- **#209** — `openai-completions-deepseek-reasoning.test.ts`: reasoning replays under `reasoning_content` for DeepSeek and the captured field is preserved otherwise. Verified failing before the fix (`expected '' to be 'internal chain'`) and passing after. The full pi-ai vitest suite was run to confirm no detection regression (the one regression this surfaced — OpenRouter reasoning format — was fixed by the broad/narrow decoupling above).

## Change type

- [x] `fix` — Bug fix

## Notes

- A third issue I filed, **#141** (orphaned queued milestone re-dispatch), is intentionally **not** in this PR. Its fix touches the `isReusableGhostMilestone` invariant that guards against #4996, so it warrants a dedicated, carefully-tested change; I'll follow up separately.
- **AI-assisted:** this change was developed with AI assistance. I have reviewed it, understand it, and verified the tests fail before / pass after each fix.

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-pi/pr/244"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1782659968&installation_id=134682257&pr_number=244&repository=open-gsd%2Fgsd-pi&return_to=https%3A%2F%2Fgithub.com%2Fopen-gsd%2Fgsd-pi%2Fpull%2F244&signature=d120d743a1c444a1a558a45493fd62552cade3275de484b2d65f7c84911d1589"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith</code> with what you need. Autofix is disabled.</sup>

### DIFF
--- a/packages/pi-ai/src/providers/openai-completions.ts
+++ b/packages/pi-ai/src/providers/openai-completions.ts
@@ -853,6 +853,12 @@ export function convertMessages(
 					if (model.provider === "opencode-go" && signature === "reasoning") {
 						signature = "reasoning_content";
 					}
+					// DeepSeek (including via OpenRouter, which normalises the inbound reasoning field
+					// to "reasoning") requires reasoning to be replayed under "reasoning_content";
+					// otherwise the backend rejects the next request with a 400.
+					if (compat.requiresReasoningContentOnAssistantMessages && signature === "reasoning") {
+						signature = "reasoning_content";
+					}
 					if (signature && signature.length > 0) {
 						(assistantMsg as any)[signature] = nonEmptyThinkingBlocks.map((block) => block.thinking).join("\n");
 					}
@@ -1077,6 +1083,19 @@ function detectCompat(model: Model<"openai-completions">): ResolvedOpenAIComplet
 	const isMoonshot = provider === "moonshotai" || provider === "moonshotai-cn" || baseUrl.includes("api.moonshot.");
 	const isCloudflareWorkersAI = provider === "cloudflare-workers-ai" || baseUrl.includes("api.cloudflare.com");
 	const isCloudflareAiGateway = provider === "cloudflare-ai-gateway" || baseUrl.includes("gateway.ai.cloudflare.com");
+	// Direct DeepSeek: provider "deepseek" or the deepseek.com endpoint. These own the
+	// non-standard endpoint flags (supportsStore/supportsDeveloperRole) and the "deepseek"
+	// thinking format.
+	const isDeepSeek = provider === "deepseek" || baseUrl.includes("deepseek.com");
+	// A DeepSeek model routed through a gateway that rewrites the model id (OpenRouter
+	// "deepseek/...", Vercel AI Gateway "...deepseek...") is still served by a DeepSeek backend
+	// that requires reasoning to be replayed under `reasoning_content` on the next request.
+	// The gateway — not DeepSeek — owns the request/thinking format, so ONLY this replay flag
+	// applies to the gateway case (not isNonStandard or thinkingFormat).
+	const requiresReasoningContent =
+		isDeepSeek ||
+		(baseUrl.includes("openrouter.ai") && model.id.startsWith("deepseek/")) ||
+		(baseUrl.includes("gateway.ai.vercel") && model.id.includes("deepseek"));
 
 	const isNonStandard =
 		provider === "cerebras" ||
@@ -1085,7 +1104,7 @@ function detectCompat(model: Model<"openai-completions">): ResolvedOpenAIComplet
 		baseUrl.includes("api.x.ai") ||
 		isTogether ||
 		baseUrl.includes("chutes.ai") ||
-		baseUrl.includes("deepseek.com") ||
+		isDeepSeek ||
 		isZai ||
 		isMoonshot ||
 		provider === "opencode" ||
@@ -1096,7 +1115,6 @@ function detectCompat(model: Model<"openai-completions">): ResolvedOpenAIComplet
 	const useMaxTokens = baseUrl.includes("chutes.ai") || isMoonshot || isCloudflareAiGateway || isTogether;
 
 	const isGrok = provider === "xai" || baseUrl.includes("api.x.ai");
-	const isDeepSeek = provider === "deepseek" || baseUrl.includes("deepseek.com");
 	const cacheControlFormat = provider === "openrouter" && model.id.startsWith("anthropic/") ? "anthropic" : undefined;
 
 	return {
@@ -1108,7 +1126,7 @@ function detectCompat(model: Model<"openai-completions">): ResolvedOpenAIComplet
 		requiresToolResultName: false,
 		requiresAssistantAfterToolResult: false,
 		requiresThinkingAsText: false,
-		requiresReasoningContentOnAssistantMessages: isDeepSeek,
+		requiresReasoningContentOnAssistantMessages: requiresReasoningContent,
 		thinkingFormat: isDeepSeek
 			? "deepseek"
 			: isZai
--- a/packages/pi-ai/test/openai-completions-deepseek-reasoning.test.ts
+++ b/packages/pi-ai/test/openai-completions-deepseek-reasoning.test.ts
@@ -1,0 +1,106 @@
+import { describe, expect, it } from "vitest";
+import { convertMessages } from "../src/providers/openai-completions.ts";
+import type { AssistantMessage, Context, Model, OpenAICompletionsCompat, Usage } from "../src/types.ts";
+
+const emptyUsage: Usage = {
+	input: 0,
+	output: 0,
+	cacheRead: 0,
+	cacheWrite: 0,
+	totalTokens: 0,
+	cost: { input: 0, output: 0, cacheRead: 0, cacheWrite: 0, total: 0 },
+};
+
+// Full Required compat as detectCompat produces it for DeepSeek: replay reasoning
+// under reasoning_content. Mirrors the fixture shape used by the sibling
+// openai-completions tests in this directory.
+const deepseekCompat = {
+	supportsStore: false,
+	supportsDeveloperRole: false,
+	supportsReasoningEffort: true,
+	supportsUsageInStreaming: true,
+	maxTokensField: "max_completion_tokens",
+	requiresToolResultName: false,
+	requiresAssistantAfterToolResult: false,
+	requiresThinkingAsText: false,
+	requiresReasoningContentOnAssistantMessages: true,
+	thinkingFormat: "deepseek",
+	openRouterRouting: {},
+	vercelGatewayRouting: {},
+	zaiToolStream: false,
+	supportsStrictMode: true,
+	cacheControlFormat: undefined,
+	sendSessionAffinityHeaders: false,
+	supportsLongCacheRetention: true,
+} satisfies Required<Omit<OpenAICompletionsCompat, "cacheControlFormat">> & {
+	cacheControlFormat?: OpenAICompletionsCompat["cacheControlFormat"];
+};
+
+// Same wire shape, but the provider does NOT require reasoning_content (e.g. a
+// vanilla OpenAI-compatible endpoint): the captured field name must be preserved.
+const standardCompat = {
+	...deepseekCompat,
+	requiresReasoningContentOnAssistantMessages: false,
+	thinkingFormat: "openai" as const,
+};
+
+function buildDeepSeekModel(): Model<"openai-completions"> {
+	return {
+		id: "deepseek/deepseek-chat",
+		name: "DeepSeek Chat",
+		api: "openai-completions",
+		provider: "deepseek",
+		baseUrl: "https://openrouter.ai/api/v1",
+		reasoning: true,
+		input: ["text"],
+		cost: { input: 0, output: 0, cacheRead: 0, cacheWrite: 0 },
+		contextWindow: 128000,
+		maxTokens: 4096,
+	};
+}
+
+// OpenRouter normalises DeepSeek's reasoning to the "reasoning" field, so that is
+// the signature captured during streaming and carried into the next request.
+function buildContextWithCapturedReasoning(): Context {
+	const assistant: AssistantMessage = {
+		role: "assistant",
+		content: [
+			{ type: "thinking", thinking: "internal chain", thinkingSignature: "reasoning" },
+			{ type: "text", text: "Hi" },
+		],
+		api: "openai-completions",
+		provider: "deepseek",
+		model: "deepseek/deepseek-chat",
+		usage: emptyUsage,
+		stopReason: "stop",
+		timestamp: 2,
+	};
+	return {
+		messages: [
+			{ role: "user", content: "hello", timestamp: 1 },
+			assistant,
+			{ role: "user", content: "continue", timestamp: 3 },
+		],
+	};
+}
+
+describe("openai-completions DeepSeek reasoning_content replay (#209)", () => {
+	it("replays reasoning under reasoning_content when the captured field was 'reasoning'", () => {
+		const messages = convertMessages(buildDeepSeekModel(), buildContextWithCapturedReasoning(), deepseekCompat);
+		const assistant = messages.find((m) => m.role === "assistant") as Record<string, unknown> | undefined;
+		expect(assistant).toBeDefined();
+		// The reasoning text must land in reasoning_content (what DeepSeek requires on replay),
+		// not the inbound "reasoning" field that OpenRouter happened to use.
+		expect(assistant?.reasoning_content).toBe("internal chain");
+		expect(assistant?.reasoning).toBeUndefined();
+	});
+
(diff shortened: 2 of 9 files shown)


jeremymcs:
@m1dm4n is this ready for review ?

m1dm4n:
Yes — ready. ✅

**Scope:** two independent fixes — #142 (`burn-max` / `unit_cost_spike_multiplier` opt-out of the 3× cost-spike pause) and #209 (replay DeepSeek reasoning under `reasoning_content` on non-`deepseek.com` baseUrls, incl. OpenRouter/Vercel gateway).

**Synced against latest main:** both issues still open — main hasn't touched either path. The two shared files (`auto-post-unit.ts`, `auto-verification.ts`) changed on main only for unrelated retry/pause-message fixes; both still call `getUnitCostSpikeAction(…, 3.0)` with the hardcoded literal, so this change is still needed and merges clean (0 conflicts). No redundant lines to drop.

**CI:** 13/13 green. Tests added for both paths.

The 🔴 CRITICAL flag is the heuristic firing on any Auto Engine touch — verified manually: burn-max → `Infinity` → existing non-finite guard yields `"none"` (loosens the pause, never blocks loop termination), and the DeepSeek field coercion is gated on `requiresReasoningContentOnAssistantMessages` so other providers are unaffected.

226 commits behind main but merges cleanly — happy to rebase if you'd prefer it current.

PR #479, opened 2026-06-04
fix: apply tool result truncation in all modes, not just auto-mode

gianlorenzomungiovino:
## Problem

In interactive/manual mode, tool results were never truncated because the truncation logic was gated behind `isAutoActive()`.

With a local model (llama-cpp), the flow `streamViaClaudeCode() → buildPromptFromContext()` pastes everything into XML without truncation, and the `before_provider_request` hook applied truncation **only** when `isAutoActive() === true`.

Result: tool results consumed the entire context window (53.5k tokens = 42% in the screenshot below).

Context Breakdown - tool results at 53.5k tokens (42%)

## Solution

- Extract preferences loading outside the `isAutoActive()` guard
- Move **truncation** logic to run in all modes (universal)
- Keep **observation masking** auto-mode-only (intentional — it is specific to auto-mode behavior)

## Testing

This is a pure control-flow reorganization — the truncation code itself is unchanged, only *when* it executes is different. No new behavior, just a less restrictive guard condition.

## Diff

```diff
- if (isAutoActive()) {
-   // ... truncation inside isAutoActive()
- }
+ // truncation always applied
+ if (cmConfig) { ... }
+ 
+ // observation masking only in auto-mode
+ if (isAutoActive() && cmConfig) { ... }
```


jeremymcs:
@gianlorenzomungiovino thanks for the PR. For future, please open an issue and tag it.

PR #480, opened 2026-06-04
fix(auto): run-latency quick wins — stalled-tool timeout, resume thrash, skill bloat, switch noise

jeremymcs:
## TL;DR

**What:** Four targeted engine fixes that remove avoidable latency from auto-mode runs — a dedicated hung-tool timeout, a resume-replay guard, slice-closeout skill trimming, and suppression of bogus provider-switch warnings.
**Why:** Diagnosing a slow run in `test-apps/456` (a trivial single-file app) surfaced these as concrete, independently-fixable latency/noise sinks.
**How:** Each is a small, localized change with its own regression test; the two larger architectural levers are intentionally deferred.

## What

- **R2 — Dedicated hung-tool timeout.** Adds `auto_supervisor.stalled_tool_timeout_minutes` (default **5**) and evaluates the in-flight-tool check on its own budget, independent of the general idle gate. Files: `auto-timers.ts`, `preferences-types.ts`, `preferences-models.ts`, `config-overlay.ts`.
- **R5 — Resume replay guard.** When a paused-session resume has no recorded unit identity, skip the full tool-call replay and resume from rebuilt disk state. File: `auto.ts`.
- **R7 — Slice-closeout skill allowlist.** `complete-slice` gets a focused closeout allowlist instead of the full ~49-skill catalog. File: `skill-manifest.ts` (the `unit-context-manifest` policy auto-derives). `execute-task` deliberately left wildcard per RFC #4779.
- **R8 — Provider-switch noise.** Only notify the provider-switch observer on a genuine cross-API switch. File: `packages/pi-ai/src/providers/transform-messages.ts`.
- Tests: `auto-supervisor.test.mjs`, `interrupted-session-auto.test.ts`, `provider-switch-observer.test.ts`, `skill-manifest.test.ts`.

## Why

A small milestone in `test-apps/456` ran far slower than it should. Diagnosis isolated several engine-level causes, of which these four are the low-risk quick wins:

- The idle watchdog reused `idle_timeout_minutes` to judge a single in-flight tool as hung, so a stuck tool ran out the full idle window (~58 min observed) before recovery fired (**R2**).
- On resume with no recorded unit identity, recovery fell back to a literal `"unknown"` unit — unverifiable and un-targetable — then synthesized a full tool-call replay that re-injected an unbounded blob into the fresh context and re-overflowed it (**R5**).
- `complete-slice` injected the full skill catalog into every closeout (**R7**).
- Callers omitting `sourceApi` defaulted `fromApi` to the target api, so within-provider synthetic tool-result backfills fired as `Provider switch X → X` warnings — 88% of notifications in the sample run, burying real signals (**R8**).

The two deeper levers (auto-advance between phases; retry-ceiling + context compaction on resume) are intentionally **out of scope** here and tracked for follow-up.

## How

- **R2:** In `startUnitSupervision`, the in-flight-tool branch now uses `stalledToolTimeoutMs` and runs before the general idle gate, so a hung tool is caught in minutes while a tool progressing within budget still counts as real progress. New config field is additive with a default; setting it to `0` falls back to the idle window.
- **R5:** `handlePausedSessionResumeRecovery` returns early (skipping replay) when neither the live unit nor persisted paused-unit identity is present; disk state has already been rebuilt by `rebuildState` + doctor before this point.
- **R7:** Adds a `complete-slice` entry to `UNIT_TYPE_SKILL_MANIFEST`; `unit-context-manifest` derives its skills policy from the resolver, so both stay consistent (covered by existing test #4782).
- **R8:** Guards `notifyProviderSwitchObserver` on `report.fromApi !== report.toApi`. Existing cross-API observer tests are unaffected; a new test covers the same-API suppression path.

## Change type

- [x] `fix` — Bug fix
- [ ] `feat` / `refactor` / `docs` / `chore`

## Breaking changes

None. R2 adds one **optional** preference (`auto_supervisor.stalled_tool_timeout_minutes`, default 5) — fully backward-compatible. R8 changes notification behavior only (suppresses non-actionable same→same warnings); no public API, CLI, or file-format change.

## Notes

- Touches auto-mode (`auto-timers.ts`, `auto.ts`) — flagging per the RFC-for-core-systems guideline; these are localized bug fixes rather than architectural changes, but happy to convert to an issue/ADR first if maintainers prefer.
- Bundled as one cohesive "run-latency quick wins" set from a single diagnosis; can split into per-fix PRs on request.
- This PR is AI-assisted (Claude Code). Changes were built, tested, and smoke-run against a source-dev runtime before submission.

---

> [!NOTE]
> **Medium Risk**
> Changes auto supervision, resume recovery, and pi-ai message transform notification paths; behavior is well-tested but affects long-running auto sessions and provider telemetry.
> 
> **Overview**
> Four targeted **auto-mode latency** fixes: hung tools, resume thrash, closeout prompt bloat, and provider-switch telemetry noise.
> 
> **Hung tools** get a dedicated **`stalled_tool_timeout_minutes`** (default **5m**, separate from idle). The idle watchdog judges in-flight tools on that budget, refreshes progress for tools still within budget, and only then applies the general idle gate—so stuck tools recover in minutes instead of riding the full idle window.
> 
> **Paused resume** no longer falls back to an `"unknown"` unit when pause metadata lacks unit identity; it **skips tool-call replay** and continues from rebuilt disk state after doctor/rebuild.
> 
> **`complete-slice`** now uses a focused **skill allowlist** (like milestone closeout) instead of the full catalog; **`execute-task`** stays wildcard per RFC #4779.
> 
> **`transformMessages`** only notifies the provider-switch observer when **`fromApi !== toApi`**, so same-provider synthetic tool-result backfills no longer flood telemetry.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 336c2a9ab26dfe04d9dda480f557d410efd0ba8a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/pi-ai/src/providers/transform-messages.ts
+++ b/packages/pi-ai/src/providers/transform-messages.ts
@@ -281,7 +281,14 @@ export function transformMessagesWithReport<TApi extends Api>(
 	// If the conversation ends with unresolved tool calls, synthesize results now.
 	insertSyntheticToolResults();
 
-	if (hasReportChanges(report)) {
+	// Only surface a provider-switch report when the source and target APIs
+	// actually differ. Within-API transforms — most notably synthetic
+	// tool-result backfills inserted when a same-provider conversation ends on
+	// an unresolved tool call — are not cross-provider data loss and must not be
+	// reported as a "provider switch". Callers that omit `sourceApi` default
+	// `fromApi` to the target api, so without this guard every such call emits a
+	// spurious same→same report that floods telemetry and buries real switches.
+	if (hasReportChanges(report) && report.fromApi !== report.toApi) {
 		notifyProviderSwitchObserver(report);
 	}
 
--- a/src/resources/extensions/gsd/auto-timers.ts
+++ b/src/resources/extensions/gsd/auto-timers.ts
@@ -147,6 +147,15 @@ export function startUnitSupervision(sctx: SupervisionContext): void {
   const softTimeoutMs = supervisionTimeouts.softTimeoutMs;
   const idleTimeoutMs = supervisionTimeouts.idleTimeoutMs;
   const hardTimeoutMs = supervisionTimeouts.hardTimeoutMs;
+  // A single hung tool gets its own short budget, NOT the general idle window:
+  // a long-but-progressing session is not idle, but a tool stuck for minutes
+  // is. Falls back to the idle window only if misconfigured to zero. The
+  // hung-tool budget is intentionally not scaled by task estimate — a stuck
+  // tool call is stuck regardless of how long the overall task should take.
+  const stalledToolTimeoutMs =
+    (supervisor.stalled_tool_timeout_minutes ?? 0) > 0
+      ? supervisor.stalled_tool_timeout_minutes! * 60 * 1000
+      : idleTimeoutMs;
 
   // ── 1. Soft timeout warning ──
   s.wrapupWarningHandle = setTimeout(() => {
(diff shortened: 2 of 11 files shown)


jeremymcs:
**Local `verify:merge` triage** (per CONTRIBUTING): ran the full gate locally — **10059 tests pass, including all new tests in this PR**. Two failures surfaced, both confirmed **pre-existing on `main` (75048e7d)** and environmental to my macOS/Node-26 box, unrelated to this change:

- `repository-registry › uses external-state worktree checkout as project root` — `/tmp` vs `/private/tmp` realpath mismatch (macOS tmpdir symlink).
- `doctor-scope-db-unavailable › checkEngineHealth treats explicit reopen as authoritative when dispatch timestamps are missing` — `table sqlite_master may not be modified` in the db-adapter harness.

Both reproduce identically on a clean `main` checkout with this branch's changes stashed, so they are not regressions. Relying on GitHub CI (Linux) as the authoritative gate for these two.

PR #473, opened 2026-06-04
fix(issue): Auto-mode re-dispatches zero-tool-call units with no cap, wasting ~4 billed prompts per stuck unit (phases.js:2055)

jeremymcs:
## Summary
- Capped zero-tool-call retries in `phases.ts` at `MAX_ZERO_TOOL_RETRIES=1` using a new per-unit `zeroToolRetryCount` map in `AutoSession`, pausing auto-mode on exhaustion instead of falling through to the 5-dispatch `CONSECUTIVE_SAME_UNIT_CAP`; verified with a new test asserting `action: "break"` on the second occurrence.

## Verification
- Completed in the repository worktree before push.

## Related Issue
- Closes #472
- [#472 Auto-mode re-dispatches zero-tool-call units with no cap, wasting ~4 billed prompts per stuck unit (phases.js:2055)](https://github.com/open-gsd/gsd-pi/issues/472)

## User
- @mattiaverio

## Repo
- `open-gsd/gsd-pi`

## Branch
- `issue/472-auto-mode-re-dispatches-zero-tool-call-u-1780575239`

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-pi/pr/473"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783167608&installation_id=134682257&pr_number=473&repository=open-gsd%2Fgsd-pi&return_to=https%3A%2F%2Fgithub.com%2Fopen-gsd%2Fgsd-pi%2Fpull%2F473&signature=a8667b4d9fea5794901e754815a1b09c9fc7e99f095499f634de270fb8cabc0d"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>
<sup>Need help on this PR? Tag <code>/codesmith</code> with what you need. Autofix is disabled.</sup>

---

> [!NOTE]
[...]

Issue #469, opened 2026-06-04
Guided flow under claude-code falsely aborts: workflow-MCP surface tools re-validated against pi.getActiveTools()


mashpie:
## Problem
Guided flow (`/gsd`) under the `claude-code` provider aborts every unit with a false
`cannot run … this unit requires <tools>` error, even though the GSD workflow MCP tools
are present (delivered as `mcp__gsd-workflow__*`). Reproduced on a fresh project with
`discuss-milestone`:

```
Provider "claude-code" cannot run guided flow for discuss-milestone: this unit requires
gsd_summary_save, gsd_requirement_save, gsd_requirement_update, gsd_plan_milestone,
gsd_milestone_generate_id, ...
```

## Root Cause
`extensions/gsd/workflow-mcp.ts` → `getWorkflowTransportSupportError`, the missing-tool
computation (compiled `workflow-mcp.js:463-466`):

```js
const uniqueRequired = [...new Set(requiredTools)];
const missing = (options.activeTools && options.activeTools.length > 0)
    ? uniqueRequired.filter((tool) => !hasRequiredTool(tool, options.activeTools))
    : uniqueRequired.filter((tool) => !MCP_WORKFLOW_TOOL_SURFACE.has(tool));
```

For the `claude-code` transport (authMode `externalCli`, baseUrl `local://`,
`usesWorkflowMcpTransport()` → true), GSD workflow tools are delivered to the spawned
`claude` CLI subprocess via a project-scoped MCP config
(`workflow-mcp-auto-prep.ts` → `ensureProjectWorkflowMcpConfig`). They are therefore NOT
present in `pi.getActiveTools()`, which only lists the pi host harness tools. Because that
host list is non-empty, the `activeTools` branch is taken and `hasRequiredTool()`
(`workflow-mcp.js:423`) matches none of the GSD tools, so all required tools are flagged
missing → false abort (`workflow-mcp.js:470`).

The static fallback branch (line 466) checks `MCP_WORKFLOW_TOOL_SURFACE`
(`workflow-mcp.js:21-65`), which already contains all five `discuss-milestone` tools —
proving the transport satisfies them. Only the `activeTools` branch regresses.

All three guided/auto callers pass `activeTools: pi.getActiveTools()`:
- `guided-flow.js:890`
- `auto/phases.js:1726`
- `auto-direct-dispatch.js:234`

## Expected Behavior
[...]

PR #471, opened 2026-06-04
fix(issue): Guided flow under claude-code falsely aborts: workflow-MCP surface tools re-validated against pi.getActiveTools()

jeremymcs:
## Summary
- Added regression test for discuss-milestone guided-flow false-abort (issue #469); the underlying `piRuntimeRequired` fix was already applied in workflow-mcp.ts, so only the dedicated test was missing.

## Verification
- Completed in the repository worktree before push.

## Related Issue
- Closes #469
- [#469 Guided flow under claude-code falsely aborts: workflow-MCP surface tools re-validated against pi.getActiveTools()](https://github.com/open-gsd/gsd-pi/issues/469)

## User
- @mashpie

## Repo
- `open-gsd/gsd-pi`

## Branch
- `issue/469-guided-flow-under-claude-code-falsely-ab-1780572840`

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-pi/pr/471"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783165454&installation_id=134682257&pr_number=471&repository=open-gsd%2Fgsd-pi&return_to=https%3A%2F%2Fgithub.com%2Fopen-gsd%2Fgsd-pi%2Fpull%2F471&signature=d9f0778e711ce45230bae5504bb47368c64a831e0e6d5368ddc32c72b03feae0"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>
<sup>Need help on this PR? Tag <code>/codesmith</code> with what you need. Autofix is disabled.</sup>

---

> [!NOTE]
> **Low Risk**
> Test-only change with no runtime behavior modified in this PR.
> 
> **Overview**
[...]

PR #481, opened 2026-06-04
test(gsd): add zero-tool retry state to session mocks

jeremymcs:
## TL;DR

**What:** Keeps GSD auto-session test mocks and workflow-MCP regression coverage aligned with current runtime contracts.
**Why:** PR #473 added `zeroToolRetryCount`, and the merged workflow-MCP tests also needed to reflect the guided-flow dispatch path after the active-tool preflight tightening.
**How:** Adds the missing retry map to hand-rolled session mocks and updates the discuss-milestone regression test to rely on the configured workflow MCP launch surface instead of a parent-session active-tool fixture.

## What

Updated the session helper objects in `custom-engine-loop-integration.test.ts` and `journal-integration.test.ts` so they include `zeroToolRetryCount`.

Updated the discuss-milestone workflow-MCP regression test so it matches guided flow's actual compatibility check: guided flow starts workflow MCP as part of dispatch, so the parent session active-tool list is not authoritative for MCP-surface tools.

## Why

PR #473 clears `zeroToolRetryCount` on successful artifact verification. A couple of integration tests use plain object session mocks instead of `AutoSession`, so those mocks must track the same field or the auto-loop can throw and retry instead of completing.

After the run-uat active-tool preflight tightened, the #469 guided-flow regression test was still passing a parent active-tool fixture directly into the helper. The production guided-flow caller no longer does that, because workflow MCP tools are launched for the dispatched turn.

## How

The mock fix mirrors the production session shape by adding `new Map<string, number>()` for `zeroToolRetryCount` alongside the existing dispatch, recovery, and verification counters.

The workflow-MCP test fix preserves the auto-mode coverage that rejects genuinely tool-starved active sessions, while keeping the guided-flow regression focused on the launch-configured workflow MCP surface.

## Verification

- `pnpm run verify:fast`
- `pnpm run test:compile`
- `node --import ./scripts/dist-test-resolve.mjs --experimental-test-isolation=process --test-reporter=spec --test-name-pattern "discuss-milestone guided flow does not abort" --test dist-test/src/resources/extensions/gsd/tests/workflow-mcp.test.js`
- `node --import ./scripts/dist-test-resolve.mjs --experimental-test-isolation=process --test-reporter=./scripts/test-reporter-compact.mjs --test dist-test/src/resources/extensions/gsd/tests/workflow-mcp.test.js`
- `node --import ./scripts/dist-test-resolve.mjs --experimental-test-isolation=process --test-reporter=./scripts/test-reporter-compact.mjs --test dist-test/src/resources/extensions/gsd/tests/custom-engine-loop-integration.test.js dist-test/src/resources/extensions/gsd/tests/journal-integration.test.js`

## Change type

- [ ] `fix` — Bug fix

---

> [!NOTE]
> **Low Risk**
> Changes are limited to test helpers and one regression test assertion; no production runtime behavior is modified in this diff.
> 
> **Overview**
> Keeps GSD auto-loop integration test session mocks in sync with production **`AutoSession`** by adding **`zeroToolRetryCount`** (`Map<string, number>`) next to the other per-unit counters, so successful paths that clear that map no longer throw on plain-object mocks.
> 
> Updates the **`discuss-milestone`** workflow MCP regression test: it no longer passes a parent **`activeTools`** list that omits workflow tools, and the comment states guided flow starts the workflow MCP server during dispatch, so the parent session tool list should not gate MCP tool availability.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit df6d78f4bdf93db6050750b495e5f2b48a29082d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/resources/extensions/gsd/tests/custom-engine-loop-integration.test.ts
+++ b/src/resources/extensions/gsd/tests/custom-engine-loop-integration.test.ts
@@ -129,6 +129,7 @@ function makeLoopSession(overrides?: Record<string, unknown>) {
     unitLifetimeDispatches: new Map<string, number>(),
     unitRecoveryCount: new Map<string, number>(),
     verificationRetryCount: new Map<string, number>(),
+    zeroToolRetryCount: new Map<string, number>(),
     gitService: null,
     autoStartTime: Date.now(),
     activeEngineId: null,
--- a/src/resources/extensions/gsd/tests/journal-integration.test.ts
+++ b/src/resources/extensions/gsd/tests/journal-integration.test.ts
@@ -200,6 +200,7 @@ function makeSession() {
     unitLifetimeDispatches: new Map<string, number>(),
     unitRecoveryCount: new Map<string, number>(),
     verificationRetryCount: new Map<string, number>(),
+    zeroToolRetryCount: new Map<string, number>(),
     gitService: null,
     autoStartTime: Date.now(),
     cmdCtx: {
--- a/src/resources/extensions/gsd/tests/workflow-mcp.test.ts
+++ b/src/resources/extensions/gsd/tests/workflow-mcp.test.ts
@@ -852,9 +852,8 @@ test("transport compatibility still blocks units whose MCP tools are not exposed
 });
 
 test("discuss-milestone guided flow does not abort when all required tools are on MCP surface (regression #469)", () => {
-  // Before the fix, activeTools being non-empty caused uniqueRequired to be filtered
-  // against the pi host tool list instead of MCP_WORKFLOW_TOOL_SURFACE, making every
-  // surface tool appear missing and aborting the guided flow with a false error.
+  // Guided flow starts the workflow MCP server as part of dispatch, so the
+  // parent session active-tool list is not authoritative for MCP tools.
   const discussMilestoneTools = [
     "gsd_summary_save",
     "gsd_requirement_save",
@@ -872,7 +871,6 @@ test("discuss-milestone guided flow does not abort when all required tools are o
       unitType: "discuss-milestone",
       authMode: "externalCli",
       baseUrl: "local://claude-code",
-      activeTools: ["ScheduleWakeup", "ToolSearch", "bash", "read", "write"],
     },
   );
 


jeremymcs:
## Code owner review — CI failure root cause analysis

**PR is merged**, but leaving this analysis for the record since the CI failure was non-trivial.

### Root Cause

The failing unit test was:
> `discuss-milestone guided flow does not abort when all required tools are on MCP surface (regression #469)`

This test was added by PR #471 (commit `14fa674d`) to document that `getWorkflowTransportSupportError` should not block guided flows when MCP surface tools are present. It expected `null` when `activeTools` only contained pi host tools and all required tools were in `MCP_WORKFLOW_TOOL_SURFACE`.

Then commit `c69bb283` (part of PR #478 merge) changed `getWorkflowTransportSupportError` to check ALL required tools against `activeTools` when `activeTools` is known — to fix the run-uat dispatch blocking for tool-starved sessions. This made the test expect `null` when the new code returned an error, failing CI.

### The Contradiction

Tests 798 and 854 used the same `activeTools` parameter but expected opposite results:
- Test 798 (run-uat): `gsd_summary_save` not in `activeTools` → expects ERROR ✓ intent: MCP not connected, block dispatch
- Test 854 (guided flow): same `activeTools` → expects NULL ✓ intent: guided flow starts MCP during launch, don't abort

### Fix Applied

The merged fix (`df6d78f4`) removed `activeTools` from test 854, aligning the test with actual production behavior: `guided-flow.ts` explicitly does NOT pass `activeTools` when calling this function (as documented in the code comment: "Guided flow starts the MCP workflow server as part of dispatch, so the parent session's activeTools doesn't include MCP tools yet").

A more defensive fix is on branch `codex/fix-pr473-ci-after-merge` as commit `9e6915f2` — it adds `requireMcpToolsConnected?: boolean` to `WorkflowCapabilityOptions` so the function can't be misused for guided flows even if `activeTools` is passed. This can be landed as a follow-up if desired.

PR #506, opened 2026-06-05
perf(pi-tui): eliminate render-loop Intl.Segmenter GC churn

### LINKED ISSUE #505: [Bug]: TUI render loop burns CPU via Intl.Segmenter GC churn on long sessions
### Bug type
Performance problem

### Summary
Long-lived TUI sessions slowly accumulate CPU because the spinner-driven render loop re-measures transcript text through `Intl.Segmenter`, allocating an ICU break iterator per call and producing steady V8 GC finalization churn.

### Steps to reproduce
1. Start a TUI session and let it build scrollback beyond ~512 distinct rendered lines (a long auto-mode run, a large streamed response, or scrolled history).
2. Leave it on the working-indicator spinner, or resize the window over a large transcript.
3. Sample the process CPU profile (e.g. `sample <pid> 5` on macOS, or any sampling profiler).
4. Observe the hot frames in the active (non-`kevent`) portion of the main thread.

### Expected behavior
Idle or lightly-active sessions should not accumulate meaningful CPU. Text width measurement should not dominate render cost or generate sustained GC pressure.

### Actual behavior
The active main-thread time is dominated by ICU break-iterator work from `Intl.Segmenter`, including the destructor as a hot leaf — the tell that a break iterator is allocated and freed per call:

```
icu_77::RuleBasedBreakIterator::~RuleBasedBreakIterator()
v8::internal::Builtin_SegmentIteratorPrototypeNext
v8::internal::JSSegmentIterator::Next
GlobalHandles::PostGarbageCollectionProcessing
  -> ManagedObjectFinalizerSecondPass -> ~RuleBasedBreakIterator
```

Under load (window resize over a large transcript), segmentation accounts for a large share of active main-thread time, and the spinner keeps that path hot continuously — so a long-running or idle-but-open session steadily accrues CPU behind the spinner.

Root cause: `visibleWidth()` in `packages/pi-tui/src/utils.ts` is called per changed line, per frame, by the differential renderer. When scrollback exceeds the 512-entry width cache, each frame re-segments uncached lines via `Intl.Segmenter`.

### GSD version
1.1.1

### Node.js version
v24.11.0

### Operating system
macOS 26.5 (arm64)

### Install method
npm global install

### Affected area
TUI / terminal UI

```
A CPU sample of an active session shows ~RuleBasedBreakIterator as a hot leaf,
with JSSegmentIterator::Next and Builtin_SegmentIteratorPrototypeNext in the
active main-thread subtree.
```

### Impact
Slow CPU drain (and battery/thermal cost) on any long-running or idle-but-open session; the worst case is unattended long auto-mode runs. The spinner animates throughout, so the cost is not visible to the user. Severity: moderate — affects every sufficiently long session; not a crash.

### Additional information
`Intl.Segmenter` is only needed for multi-codepoint grapheme clusters; the bulk of transcript content is one codepoint per grapheme, so the segmentation is avoidable with a codepoint fast-path in `visibleWidth`.

aidanthewiz:
## Linked issue

Closes #505

---

## TL;DR

**What:** Eliminate the TUI render-loop `Intl.Segmenter` GC churn that drains CPU on long-lived sessions, while keeping the spinner and all output visually identical.
**Why:** On long sessions the spinner-driven re-render re-segments every transcript line through `Intl.Segmenter`, allocating an ICU break iterator per call — steady V8 GC churn (`ICU ~RuleBasedBreakIterator` finalization) that burns CPU even while idle behind the spinner.
**How:** Add a codepoint fast-path to `visibleWidth` that skips `Intl.Segmenter` when a string has no multi-codepoint graphemes (reusing the exact width logic), and route `truncateToWidth` to the native text engine behind a capability probe.

## What

Affects **`pi-tui`** (the hot path) and **`native`** (supporting fixes):

- `packages/pi-tui/src/utils.ts`
  - `visibleWidth`: codepoint fast-path. When the (ANSI/tab-stripped) string contains no codepoint that can form or modify a multi-codepoint grapheme, sum `graphemeWidth()` per codepoint instead of walking `Intl.Segmenter`. Falls back to the segmenter when a combining-capable codepoint is present. Width logic is reused verbatim, so output is unchanged.
  - `truncateToWidth`: routes to the native engine for the three known ellipsis kinds (`…`, `...`, `""`) behind a startup capability probe (so an older engine binary transparently falls back to JS), with a guard that falls back for lone/malformed-ESC input.
  - `finalizeTruncatedResult`: always reset-brackets the ellipsis (`prefix + \x1b[0m + ellipsis + \x1b[0m`) so truncated output is safe to composite into a styled stream.
- `native/crates/engine/src/text.rs` — align `truncate_to_width` with the JS reset-bracketing and fix two pre-existing native bugs found during verification: a NUL terminator leaking into padded output, and a missing/duplicated reset on the ellipsis-wider-than-width and dangling-SGR boundaries.
- `packages/native/src/native.ts` — add a `GSD_NATIVE_PREFER_LOCAL=1` loader override so a locally built engine can be loaded ahead of the pinned npm binary when iterating on `native/crates`.
- Tests: `packages/pi-tui/test/visible-width-fastpath.test.ts` (new) and 4 new native truncate-bracketing tests in `packages/native/src/__tests__/text.test.mjs`.

## Why

`visibleWidth()` is called per changed line, per frame, by the TUI differential renderer; the working-indicator spinner keeps that path hot. When transcript scrollback exceeds the 512-entry width cache, every frame re-measures uncached lines, and the `Intl.Segmenter` walk allocates and frees an ICU break iterator per call. Over a long session this surfaces as steady GC churn — `GlobalHandles::PostGarbageCollectionProcessing → ManagedObjectFinalizerSecondPass → ~RuleBasedBreakIterator` — that accumulates CPU even while the session sits idle behind the spinner.

## How

`Intl.Segmenter` is only required to locate *multi-codepoint* grapheme clusters. The overwhelming majority of transcript content (code, paths, ANSI-styled ASCII, plain CJK) is one codepoint per grapheme, so segmentation there is pure overhead. The fast-path pre-checks with a property-based regex (`\p{Mark}` + ZWJ/ZWNJ, Thai/Lao AM, halfwidth voiced marks, variation selectors, regional indicators, skin-tone modifiers, emoji tags) and, when none are present, sums the existing per-grapheme `graphemeWidth()` over codepoints. The width logic is identical in both branches, so output is byte-for-byte unchanged — including the deliberate regional-indicator = width-2 streaming-drift mitigation.

Key decisions / alternatives considered:
- **Pure-JS fast-path for `visibleWidth`** rather than routing it to the native engine. Native (`unicode-width`) and the JS width logic diverge on a set of codepoints (regional indicators, Hangul jamo, Indic/Arabic marks), and the regional-indicator difference has a dedicated regression test. Routing `visibleWidth` to native would silently revert that, so the churn is removed in JS with zero behavior change instead.
- **`truncateToWidth` does** route to native, but only behind a startup capability probe so output never diverges from the JS path regardless of which engine binary is loaded; an older engine simply falls back to JS.
- The fast-path trigger set is property-based (future-proof across Unicode versions) and verified complete against `Intl.Segmenter` across every codepoint U+0020–U+10FFFF (zero missed combining codepoints).

### Measured impact

A/B between this branch and the prior build, two sessions given an identical long markdown-generation prompt and sampled simultaneously with `sample(1)` on macOS:

- **Resize stress (full-transcript re-measure):** prior build ~495–606 ICU/Segmenter frames per 3s sample vs this branch ~0–9; at peak (both main threads ~100% active) segmentation was ~14% of active main-thread time on the prior build and 0% here.
- **Streaming the same prompt:** prior build used ~1.6× the cumulative CPU; at one equal-CPU instant the prior build showed 221 ICU frames vs 0 here.
- **Isolated microbenchmark (large uncached transcript):** ~18× faster width measurement, output byte-identical.
- **Signature:** the prior build shows the `~RuleBasedBreakIterator` destructor as a hot leaf under load; this branch shows zero ICU/Segmenter frames in every sample (text measured via the native engine + JS fast-path).

## Change type

- [x] `perf` — eliminates render-loop CPU churn (no functional/output change)
- [x] `test` — adds equivalence + native truncate tests

## Scope

- [x] `pi-tui` — Terminal UI
- [x] `native` — Native bindings

## Breaking changes

- [x] No breaking changes

Output is byte-identical: `visibleWidth` and `truncateToWidth` results were verified against the prior implementation across the full codepoint range, multi-codepoint clusters (combining marks, ZWJ, regional indicators, variation selectors, skin tones), and ANSI/OSC/APC content.

## Test plan

- [x] CI passes
  - `visible-width-fastpath.test.ts`: fast-path is byte-identical to an `Intl.Segmenter` reference over every codepoint U+0020–U+1FAFF, multi-codepoint clusters, and ANSI/OSC/APC content; isolated regional indicators stay width 2.
  - 4 new native truncate tests: reset-bracketing (plain + wide-ellipsis), dangling-SGR trim, no-NUL-in-padding.
  - `format-utils.test.ts` truncation assertions corrected to check `visibleWidth(result) <= width` instead of raw string length, the correct invariant for ANSI-aware output.
  - pi-tui and native suites pass on both the native and JS-fallback paths.
- [x] Manual testing — A/B CPU sampling under streaming, window resize, and concurrent stream+resize (see Measured impact).

## AI disclosure

- [x] This PR includes AI-assisted code. All changes were verified by the author: full test suites on both the native and JS-fallback paths, a byte-identical equivalence corpus, and A/B CPU sampling against the prior build.

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-pi/pr/506"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783262875&installation_id=134682257&pr_number=506&repository=open-gsd%2Fgsd-pi&return_to=https%3A%2F%2Fgithub.com%2Fopen-gsd%2Fgsd-pi%2Fpull%2F506&signature=5ee3401fdd2ceb8151747646d9992e7176f2c7e7026a331b5a7d4aefc97d86d5"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


jeremymcs:
@aidanthewiz please fix CI and we'll get this merged

aidanthewiz:
@jeremymcs the CI failure appears to already be on main from #502, not introduced by this PR. Here is a failing workflow on main: https://github.com/open-gsd/gsd-pi/actions/runs/27018535246/job/79739964097

Would you like me to investigate?

Issue #373, opened 2026-06-01
Irreversible next-step items self-execute instead of routing through ask_user_questions (confirm-first rule uncoupled from next-steps convention; destructive-guard hook is auto-only & warn-only)


splichy:
## Problem

In an **interactive** session (auto-mode off, no active milestone/unit), the agent printed a numbered "Next steps: 1. Apply the saved plan…" menu and then, **in the same turn**, executed the irreversible Bash command `terragrunt apply tfplan` (22 real AWS resources created) without stopping for a fresh user confirmation. The user had been replying "1." to advance earlier next-step menus, so the agent treated its own prose menu as a standing queue and self-executed item 1.

The user's framing pinpoints the design gap: GSD's only primitive that *actually blocks a turn and waits* is the `ask_user_questions` tool. The irreversible action was never routed through it — it leaned on the prose "Next steps" convention, which carries zero enforcement. GSD does **not** answer its own questions (no self-response feature exists); the bug is that the agent over-continued off a decorative prose list instead of ending the turn.

## Root Cause

Three independent gaps, none of which would have stopped this:

**1. The "Next steps" prose convention is uncoupled from the confirm-first rule.**
- `prompts/system.md:161` — *"After completing a task, give a brief summary and 2-4 numbered next-step options; last option is always 'Other'."* This is plain prose, not a tool call. The harness does not block on it.
- `prompts/system.md:35` — the confirm-first rule enumerates *"GitHub or external services… any state change outside local filesystem,"* but every example is GitHub/PR/publish-centric. No mention of cloud-infra mutation, so `terragrunt apply` reads as a benign local command.
- The two lines never reference each other. Nothing states that a next-step item which is irreversible/outward-facing must be confirmed via the blocking `ask_user_questions` primitive rather than executed off a prior numbered reply.

**2. The destructive-command classifier omits IaC/cloud-mutation verbs** — `safety/destructive-guard.js:8-19`:
```js
const DESTRUCTIVE_PATTERNS = [
  { pattern: /\brm\s+(-[^\s]*[rfRF]...)/, label: "recursive delete" },
  { pattern: /\bgit\s+push\s+.*--force/, label: "force push" },
  { pattern: /\bdrop\s+(database|table|index)\b/i, label: "SQL drop" },
  // ... no terraform/terragrunt apply|destroy, no `aws ... delete/put/create/terminate`, no kubectl
];
```
`terragrunt apply` matches nothing → never even classified as destructive.

[...]

PR #679, opened 2026-06-11
fix(gsd): make destructive-command HARD BLOCK escapable via confirmation

### LINKED ISSUE #678: [Bug]: Destructive-command HARD BLOCK is unwinnable — confirming via ask_user_questions never clears the gate
# [Bug]: Destructive-command HARD BLOCK is unwinnable — confirming via ask_user_questions never clears the gate

**Bug type:** Regression (worked before, now fails)

## Summary

Any bash command classified as destructive (force push, `rm -rf`, SQL drop, etc.) is HARD-blocked with instructions to confirm via `ask_user_questions` and re-issue it — but nothing ever detects the confirmation, so re-issuing hits the identical block every time. The result is an unwinnable loop.

## Steps to reproduce

1. In any mode (interactive or auto), have the agent attempt a destructive bash command, e.g. `git push --force-with-lease origin <branch>`.
2. The guard returns `HARD BLOCK: destructive Bash command requires explicit human confirmation … then issue the command only when confirmed in the current turn.`
3. Confirm the action via `ask_user_questions`.
4. Re-issue the exact command in the same turn.
5. Observe: the identical HARD BLOCK fires again. No sequence of confirmations clears it.

## Expected behavior

After the user explicitly confirms the destructive command, re-issuing that exact command in the same turn should run it once. The block should be *escapable* via confirmation, not unwinnable.

## Actual behavior

The `tool_call` guard in `src/resources/extensions/gsd/bootstrap/register-hooks.ts` returns `{ block: true }` unconditionally for any classified-destructive command. There is no state tracking that the user confirmed, and nothing downstream can release the block (the hook contract: returning `{ block: true }` prevents the tool from executing). The "confirm in the current turn" escape hatch promised by the error message was never implemented. A repo-wide search for confirmation-tracking state returns zero hits.

## Root cause / regression origin

Commit `06212d00` (closing #373), titled *"…destructive-guard hook is auto-only & warn-only"*, did the opposite of its own description:

- Removed `if (!isAutoActive()) return;` — the guard now fires in **all modes** (including interactive, which is how this was hit).
- Replaced `ctx.ui.notify(…warning…)` with `return { block: true }` — turning a non-blocking warning into a **hard, inescapable block**.

## GSD version

1.2.0

## Node.js version

v24.11.0

## Operating system

macOS 27.0

## Install method

npm link / local checkout

## Affected area

MCP / tools / extensions (GSD safety harness — destructive-command guard)

## Impact

- Severity: blocks any legitimate destructive operation behind an unwinnable loop; wastes turns; can strand auto-mode.
- Frequency: 100% repro for any classified-destructive command.
- Consequence: the user cannot complete a confirmed force push / cleanup even after explicitly approving it.

## Additional information

Last known good: before `06212d00` the guard was warn-only and auto-only, so interactive destructive commands were never blocked. Fix branch `fix/destructive-guard-loop` adds a one-shot, command-bound, in-memory confirmation token that makes the block escapable while keeping it in all modes.

aidanthewiz:
## Linked issue

Closes #678

---

## TL;DR

**What:** Make the destructive-command HARD BLOCK escapable via an explicit one-shot confirmation token instead of an unwinnable loop.
**Why:** Confirming a destructive command via `ask_user_questions` never cleared the gate, so re-issuing it hit the identical block forever (regression from `06212d00`).
**How:** A new in-memory, command-bound, per-basePath confirmation token: the guard records the blocked command and blocks; an affirmative `destructive_confirm` gate answer promotes it; the next matching bash call consumes it and runs once.

## What

Adds the missing confirmation path for the destructive-command safety guard.

- **`src/resources/extensions/gsd/safety/destructive-confirmation.ts`** (new) — one-shot confirmation token module: `requestDestructiveConfirmation` (record pending at block time), `confirmDestructiveCommand` (promote pending → confirmed on affirmative answer), `consumeDestructiveConfirmation` (allow the exact command through once), plus `normalizeDestructiveCommand`, `isDestructiveConfirmGateId`, and test/diagnostic helpers.
- **`src/resources/extensions/gsd/bootstrap/register-hooks.ts`** — wires the token in:
  1. bash `tool_call` guard: consume a matching token (allow once) before blocking; otherwise record the command as pending and block with corrected instructions (use a question id containing `destructive_confirm` whose first option affirms the action).
  2. `ask_user_questions` `tool_result` handler: an affirmative answer to a `destructive_confirm` gate promotes the pending command to a confirmed token. Reuses the existing fail-closed `isDepthConfirmationAnswer` predicate.
- **`src/resources/extensions/gsd/tests/destructive-confirmation.test.ts`** (new) — 13 tests across two layers: **10 unit** (loop repro, block→confirm→allow-once, one-shot re-block, command-binding, whitespace-normalization, stale-token invalidation, per-basePath isolation, gate-id detection, non-destructive passthrough) and **3 integration** that drive the real `registerHooks()` bash guard + `ask_user_questions` handler end-to-end (block→confirm→allow-once, decline-stays-blocked, and stale-token invalidation on an intervening destructive command).

The hard block remains in all modes; it is now *escapable* rather than *unwinnable*.

## Why

The bash `tool_call` guard returned `{ block: true }` unconditionally for any classified-destructive command, with no state tracking user confirmation. The error told the model to confirm via `ask_user_questions` and re-issue "in the current turn", but nothing ever detected that confirmation — an unwinnable loop (observed on a `git push --force-with-lease`).

Regression from `06212d00` (closing #373), whose message claimed "auto-only & warn-only" but whose diff removed the `isAutoActive()` guard and replaced `ctx.ui.notify` with a hard block in all modes.

## How

A one-shot, command-bound confirmation token with these safety invariants (each pinned by a test):

- **In-memory only** — never persisted. A token on disk could silently auto-approve a destructive command in a later session; confirmation must be re-obtained every process lifetime.
- **One-shot** — consuming a token clears it, so a second destructive command (even identical) re-blocks and re-prompts.
- **Command-bound** — only the exact whitespace-normalized command the user confirmed matches; a reworded command re-blocks.
- **Per basePath** — concurrent workspaces in one process never share tokens.

The `destructive_confirm` gate id is intentionally distinct from the `depth_verification` gate patterns (the only entries in `GATE_QUESTION_PATTERNS`), so the two gate systems do not interfere — a `destructive_confirm` answer never trips the depth-gate's early "waiting" return, and the confirm handler is always reached. The guard and the `tool_result` handler both derive basePath from `contextBasePath(ctx)`, so token write/lookup keys match.

Alternatives considered: (a) reverting `06212d00` to warn-only — rejected, it would re-open #373's self-executing-destructive-next-step problem; (b) persisting the token — rejected for the silent-auto-approve-in-later-session risk above.

## Change type

- [x] `fix` — Bug fix

## Scope

## Breaking changes

- [x] No breaking changes

(The destructive guard's user-visible block message changed wording to instruct the new `destructive_confirm` gate flow; behavior is strictly more permissive — previously-blocked confirmed commands now run once.)

## Test plan

- [x] New/updated tests included — `destructive-confirmation.test.ts` (13 tests). The `repro:` unit test and the `integration: real hooks block a force push, then allow it once after confirmation` test both fail without the escape hatch and pass with it.
- [x] `destructive-confirmation.test.ts` → **13/13 pass** (10 unit + 3 integration) via the TS-strip resolver.
- [x] Adjacent write-gate / depth-verification suites → **159/159 pass** (no regression from the rebase).
- [x] `typecheck:extensions` → **0 errors** (run after `build:core`; a stale `@opengsd/contracts` dist otherwise reports a pre-existing error in the untouched `unit-registry.ts`).
- [x] Full `test:unit` → 10510 pass / 15 fail. **All 15 failures reproduce verbatim on a clean `origin/main` checkout** (onboarding-state, cost-spike telemetry, `/gsd discuss` routing, `shared/mod.ts` import, dev-CLI `GSD_BIN_PATH` — one even references a sibling worktree path). They are pre-existing / environmental and untouched by this 3-file change.
- [x] Manual: the live guard in this very session still emitted the **old** unfixed block message on a `rm` symlink cleanup, confirming the bug is active on the base and this PR is the fix.

## AI disclosure

- [x] This PR includes AI-assisted code. Investigation, implementation, and tests were AI-assisted and verified locally (full suite run, root-cause traced to `06212d00`, regression test confirmed to fail pre-fix / pass post-fix).

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-pi/pr/679"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a>

### DIFF
--- a/src/resources/extensions/gsd/bootstrap/register-hooks.ts
+++ b/src/resources/extensions/gsd/bootstrap/register-hooks.ts
@@ -13,7 +13,7 @@ import type { GSDEcosystemBeforeAgentStartHandler } from "../ecosystem/gsd-exten
 import { updateSnapshot } from "../ecosystem/gsd-extension-api.js";
 
 import { buildMilestoneFileName, clearPathCache, milestonesDir, resolveMilestonePath, resolveSliceFile, resolveSlicePath } from "../paths.js";
-import { applyAskUserQuestionsGateResult, canonicalToolName, clearDiscussionFlowState, formatPendingAskUserQuestionsGateMessage, isApprovalGateVerifiedInSnapshot, isMilestoneDepthVerified, isMilestoneDepthVerifiedInSnapshot, isQueuePhaseActive, loadWriteGateSnapshot, markApprovalGateVerified, markDepthVerified, refreshWriteGateStateFromDisk, resetWriteGateState, shouldBlockContextWrite, shouldBlockPlanningUnit, shouldBlockQueueExecution, shouldBlockWorktreeWrite, isGateQuestionId, setPendingGate, clearPendingGate, getPendingGate, shouldBlockPendingGate, shouldBlockPendingGateBash, extractDepthVerificationMilestoneId } from "./write-gate.js";
+import { applyAskUserQuestionsGateResult, canonicalToolName, clearDiscussionFlowState, formatPendingAskUserQuestionsGateMessage, isApprovalGateVerifiedInSnapshot, isDepthConfirmationAnswer, isMilestoneDepthVerified, isMilestoneDepthVerifiedInSnapshot, isQueuePhaseActive, loadWriteGateSnapshot, markApprovalGateVerified, markDepthVerified, refreshWriteGateStateFromDisk, resetWriteGateState, shouldBlockContextWrite, shouldBlockPlanningUnit, shouldBlockQueueExecution, shouldBlockWorktreeWrite, isGateQuestionId, setPendingGate, clearPendingGate, getPendingGate, shouldBlockPendingGate, shouldBlockPendingGateBash, extractDepthVerificationMilestoneId } from "./write-gate.js";
 import { resolveManifest } from "../unit-context-manifest.js";
 import { isBlockedStateFile, isBashWriteToStateFile, BLOCKED_WRITE_ERROR } from "../write-intercept.js";
 import { loadFile, saveFile, formatContinue } from "../files.js";
@@ -39,6 +39,12 @@ import { saveActivityLog } from "../activity-log.js";
 import { recordToolCall as safetyRecordToolCall, recordToolResult as safetyRecordToolResult, saveEvidenceToDisk } from "../safety/evidence-collector.js";
 import { parseUnitId } from "../unit-id.js";
 import { classifyCommand } from "../safety/destructive-guard.js";
+import {
+  confirmDestructiveCommand,
+  consumeDestructiveConfirmation,
+  isDestructiveConfirmGateId,
+  requestDestructiveConfirmation,
+} from "../safety/destructive-confirmation.js";
 import { logWarning as safetyLogWarning } from "../workflow-logger.js";
 import { isUnitCloseoutTool, runInteractiveUnitCloseout } from "../unit-closeout.js";
 import { installNotifyInterceptor } from "./notify-interceptor.js";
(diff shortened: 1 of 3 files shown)


aidanthewiz:
## ✅ Risk Verification — PASS

**Commit verified:** `a6b334db4d45f667a599324801721db5d6609902` · **Date:** 2026-06-12 · **Verified by:** GSD coding agent

| GSD Workflow | verify GSD workflow state transitions end-to-end | ✅ |

### Evidence
| Check | Command | Result |
|-------|---------|--------|
| destructive-confirmation (unit + integration) | `node --import ./src/resources/extensions/gsd/tests/resolve-ts.mjs --experimental-strip-types --test .../destructive-confirmation.test.ts` | 13/13 pass |
| workflow gates (depth-verification + write-gate) | `node ... --test .../register-hooks-depth-verification.test.ts .../write-gate.test.ts` | 54/54 pass |
| typecheck | `pnpm run typecheck:extensions` | 0 errors |

### Root cause & scope
- **Root cause:** The destructive-command HARD BLOCK in `register-hooks.ts` had no confirmation path — re-issuing a classified command (force-push, rm -rf, SQL drop, etc.) after confirming via `ask_user_questions` hit the identical block every turn, an unwinnable loop. Fix adds a one-shot, in-memory, command-bound confirmation token (`safety/destructive-confirmation.ts`) wired into the bash `tool_call` guard (consume/request) and the `ask_user_questions` `tool_result` handler (confirm).
- **Scope:** Single enforcement site — `classifyCommand` + "HARD BLOCK: destructive" appear exactly once in non-test source (`register-hooks.ts:1400`/`:1417`). `safety-harness.ts` only re-exports `classifyCommand`; no second guard loops. No duplicated logic.
- **Downstream/stale deps:** None. New module is consumed only by `register-hooks.ts`. Reused exports (`isDepthConfirmationAnswer`, `classifyCommand`, `contextBasePath`) all pre-exist and are unchanged.

### State transitions confirmed
`block → request-confirmation (pending) → affirmative gate answer → confirm (one-shot token) → allow-once → re-block`. Safety invariants verified: one-shot consumption, command-bound matching, whitespace-normalized match, stale-token invalidation on a new request, per-basePath isolation, decline keeps command blocked, no leak to a different destructive command.

### Notes / caveats
- Token is intentionally in-memory only (never persisted) — confirmation must be re-obtained every process lifetime; this is a deliberate safety property, not a gap.
- **Meta note:** posting this very comment initially tripped the destructive guard because the comment body contained a force-push command string — a reminder that `classifyCommand` matches on raw command text, so embedding example destructive commands in tool arguments can cause benign false positives.

PR #681, opened 2026-06-11
feat(gsd): fix shell-tool selection — uncapped sync bash, interruptible await, graceful kills

### LINKED ISSUE #680: [Feature]: Fix shell-tool selection — agent reaches for async_bash + immediate await instead of synchronous bash
## Summary

The agent habitually reaches for `async_bash` and then immediately calls `await_job` on it — a non-blocking tool used in a blocking way — for ordinary commands that should be synchronous `bash`. It also uses `bg_shell` + `wait_for_ready` for run-to-completion commands. Fix the tool behavior and guidance so the agent picks the right tool the first time.

## What a fix should do (at a glance)

- Stop the agent reaching for `async_bash` + an immediate `await_job` for ordinary commands
- Remove the hidden 120s cap on synchronous `bash` (the thing that made that habit "rational")
- Rewrite tool guidance into one rule: Quick → `bash`, background → `async_bash`, long-lived → `bg_shell`
- Make `await_job` interruptible — ESC ends the wait without killing the job
- Stop `bg_shell` / `wait_for_ready` from reporting a clean exit-0 as a failure
- Add a graceful kill ladder so the now-uncapped `bash` can't corrupt `terraform apply` / DB migrations on timeout/ESC
- Fix the leak it surfaced: SIGTERM-immune timed-out `async_bash` jobs left running in the background

## Problem to solve

Two bad habits, each with a concrete cause:

**1. `async_bash` + immediate `await_job` instead of sync `bash`.**
- The guidance said *"use async_bash for commands that take more than a few seconds"*, framing async_bash as the safe choice for anything non-trivial.
- A **silent 120s cap on synchronous `bash`** (injected by the GSD wrapper) made that instinct rational — sync `bash` really was secretly time-limited. So the agent kicked off `async_bash`, then immediately blocked on `await_job`: a pointless round-trip that adds latency and transcript clutter for zero benefit.
- Worse, `await_job` could not be interrupted. Once the agent blocked on it, ESC didn't free it — it was stuck until the job finished or the full timeout elapsed. And `await_job` suppressed follow-ups for *all* watched jobs upfront, so an aborted job's result could be lost.

**2. `bg_shell` + `wait_for_ready` for run-to-completion commands.**
- `wait_for_ready` is for long-lived processes that open a port or print a ready pattern. Run-to-completion commands (`terraform apply`, migrations, builds, tests, installs) exit instead of becoming "ready", so a clean exit-0 was reported as a readiness failure — making the agent think the command failed.

Underlying both: the tools don't do what their descriptions imply, and a hidden cap silently changes synchronous semantics.

## Proposed solution

Grouped by area, with the reason for each:

**A. Make the behavior honest, then describe it accurately.**
- **Remove the silent 120s sync-`bash` cap.** Sync `bash` becomes genuinely uncapped, so the agent no longer needs `async_bash` to escape a hidden timeout. Genuine hangs are owned by the auto-mode stalled-tool watchdog (5m/10m/20m/30m) in auto-mode and by human ESC interactively.
- **Rewrite the guidance** for `bash`, `async_bash`, `await_job`, `bg_shell`, and `system.md` around one decision rule — *Quick → sync `bash`; one-shot in the background so you can keep working → `async_bash` + later `await_job`; long-lived process you interact with over time → `bg_shell`* — and frame async_bash as "for non-blocking behavior, not because of a timeout concern."
- **`wait_for_ready` rejects run-to-completion use** and reports a clean exit-0 as completion (pointing at `async_bash`) instead of a crash.

**B. Make `await_job` interruptible**, so blocking on a job is no longer a trap: ESC ends the wait without killing the job; the job keeps running and resurfaces when complete; suppression applies only to jobs that actually completed.

**C. Make "uncapped sync `bash`" safe** (the precondition for removing the cap):
- Route every kill path (sync bash, async_bash, bg_shell, GSD exec-sandbox) through one graceful SIGTERM → grace → SIGKILL ladder, with a caller-side hard deadline so a non-closing (D-state) child can't hang `bash` forever. This protects stateful commands (`terraform apply`, DB migrations) from a bare SIGKILL on timeout/ESC.
- This also fixes a real leak found along the way: the old `async_bash` kill helper only ever sent SIGTERM, so a SIGTERM-immune timed-out job was left running in the background while its promise force-resolved.

**D. Fix a pre-existing TUI bug on the same ESC/await surface:** pressing ESC during a later await flipped already-completed tool cards to red. Only genuinely-incomplete tools should be marked interrupted.

## Alternatives considered

- **Just rewrite the guidance, keep the 120s cap.** Rejected — the guidance would still be dishonest (sync bash *is* capped), so the agent's avoidance instinct would remain rational.
- **Raise the cap instead of removing it.** Rejected — still arbitrary, still silently truncates long commands, still overlaps the watchdog.
- **Keep the bare SIGKILL.** Rejected — removing the cap means sync `bash` now runs arbitrarily long stateful commands, which must not be hard-killed mid-write on timeout/ESC.
- **A shared constants module for the kill timing.** `pi-agent-core` is the lowest layer and must not depend on `pi-coding-agent`, so a shared import is impossible there. Chosen compromise: one canonical definition in `shell.ts` imported by the higher layers, plus one deliberate mirror in `pi-agent-core` locked by a parity test.

## Use cases

- A 30s build or a 3-minute `terraform apply` run as plain synchronous `bash` — read the result inline, no async round-trip.
- `async_bash` used for what it's actually for: start a long job, keep working, `await_job` later.
- Cancelling/ESC-ing a long `await_job` without orphaning the background job.
- Run-to-completion commands no longer mis-reported as "not ready" by `bg_shell`.

## Impact

- **Affected:** every agent run that shells out — auto-mode and interactive.
- **Severity:** medium-high. Not a crash, but a pervasive workflow-quality problem: wasted round-trips, confusing "not ready" results, lost aborted-job results, and (on the safety side) a real risk of corrupting stateful commands and leaking background processes once the cap is removed.
- **Frequency:** constant — the async_bash-then-immediate-await reflex shows up in routine command execution.
- **Consequence:** slower, noisier agent transcripts; misread failures; unsafe kills of long-running stateful commands.

## Evidence or prior art

Old guidance string: "Use async_bash for commands that take more than a few seconds (builds, tests, installs, large git operations)." Implementation references bug IDs #2186 (D-state hang force-resolve), #2248 (follow-up delivery), #3787 (cross-turn suppression), #1209 (session process cleanup). Covered by behavior-executing tests (no source-grep), including two regression tests proven to fail without their fix (cancelled-status guard; constant-parity drift).

## Additional information

- Environment: GSD 1.2.0, Node v24.11.0, macOS 27.0.
- Breaking change: removal of the silent 120s sync-bash cap (uncapped now; watchdog/ESC own hangs).
- The `pi-agent-core` kill-helper change is behavior-preserving (bare SIGKILL → graceful ladder) and touches only the local helper, not orchestration — flagged given that package's RFC-sensitivity.

aidanthewiz:
## Linked issue

Closes #680

---

## TL;DR

**What:** Fix how the agent chooses and uses shell tools — stop the reflex of reaching for `async_bash` + an immediate `await_job` (and `bg_shell` for run-to-completion commands) — by making synchronous `bash` genuinely uncapped and the tool verbiage honest about it.
**Why:** The agent defaulted to `async_bash` then immediately blocked on `await_job` for ordinary commands, because the guidance implied sync `bash` was time-limited/risky and a silent 120s cap actually made it so. That's a pointless round-trip, and `await_job` couldn't even be interrupted.
**How:** Remove the silent 120s sync-bash cap, rewrite the bash/async_bash/await_job/bg_shell/system.md guidance around a clear Quick(sync) / One-shot(async_bash) / Background(bg_shell) decision rule, make `await_job` interruptible, and back it with a graceful kill ladder so "sync bash is safe to run uncapped" is actually true.

This is primarily a **behavioral / agent-guidance** change. The kill-safety work exists to make the behavioral change honest and safe — it is the supporting layer, not the headline.

- Stops the agent reaching for `async_bash` + an immediate `await_job` for ordinary commands
- Removes the hidden 120s cap on synchronous `bash` (the thing that made that habit "rational")
- Rewrites tool guidance into one rule: Quick → `bash`, background → `async_bash`, long-lived → `bg_shell`
- Makes `await_job` interruptible — ESC ends the wait without killing the job
- Stops `bg_shell` / `wait_for_ready` from reporting a clean exit-0 as a failure
- Adds a graceful SIGTERM → grace → SIGKILL kill ladder so the now-uncapped `bash` can't corrupt `terraform apply` / DB migrations on timeout/ESC
- Fixes a real leak: SIGTERM-immune timed-out `async_bash` jobs were left running in the background
- Fixes a TUI bug: pressing ESC during a later await flipped already-finished tool cards red

---

## Why (the root problem)

Two bad agent habits, both caused by tool descriptions that didn't match reality plus a hidden cap:

1. **`async_bash` + immediate `await_job` instead of synchronous `bash`.** The old guidance said *"use async_bash for commands that take more than a few seconds"*, framing async_bash as the safe choice for anything non-trivial. A **silent 120s cap on synchronous `bash`** (injected by the GSD wrapper) made that instinct rational — sync `bash` really was secretly time-limited. So the agent kicked off `async_bash` and then immediately blocked on `await_job`: a non-blocking tool used in a blocking way, adding a pointless round-trip. And `await_job` couldn't be interrupted, so once it blocked it was stuck.

2. **`bg_shell` + `wait_for_ready` for run-to-completion commands.** `wait_for_ready` is for long-lived processes that open a port / print a ready pattern. Run-to-completion commands (`terraform apply`, migrations, builds) exit instead of becoming "ready", so a clean exit-0 was reported as a readiness failure.

The fix: make the tools behave honestly (uncapped, interruptible, graceful) and describe them accurately, so the agent picks the right tool the first time. Removing the cap is only safe if a timeout/ESC can no longer corrupt a long stateful command — hence the graceful kill ladder.

---

### A. Behavioral fix: tool guidance + uncapped sync bash

| File | What changed | Why |
| --- | --- | --- |
| `gsd/bootstrap/dynamic-tools.ts` | Stopped injecting a default 120s timeout into the GSD `bash` tool. Appends the auto-mode watchdog detail to the bash tool's description and `timeout` param. | Removing the silent cap is the core fix — sync `bash` is now genuinely uncapped, so the agent no longer needs `async_bash` to escape a hidden timeout. The watchdog detail is injected here (GSD-scoped) because only GSD/auto-mode has the watchdog; non-GSD embeddings don't. |
| `gsd/constants.ts` | Deleted `DEFAULT_BASH_TIMEOUT_SECS = 120`. | Dead once the cap injection is gone. |
| `gsd/prompts/system.md` | Replaced the old "Background processes / One-shot commands" prose with a single decision rule: Quick → sync `bash`; one-shot non-blocking → `async_bash` + later `await_job`; long-lived → `bg_shell`. States sync bash is uncapped (hangs owned by the auto-mode watchdog / human ESC). | The agent was choosing tools by *how long* a command takes; the new rule teaches it to choose by *how the command runs*. |
| `pi-coding-agent/src/core/tools/bash.ts` | Tool description now says "No default timeout — pass an explicit timeout or rely on human cancellation." | Makes the core tool's self-description honest for all embeddings. |
| `async-jobs/async-bash-tool.ts` | Rewrote `promptGuidelines`: async_bash is "for non-blocking behavior, not because of a timeout concern." | Directly kills the "async_bash because sync is capped" reasoning. |
| `bg-shell/bg-shell-tool.ts` | Rewrote `promptGuidelines`: use `bg_shell` only for processes that STAY ALIVE; never use `wait_for_ready` on a run-to-completion command. | Stops the second bad habit (bg_shell for batch commands). |
| `bg-shell/readiness-detector.ts` | When a watched process exits 0 before signaling readiness, return a "ran to completion — use async_bash" message instead of a crash/error. | A clean exit was being reported as a readiness failure; now it's reported as success pointing at the right tool. |

### B. Interruptible `await_job` (so blocking on a job is no longer a trap)

| File | What changed | Why |
| --- | --- | --- |
| `async-jobs/await-tool.ts` | ESC/AbortSignal now ends the wait promptly. Moved follow-up suppression from upfront to *after* the race, so only completed jobs are suppressed; an aborted-but-still-running job stays unsuppressed and resurfaces later. Tears down the race timer + abort listener on settle. Renders already-`delivered` jobs as a one-line acknowledgement instead of reprinting. | Previously `await_job` couldn't be interrupted and suppressed follow-ups for *all* watched jobs upfront — so an aborted job's result was lost. Now ESC frees the agent without killing the job. |
| `async-jobs/job-manager.ts` | Added a `delivered` flag set when the follow-up actually fires; **added a `cancelled`-status guard to the `.then` branch** (mirrors the existing `.catch` guard). | `delivered` stops a later-turn `await_job` from reprinting output already in context. The `.then` guard fixes a real bug: async_bash *resolves* (not rejects) even when aborted, so a user-cancelled job's status was clobbered back to `completed`. |
| `async-jobs/index.ts` | `/jobs` is now an interactive pick → confirm → cancel kill-picker (was select-immediately-cancels). Labels drop the baked-in elapsed time (which `ui.select` renders once and freezes); accurate times go in the post-picker summary. Headless falls back to text. | Prevents destroying a job on a single keypress and avoids a misleading frozen "(24s)". |

### C. Graceful kill ladder (the safety net that makes "uncapped" honest)

| File | What changed | Why |
| --- | --- | --- |
| `pi-coding-agent/src/utils/shell.ts` | `killProcessTree` escalates SIGTERM → 5s grace → SIGKILL (was a bare SIGKILL), every timer `.unref()`'d. Windows force-kills the tree via `taskkill /F /T`. Exports `SIGKILL_GRACE_MS` / `HARD_DEADLINE_MS` as the single source of truth. | A bare SIGKILL on timeout/ESC can corrupt a stateful command (`terraform apply`, DB migration). With the cap removed, sync bash runs arbitrarily long, so a clean-shutdown window matters. |
| `pi-coding-agent/src/core/tools/bash.ts` | Added a caller-side hard deadline: after a kill is initiated, race the real `close` against a deadline so a non-closing (D-state) child force-resolves with a distinct `force-killed` marker instead of hanging. | `killProcessTree` can signal but can't force-resolve an awaiting caller; a D-state child never emits `close`, so without this the tool hangs forever even after SIGKILL. |
| `pi-coding-agent/src/index.ts` | Re-export `killProcessTree`, `SIGKILL_GRACE_MS`, `HARD_DEADLINE_MS`. | So the async_bash tool and exec-sandbox use the one canonical kill path + constants instead of redefining them. |
| `async-jobs/async-bash-tool.ts` | Deleted the local `killTree` helper; both kill paths (timeout and onAbort/ESC) now call the shared `killProcessTree`. onAbort arms the same hard deadline. Imports the shared constants. | **Real bug fix:** the old `killTree` only ever sent SIGTERM (even its "escalate" branch re-sent SIGTERM), so a SIGTERM-immune timed-out job was left running in the background while the promise force-resolved. Now it actually escalates to SIGKILL. |
| `gsd/exec-sandbox.ts` | Timeout path uses `killProcessTree` (was `child.kill("SIGTERM")`, which on Windows orphaned grandchildren) + the hard-deadline force-resolve. Added a `force_resolved` flag persisted to the `.meta.json` sidecar. Imports the shared constants. | Closes the identical bare-kill / orphaned-tree hazard on `gsd_exec` / `gsd_uat_exec`, and lets callers tell a hard-deadline force-resolve from a real exit. |
| `gsd/tools/exec-tool.ts` | Surfaces `force_resolved` in tool details; renders `exit=timeout(force-killed)` in the summary. | Operator/agent can distinguish a force-killed D-state child from a clean timeout. |
| `pi-agent-core/src/harness/env/nodejs.ts` | Same SIGTERM→grace→SIGKILL escalation in the local kill helper. Keeps a deliberate local mirror of `SIGKILL_GRACE_MS` (this is the lowest layer and must not import `pi-coding-agent`) and exports it for a parity test. | Consistency on the non-GSD embedding path. Mirror is deliberate; a parity test prevents drift. |
| `pi-agent-core/src/index.ts` | Re-export the mirror as `NODE_ENV_SIGKILL_GRACE_MS`. | Lets a higher-layer test lock it against the canonical value. |
| `bg-shell/process-manager.ts` | Added `terminateProcess` (graceful ladder) for `kill`/`restart`/session cleanup; kept `killProcess` for exact single-signal sends and synchronous `cleanupAll`. Removed unused `killSessionProcesses`. | `kill`/`restart`/cleanup should give a clean-shutdown window; an explicit `signal` send and exit-handler teardown should not (deferred timers don't fire during process exit). |
| `bg-shell/bg-shell-tool.ts`, `bg-shell-command.ts`, `overlay.ts` | Their kill paths call `terminateProcess` and poll for death (grace + slack) instead of a fixed SIGTERM-then-SIGKILL sleep. | Route every bg-shell kill through the one graceful ladder. |

### D. TUI fix (pre-existing bug on the same ESC/await surface)

| File | What changed | Why |
| --- | --- | --- |
| `gsd-agent-modes/.../components/tool-execution.ts` | `completeWithError` no-ops on an already-successful, non-partial result. | Pressing ESC during a *later* await used to flip every still-tracked tool — including already-succeeded ones — to red. |
| `gsd-agent-modes/.../controllers/chat-controller.ts` | Documented why a completed tool component is intentionally left in `pendingTools` (the message_end rebuild relies on it; the real protection is the `completeWithError` no-op). | Prevents a future "fix" from re-introducing the regression. |
| `gsd-agent-modes/.../interactive-chat-render.ts` | On an aborted/errored turn, look ahead for a matching `toolResult` and render the true result instead of discarding it and painting the row aborted. | Session-replay was the worse variant of the same bug — it dropped the real result entirely. |

---

## Change type

- [x] `feat` — uncapped sync bash, interruptible await, decision-rule guidance
- [x] `fix` — SIGTERM-immune async job leak, D-state hang, TUI ESC mis-mark, cancelled-status clobber
- [x] `refactor` — DRY'd the kill-timing constants to one source of truth
- [x] `test` — new behavior-executing tests

## Scope

- [x] `pi-agent-core` — local kill helper only
- [x] `pi-coding-agent` — coding agent
- [x] `gsd extension` — tool guidance, system.md, exec-sandbox, bg-shell
- [x] `pi-tui` — gsd-agent-modes tool cards

## Breaking changes

- [x] Yes

- **Removed the silent 120s synchronous-`bash` cap.** Commands that previously timed out at 120s now run uncapped (hangs owned by the auto-mode watchdog in auto-mode, or human ESC interactively). Behavior change for any embedding that relied on the implicit cap.
- **`pi-agent-core` `nodejs.ts` kill helper:** bare SIGKILL → graceful ladder. Behavior-preserving safety alignment; flagged for reviewer visibility given that package's RFC-sensitivity.
- **Removed the internal `killSessionProcesses` export** (no consumers).

## Test plan

- [x] CI passes
- [x] Manual testing — described above

Behavior-executing tests only (no source-grep). Local run on the rebased base — **105 tests pass** across the affected suites. New/updated test files, each tied to the change it covers:

| Test file | Verifies |
| --- | --- |
| `gsd/tests/dynamic-bash-no-cap.test.ts` (new) | GSD bash exposes no injected default timeout; advertises the watchdog; async_bash guidance drops the sync-cap implication; explicit timeout still forwarded |
| `bg-shell/tests/lifecycle-and-utilities.test.ts` | clean exit-0 reads as completion (points at async_bash), not a crash; graceful `terminateProcess` force-kills a SIGTERM-immune process |
| `async-jobs/await-tool.test.ts` (new) | await is interruptible; aborted job resurfaces; no duplicate-in-context; suppression only for completed jobs |
| `async-jobs/async-bash-cancel.test.ts` (new) | graceful cancel; SIGTERM-immune timeout-path SIGKILL regression; cancel-path D-state no-hang; **cancelled-status persistence regression** (proven to fail without the `.then` guard) |
| `gsd/tests/exec-graceful-kill.test.ts` (new) | exec-sandbox cooperative/non-closing kill timing, no hang |
| `gsd/tests/exec-tool.test.ts` | `force_resolved` rendered as `timeout(force-killed)` |
| `pi-coding-agent/.../kill-process-tree.test.ts` (new) | SIGTERM-cooperative/resistant escalation timing |
| `pi-coding-agent/.../bash-hard-deadline.test.ts` (new) | D-state child force-resolves, no hang |
| `pi-coding-agent/.../kill-constant-parity.test.ts` (new) | **cross-layer constant parity lock** (proven to fail when the mirror is perturbed) |
| `pi-agent-core/.../nodejs-kill-graceful.test.ts` (new) | SIGTERM→SIGKILL escalation, no hang |
| `gsd-agent-modes/.../chat-controller.test.ts`, `interactive-chat-render.test.ts` | ESC keeps a completed tool green, marks a pending tool interrupted; replay preserves the real result |

`typecheck:extensions`, `build:core`, and the three changed-package typechecks are clean.

Reviewer note: a bare `tsc --noEmit` from `packages/pi-coding-agent/` reports stale-`dist/` dual-module-identity errors in the untouched `package-commands.ts`; the package's real `build` (cleans `dist/` first) is exit 0. Use `pnpm run verify:merge` for the authoritative gate.

## AI disclosure

- [x] This PR includes AI-assisted code. Tooling: GSD-pi agent. All changes tested to the same standard as human-written code (behavior-executing tests, full affected-suite run, two proven regression tests). The AI is not credited as an author or co-author.

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-pi/pr/681"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783798682&installation_id=134682257&pr_number=681&repository=open-gsd%2Fgsd-pi&return_to=https%3A%2F%2Fgithub.com%2Fopen-gsd%2Fgsd-pi%2Fpull%2F681&signature=6233c908da795c0e0e4bf7a9fc56234dbe3d0de34912ce13d07ad992225917a2"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/packages/gsd-agent-modes/src/modes/interactive/components/tool-execution.ts
+++ b/packages/gsd-agent-modes/src/modes/interactive/components/tool-execution.ts
@@ -674,6 +674,11 @@ export class ToolExecutionComponent extends Container {
 
 	/**
 	 * Finalize a pending tool call as failed/interrupted while preserving any streamed partial output.
+	 *
+	 * Guard: a tool that already produced a settled, non-partial result must NOT be
+	 * touched just because the surrounding turn was aborted (e.g. the user pressed
+	 * ESC during a later tool's await). Only genuinely-incomplete tool calls should
+	 * be marked interrupted.
 	 */
 	completeWithError(message?: string): void {
 		if (this.result && !this.isPartial) {
--- a/packages/gsd-agent-modes/src/modes/interactive/controllers/chat-controller.test.ts
+++ b/packages/gsd-agent-modes/src/modes/interactive/controllers/chat-controller.test.ts
@@ -773,3 +773,85 @@ test("handleAgentEvent: agent_end finalizes orphaned pending tool cards", async 
 	assert.doesNotMatch(rendered, /running/, "agent_end must not leave stale running tool cards");
 	assert.match(rendered, /success/, "orphaned tool card should settle as no-result success");
 });
+
+test("handleAgentEvent: aborting a turn does NOT flip an already-completed tool to error (ESC regression)", async () => {
+	// Repro: a background-bash tool finishes successfully, then the user presses
+	// ESC during a later await. The turn ends with stopReason "aborted". The
+	// already-finished tool must stay a success — only genuinely-pending tools
+	// should be marked interrupted.
+	initTheme("dark", false);
+	const chatContainer = new Container();
+	const host = createStreamingHost(chatContainer);
+
+	const message = {
+		id: "turn-1",
+		role: "assistant",
+		provider: "claude-code",
+		model: "claude-opus-4-8",
+		timestamp: 1,
+		stopReason: "aborted",
+		content: [{ type: "toolCall", id: "bg-1", name: "async_bash", arguments: { command: "echo hi" } }],
+	};
+
+	await handleAgentEvent(host, {
+		type: "tool_execution_start",
+		toolCallId: "bg-1",
+		toolName: "async_bash",
+		args: { command: "echo hi" },
+	} as any);
+	await handleAgentEvent(host, {
+		type: "tool_execution_end",
+		toolCallId: "bg-1",
+		toolName: "async_bash",
+		result: { content: [{ type: "text", text: "Background job started: bg-1" }], isError: false },
+		isError: false,
+	} as any);
+
+	// Precondition: the completed tool rendered as success before the abort.
+	assert.match(
+		stripAnsi(chatContainer.render(100).join("\n")),
+		/success/,
+		"precondition: completed tool should render success before abort",
+	);
+
+	await handleAgentEvent(host, { type: "message_end", message } as any);
+
+	const rendered = stripAnsi(chatContainer.render(100).join("\n"));
+	assert.match(rendered, /success/, "completed tool must stay green after the turn is aborted");
+	assert.doesNotMatch(
+		rendered,
+		/Operation aborted/,
+		"a completed tool must not be labelled aborted just because the turn was",
+	);
+});
+
+test("handleAgentEvent: aborting a turn DOES mark a still-pending tool as interrupted", async () => {
+	// Counterpart to the regression above: a tool that never produced a result
+	// must still be marked interrupted when the turn aborts.
+	initTheme("dark", false);
+	const chatContainer = new Container();
+	const host = createStreamingHost(chatContainer);
+
+	const message = {
+		id: "turn-2",
+		role: "assistant",
+		provider: "claude-code",
+		model: "claude-opus-4-8",
+		timestamp: 1,
+		stopReason: "aborted",
+		content: [{ type: "toolCall", id: "await-1", name: "await_job", arguments: {} }],
+	};
+
+	await handleAgentEvent(host, {
+		type: "tool_execution_start",
+		toolCallId: "await-1",
+		toolName: "await_job",
+		args: {},
+	} as any);
+	// No tool_execution_end — the tool is still pending when the abort lands.
+
+	await handleAgentEvent(host, { type: "message_end", message } as any);
+
+	const rendered = stripAnsi(chatContainer.render(100).join("\n"));
+	assert.match(rendered, /Operation aborted/, "a genuinely-pending tool must be marked interrupted on abort");
+});
--- a/packages/gsd-agent-modes/src/modes/interactive/controllers/chat-controller.ts
+++ b/packages/gsd-agent-modes/src/modes/interactive/controllers/chat-controller.ts
@@ -1371,6 +1371,13 @@ export async function handleAgentEvent(host: InteractiveModeStateHost & {
 			const component = host.pendingTools.get(event.toolCallId);
 			if (component) {
 				component.updateResult({ ...event.result, isError: event.isError });
+				// NOTE: the component is intentionally left in host.pendingTools.
+				// The message_end rebuild path relies on finding already-completed
+				// components there to re-attach them; removing it here makes the
+				// rebuild synthesize a fresh empty component that a turn-abort would
+				// then mark errored. The real protection against downgrading a
+				// finished tool lives in ToolExecutionComponent.completeWithError,
+				// which no-ops on an already-successful result.
 				replaceCompactToolRowsWithPhaseSummary(host);
 				host.ui.requestRender();
 			}
--- a/packages/gsd-agent-modes/src/modes/interactive/interactive-chat-render.test.ts
+++ b/packages/gsd-agent-modes/src/modes/interactive/interactive-chat-render.test.ts
@@ -7,6 +7,7 @@ import type { AgentMessage } from "@gsd/pi-agent-core";
 import type { AssistantMessage } from "@gsd/pi-ai";
 import { initTheme } from "@gsd/pi-coding-agent/theme/theme.js";
 import { Container } from "@gsd/pi-tui";
+import stripAnsi from "strip-ansi";
 
 import { AssistantMessageComponent } from "./components/assistant-message.js";
 import { MAX_CHAT_COMPONENTS } from "./interactive-mode-class-constants.js";
@@ -26,6 +27,7 @@ function createHost(): InteractiveModeDelegateHost {
 		getMarkdownThemeWithSettings: () => undefined,
 		getRegisteredToolDefinition: () => undefined,
 		formatWebSearchResult: () => "",
+		toolOutputExpanded: true,
 		session: { retryAttempt: 0 },
 		editor: {},
 		footer: { invalidate() {} },
@@ -99,5 +101,71 @@ describe("interactive chat trimming", () => {
 		assert.ok(firstChild instanceof AssistantMessageComponent);
 		assert.equal(host.chatContainer.children.length, MAX_CHAT_COMPONENTS);
 		assert.equal(connectedToUser(firstChild), false);
+	});
+});
+
+describe("interactive chat replay: aborted turns preserve completed tool results", () => {
+	test("a tool that completed is rendered with its real result even though the turn aborted", () => {
+		// On reload/replay, an aborted assistant turn carries its tool calls, and
+		// each completed tool's result lives in a later `toolResult` message. The
+		// replay must surface that real result instead of discarding it and
+		// painting the row as "Operation aborted".
+		const host = createHost();
+
+		const assistant = {
+			id: "a-abort",
+			role: "assistant",
+			provider: "test",
+			model: "test-model",
+			timestamp: 1,
+			stopReason: "aborted",
+			content: [
+				{ type: "toolCall", id: "bg-1", name: "async_bash", arguments: { command: "echo hi" } },
+			],
+		} as unknown as AgentMessage;
+		const toolResult = {
+			id: "tr-1",
+			role: "toolResult",
+			toolCallId: "bg-1",
+			toolName: "async_bash",
+			timestamp: 2,
+			content: [{ type: "text", text: "UNIQUE_RESULT_TOKEN" }],
+			isError: false,
+		} as unknown as AgentMessage;
+
+		renderSessionContext(host, { messages: [assistant, toolResult] } as any);
+
+		const rendered = stripAnsi(host.chatContainer.render(100).join("\n"));
+		assert.match(rendered, /UNIQUE_RESULT_TOKEN/, "the completed tool's real result must survive replay");
+		assert.doesNotMatch(
+			rendered,
+			/Operation aborted/,
+			"a tool that actually completed must not be shown as aborted on replay",
+		);
+	});
+
+	test("a tool with no result on an aborted turn is still shown as interrupted", () => {
+		const host = createHost();
+
+		const assistant = {
+			id: "a-abort-2",
+			role: "assistant",
+			provider: "test",
+			model: "test-model",
+			timestamp: 1,
+			stopReason: "aborted",
+			content: [
+				{ type: "toolCall", id: "await-1", name: "await_job", arguments: {} },
+			],
+		} as unknown as AgentMessage;
+
+		renderSessionContext(host, { messages: [assistant] } as any);
+
+		const rendered = stripAnsi(host.chatContainer.render(100).join("\n"));
+		assert.match(
+			rendered,
+			/Operation aborted/,
+			"a tool with no result on an aborted turn should render as interrupted",
+		);
 	});
 });
--- a/packages/gsd-agent-modes/src/modes/interactive/interactive-chat-render.ts
+++ b/packages/gsd-agent-modes/src/modes/interactive/interactive-chat-render.ts
@@ -248,7 +248,22 @@ export function renderSessionContext(host: InteractiveModeDelegateHost,
 						component.setExpanded(host.toolOutputExpanded);
 						host.chatContainer.addChild(component);
 
-						if (message.stopReason === "aborted" || message.stopReason === "error") {
+						// On an aborted/errored turn, only the tool calls that never
+						// produced a result should render as interrupted. A tool that
+						// actually completed has its result in a later `toolResult`
+						// message (keyed by toolCallId) — register it as pending so the
+						// normal toolResult handler below renders the TRUE result with
+						// its real isError flag. Otherwise the successful result would be
+						// silently discarded and the row shown red.
+						const turnAbortedOrErrored =
+							message.stopReason === "aborted" || message.stopReason === "error";
+						const hasRealResult =
+							turnAbortedOrErrored &&
+							sessionContext.messages.some(
+								(m) => m.role === "toolResult" && m.toolCallId === content.id,
+							);
+
+						if (turnAbortedOrErrored && !hasRealResult) {
 							let errorMessage: string;
 							if (message.stopReason === "aborted") {
 								const retryAttempt = host.session.retryAttempt;
--- a/packages/pi-agent-core/src/harness/env/nodejs.ts
+++ b/packages/pi-agent-core/src/harness/env/nodejs.ts
@@ -189,29 +189,60 @@ function getShellEnv(baseEnv?: NodeJS.ProcessEnv, extraEnv?: Record<string, stri
 	};
 }
 
+// Grace period before escalating from SIGTERM to SIGKILL. Deliberately a local
+// mirror of @gsd/pi-coding-agent shell.ts SIGKILL_GRACE_MS: pi-agent-core is the
+// lowest layer and must not depend on pi-coding-agent, so the value cannot be
+// imported. Exported so a higher-layer test can lock it against the canonical
+// source and fail CI on drift (see
+// packages/pi-coding-agent/src/utils/kill-constant-parity.test.ts).
+export const SIGKILL_GRACE_MS = 5_000;
+
+// Note: NodeExecutionEnv.exec does not impose a hard exec deadline — that is intentionally out of
+// scope here; the sync-bash hard deadline in shell.ts covers the agent-facing case. This function
+// only provides graceful SIGTERM→SIGKILL escalation for consistency with the sync-bash kill path.
 function killProcessTree(pid: number): void {
 	if (process.platform === "win32") {
+		// No deliverable graceful signal for hidden console processes; force-kill the
+		// whole tree immediately (matches shell.ts killProcessTree on Windows).
 		try {
-			spawn("taskkill", ["/F", "/T", "/PID", String(pid)], {
+			const tk = spawn("taskkill", ["/F", "/T", "/PID", String(pid)], {
 				stdio: "ignore",
 				detached: true,
 				windowsHide: true,
 			});
+			tk.unref();
 		} catch {
 			// Ignore errors.
 		}
 		return;
 	}
 
+	// Send SIGTERM to the process group first; well-behaved processes exit promptly.
 	try {
-		process.kill(-pid, "SIGKILL");
+		process.kill(-pid, "SIGTERM");
 	} catch {
 		try {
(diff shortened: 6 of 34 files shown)


aidanthewiz:
## ✅ Risk Verification — PASS

**Commit verified:** `e18f551e4366881cb526decd9a6ef1454f04dcfc` · **Date:** 2026-06-11 · **Verified by:** GSD-pi agent (Claude)

Verified in an isolated detached worktree pinned to the immutable PR-head SHA (the shared worktree drifted to another commit mid-review — re-pinned and re-ran everything; the SHA above was re-confirmed clean after the final suite).

| Agent Core | check for race conditions and loop exit paths in the agent loop | ✅ |
| Tool System | verify tool call and response contracts are unchanged | ✅ |
| Modes | test mode switching transitions and state isolation | ✅ |
| AI Providers | confirm provider API contract and error handling are intact | ✅ |
| GSD Workflow | verify GSD workflow state transitions end-to-end | ✅ |

### Evidence
All suites run against the pinned SHA. **105/105 pass**, typecheck clean.

| Check | Command | Result |
|-------|---------|--------|
| async_jobs: await interruptible + cancel/`.then`-guard regressions | `node --test await-tool.test.ts async-bash-cancel.test.ts` | 20/20 pass |
| bg-shell: clean-exit-0 reads as completion + graceful `terminateProcess` | `node --test bg-shell/tests/lifecycle-and-utilities.test.ts` | 31/31 pass |
| gsd exec/bash: no injected default timeout, explicit-timeout passthrough, force-resolve | `node --test dynamic-bash-no-cap/exec-graceful-kill/exec-tool.test.ts` | 10/10 pass |
| Modes/TUI: ESC keeps completed tool green, marks pending interrupted, replay preserves result | `node --test chat-controller.test.ts interactive-chat-render.test.ts` | 37/37 pass |
| kill ladder: SIGTERM→grace→SIGKILL escalation, D-state hard-deadline, cross-layer parity lock | `node --test kill-process-tree/kill-constant-parity/bash-hard-deadline.test.ts` | 5/5 pass |
| pi-agent-core local kill helper escalation | `vitest run --config vitest.harness.config.ts nodejs-kill-graceful.test.ts` | 2/2 pass |
| typecheck (Tool System + GSD Workflow contracts) | `pnpm run typecheck:extensions` | 0 errors |

### Root cause & scope
- **Root cause:** Agents reflexively reached for `async_bash` + immediate `await_job` (and `bg_shell` for run-to-completion commands) because a **silent 120s cap on synchronous `bash`** (injected in `gsd/bootstrap/dynamic-tools.ts` via `DEFAULT_BASH_TIMEOUT_SECS`) made sync bash secretly time-limited, while the guidance framed async as "the safe choice." Symptom = pointless round-trips and an un-interruptible `await_job`; true cause = tool behavior not matching tool description. Fix removes the cap, rewrites guidance into a Quick(`bash`)/One-shot(`async_bash`)/Background(`bg_shell`) decision rule, makes `await_job` interruptible, and backs the now-uncapped bash with a SIGTERM→grace→SIGKILL ladder + caller-side hard deadline so a timeout/ESC can't corrupt a stateful command or hang on a D-state child.
- **Scope:** Tool call/response **contracts are unchanged** — `bash` schema, `await_job` schema/description, and the `dynamicBash.execute` signature are all identical; only the silent default-timeout side-effect was removed (params now pass through verbatim). `DEFAULT_BASH_TIMEOUT_SECS` removal is complete (zero dangling references). The duplicated-logic search found the deliberate `SIGKILL_GRACE_MS = 5000` mirror in `pi-agent-core/nodejs.ts` (cannot import `pi-coding-agent`); a new cross-layer parity test locks it against drift. Removed `killSessionProcesses` export has **no remaining consumers**.
- **Downstream/stale deps:** None. The **"AI Providers" critical flag is a classifier false positive** — it tags `pi-coding-agent/src/index.ts` and `pi-agent-core/src/index.ts` as AI Providers, but the only changes there are 11 lines of **additive** re-exports (`killProcessTree`, `SIGKILL_GRACE_MS`, `HARD_DEADLINE_MS`, `NODE_ENV_SIGKILL_GRACE_MS`); no provider API or error-handling code was touched (`pnpm` diff over `pi-ai/`/provider files is empty).

### State transitions confirmed
- **async job status:** running → cancelled now **stays** cancelled after the runFn resolves (the `.then` guard mirrors the existing `.catch` guard; async_bash resolves-not-rejects on abort, which previously clobbered the status back to `completed`). Proven by the "status STAYS 'cancelled'" regression test.
- **await_job race:** running → (completed | timeout | abort) with explicit teardown of the race timer + abort listener on settle; abort leaves still-running jobs **unsuppressed** so they resurface via `onJobComplete`.
- **bash kill ladder:** alive → SIGTERM → 5s grace → SIGKILL → (close | hard-deadline force-resolve) for a non-closing D-state child.
- **TUI tool card:** completed(green) stays green when a *later* turn is ESC-aborted; only still-pending cards flip to interrupted; session-replay renders the real `toolResult` instead of painting the row aborted.

### Notes / caveats
- Verification ran in a fresh isolated worktree; workspace `dist/` + native addon were copied from the same-SHA main checkout (no rebuild from a different commit). The `nodejs-kill-graceful` suite is a vitest-harness test (not `node:test`) and was run under `vitest.harness.config.ts` accordingly.
- Did not run the full `verify:merge` gate (web-host build + e2e) locally — CI owns that. The verification here is scoped to the five flagged systems' behavior-executing suites + extension typecheck, all green.

PR #785, opened 2026-06-20
fix(gsd): gate provider-keyed web search tools on configured credentials

### LINKED ISSUE #784: [Bug]: Provider-keyed web search tools are offered even with no API key — the agent reaches for a tool that can only error
# [Bug]: Provider-keyed web search tools are offered even with no API key — the agent reaches for a tool that can only error

## Summary

The custom web-search tools (`search-the-web` and `search_and_read`, backed by Tavily/Brave/Ollama) are registered and advertised to the model **even when no provider API key is configured**. With no key they can only return an `auth_error`, so the agent confidently reaches for them and gets `✗ No search API key set [auth_error]`. The same pattern applies to the standalone `google_search` tool (Gemini-backed) when no Gemini credentials are set. Tools that cannot authenticate should not be offered to the model at all.

## Steps to reproduce

1. Run GSD with **no** `TAVILY_API_KEY` / `BRAVE_API_KEY` / `OLLAMA_API_KEY` set, on a model/provider that does not get native Anthropic web search (e.g. `claude-opus-4-8` served through a non-native transport, or any non-Anthropic model).
2. Ask the agent to do something that needs the web (e.g. "verify the latest version of X against docs").
3. Observe the agent call `search-the-web` (and/or `search_and_read`).
4. Result: `✗ No search API key set [auth_error]` — the tool was offered, chosen, and immediately failed.

## Expected behavior

When no search provider key is configured, `search-the-web` and `search_and_read` should not be presented to the model at all — they should be unknown, so the agent never reaches for them and instead uses `fetch_page` (which works key-free via Jina Reader) or native Anthropic web search where available. `google_search` should likewise only be offered when Gemini credentials (`GEMINI_API_KEY` or Google OAuth) are present.

## Actual behavior

The web-search extension registers `search-the-web`, `search_and_read`, and `fetch_page` unconditionally at `session_start`; the standalone `google_search` extension registers its tool unconditionally on load. `resolveSearchProvider()` already returns `null` when no key is set (the tool then errors at execute time), but registration does not consult it — so the tool stays in the model-facing surface and only fails when called. The native-search hook only hides the custom tools when *native Anthropic web search* takes over, so on any non-native provider with no key the broken tools remain visible.

This is made worse by the GSD auto-mode tool scoping: `buildMinimalAutoGsdToolSet` rebuilds the model-facing surface from `getAllTools()` (all *registered* tools, regardless of active state), so merely de-activating a tool is insufficient — a registered-but-inactive tool leaks back into the surface.

## Root cause

Registration is unconditional and decoupled from credential availability:

- `src/resources/extensions/search-the-web/index.ts` registers all three tools in one deferred `session_start` block, never checking `resolveSearchProvider()`.
- `src/resources/extensions/search-the-web/native-search.ts` re-enables the custom search tools on switch-away-from-Anthropic "even if keys are missing, so the user sees the error rather than tools silently vanishing" — the opposite of what's wanted here.
- `extensions/google-search/index.ts` registers `google_search` in the extension's default export with only an execute-time credential check.

## GSD version

1.2.0

## Node.js version

v24.11.0

## Operating system

macOS

## Install method

npm link / local checkout

## Affected area

MCP / tools / extensions (web-search extension + standalone google-search extension)

## Model and provider

Observed with `claude-opus-4-8`; reproduces on any model/provider that does not receive native Anthropic web search when no search key is set.

## Impact

- Severity: medium. Not a crash, but a recurring workflow-quality and trust problem — the agent picks a tool that cannot work, burns a turn on an `auth_error`, and (in research/verification flows) can be derailed mid-task.
- Frequency: every session without a search key, on any non-native-search provider.
- Consequence: wasted tool calls, confusing `No search API key set` errors mid-research, and an agent reaching for capabilities it does not have.

## Additional information

- `fetch_page` works without any key (Jina Reader, with a direct-fetch fallback), so it should remain available regardless.
- The bundled `src/resources/extensions/google-search/` is already a deprecation no-op stub; the unconditional `google_search` registration lives only in the not-yet-published standalone `@gsd-extensions/google-search` package, so the google_search half of this fix is forward-looking.
- Fix approach: gate **registration** (not active-state) on credential availability so the tools are never advertised when they cannot authenticate; keep `fetch_page` always registered.

aidanthewiz:
## Linked issue

Closes #784

---

## TL;DR

**What:** Only register the provider-keyed web-search tools when they can actually authenticate — `search-the-web` / `search_and_read` when a Tavily/Brave/Ollama key is set, and `google_search` when Gemini credentials are present — so the agent is never offered a search tool that can only return an auth error.
**Why:** With no key configured these tools were still advertised to the model, so the agent reached for them and got `No search API key set [auth_error]` mid-task (observed on `claude-opus-4-8` with no key).
**How:** Gate **registration** (not active-state) on credential availability in each extension's `session_start`; keep `fetch_page` always registered because it works key-free via Jina Reader; make the native-search re-enable path provider-aware so an unusable tool is never re-advertised.

This is a **tool-surface / agent-correctness** change. There is no change to how any tool behaves once a key *is* present.

- Stops advertising `search-the-web` / `search_and_read` when no Tavily/Brave/Ollama key is configured
- Stops advertising `google_search` when no Gemini credential (`GEMINI_API_KEY` or Google OAuth) is present
- Keeps `fetch_page` always available (key-free via Jina Reader)
- Makes the native-search "re-enable on switch away from Anthropic" path provider-aware, so it never re-adds an unusable Brave tool (or a schema-less phantom name)
- Re-evaluates the gate each `session_start`, so a key/login added before a later session is picked up without a restart

## Why (the root problem)

`resolveSearchProvider()` already returns `null` when no key is set — but registration never consulted it, so the tools stayed in the model-facing surface and only failed at execute time with `auth_error`. The native-search hook only hides the custom tools when *native Anthropic web search* takes over, so on any non-native provider with no key the broken tools remained visible and the agent kept choosing them.

Gating **registration** (rather than just de-activating) is required because GSD auto-mode rebuilds the model surface from `getAllTools()` — *all registered tools*, regardless of active state — so a registered-but-inactive tool leaks straight back into the surface. The only reliable way to make a tool "unknown" to the model is to not register it. (Verified: `setActiveToolsByName` in `gsd-agent-core` filters every name through `_toolRegistry.get(name)`, so an unregistered name is silently dropped — which is also why a stray re-enabled phantom name is harmless, but we still avoid it for a clean active set.)

### A. Provider gating — `search-the-web` / `search_and_read` (Tavily/Brave/Ollama)

| File | What changed | Why |
| --- | --- | --- |
| `src/resources/extensions/search-the-web/index.ts` | Split registration into two memoized paths: `fetch_page` registers unconditionally; `search-the-web` + `search_and_read` register only when `resolveSearchProvider() !== null`. The provider gate is re-checked on each `session_start` (its own memo), so a key added before a later session is picked up. | Without a provider key these two tools can only return an auth error. Gating registration (not active-state) is required because GSD auto-mode rebuilds the surface from `getAllTools()`. `fetch_page` stays because it works key-free via Jina Reader. |
| `src/resources/extensions/search-the-web/native-search.ts` | Made the `model_select` "re-enable custom search tools when switching away from Anthropic" branch provider-aware: the Brave/Tavily/Ollama tools are re-added only when `resolveSearchProvider() !== null`; `google_search` (its own creds) is still re-added. | Reverses the prior "re-enable even if keys missing" behavior so an unusable tool is never re-advertised, and keeps the active set free of schema-less phantom names. |

### B. Credential gating — `google_search` (Gemini), standalone `@gsd-extensions/google-search`

| File | What changed | Why |
| --- | --- | --- |
| `extensions/google-search/index.ts` | Moved the `pi.registerTool(...)` body into `registerGoogleSearchTool(pi)` and gate it in `session_start` on credentials — `GEMINI_API_KEY` env or `google-gemini-cli` OAuth via `ctx.modelRegistry.authStorage.hasAuth`. Used a per-`pi` `WeakSet` memo so a single real process registers once across repeated `session_start` fires, while tests that construct a fresh `pi` per case each register cleanly. | Consistency with the search-the-web gating; an unconfigured `google_search` could only return an auth error. The bundled `src/resources/extensions/google-search/` is already a deprecation no-op stub, so this affects only the not-yet-published standalone package — making it correct on arrival. |

### C. Tests

| File | What changed | Why |
| --- | --- | --- |
| `src/tests/search-tool-registration-gating.test.ts` (new) | No provider key → `search-the-web` / `search_and_read` are NOT registered, `fetch_page` is; with a provider key → both register. Hermetic (isolated `GSD_HOME`, cleared keys). | Pins the core registration contract. |
| `src/tests/native-search.test.ts` (updated) | Split the re-enable test into provider-present (Brave tools re-enable on switch-away) and no-provider (Brave tools stay hidden; `google_search` still re-enabled) cases; the dedup-across-toggle-cycles test now sets a key so it still exercises dedup. | Encodes the new provider-aware re-enable behavior. |
| `src/tests/google-search-auth.repro.test.ts` (updated) | The no-credentials case now asserts the tool is NOT registered + a warning is shown (previously: registered, then an execute-time auth error). OAuth and `GEMINI_API_KEY` paths still register and execute. | Reflects the new gated contract; the old test encoded the "register always, error at execute" behavior this PR removes. |

## Change type

- [x] `fix` — provider/credential-gate web-search tool registration so unusable tools aren't advertised
- [x] `test` — registration-gating coverage + updated native-search and google-search contracts

## Scope

- [x] `gsd extension` — web-search extension (`search-the-web`) and standalone google-search extension

## Breaking changes

- [x] No breaking changes

The only behavior change is to the **model-facing tool surface**: tools that could not authenticate were previously advertised (and errored on use) and are now simply not offered. No public API, CLI, or config change. Behavior with a key present is unchanged. `fetch_page` is unaffected.

## Test plan

- [x] CI passes
- [x] Manual testing — described below

Verified locally (compiled unit harness):

| Test file | Verifies |
| --- | --- |
| `src/tests/search-tool-registration-gating.test.ts` (new) | no key → search tools not registered, `fetch_page` is; key → registered |
| `src/tests/native-search.test.ts` | provider-aware re-enable (key vs no-key), dedup across toggles, native injection unchanged |
| `src/tests/google-search-auth.repro.test.ts` | no-creds → not registered + warn; OAuth / `GEMINI_API_KEY` → registered + executes |
| `src/tests/google-search-oauth-shape.test.ts` | OAuth Cloud Code Assist wire format unchanged (creds present) |
| `src/resources/extensions/gsd/tests/google-search-stub.test.ts` | bundled deprecation stub remains a no-op |

- Targeted suites above: **60 / 60 pass**.
- `pnpm run verify:merge`: all **code-quality gates pass** — `build:core`, `build:web-host`, `typecheck:extensions`, `validate-pack`, `verify:workspace-coverage`, `verify:extension-coverage`.
- Full `pnpm run test:unit`: 10911 pass / 23 fail — **no new failures vs a clean-HEAD baseline** (clean `main` = 24; **zero failures in any changed area**). The 23 are pre-existing/environmental (a non-default local GSD home, ambient provider auth in the env, the `rtk` managed-bin path, and a borderline prompt-size gate).
- `test:packages` + `test:integration`: **baseline-confirmed** in the same checkout — 81 failing on this branch vs 79 on clean `main`, the sole delta being a flaky CLI-spawn smoke test that passes 3/3 in isolation. **Zero failures in any changed area.** (These suites are env-sensitive — git-worktree / `.gsd`-state / built-CLI integration tests fail identically on clean `main` in this local setup.)
- Runtime smoke: the built CLI loads the full extension stack cleanly; the gating itself is verified by driving the **compiled** extension through the real `session_start` flow with registration awaited (no key → `search-the-web` / `search_and_read` not registered, `fetch_page` is; key → registered).
- Re-verified end-to-end in a dedicated worktree (`pnpm install` → `build:pi` → `build:core`).

Reviewer note: the local environment (non-default GSD home + a linked git worktree) makes a fully-green `verify:merge` impossible locally — the env-sensitive worktree/`.gsd`/CLI integration suites fail identically on clean `main`. CI is the authoritative green gate; every failure observed locally is pre-existing and environmental, none in a changed area.

## AI disclosure

- [x] This PR includes AI-assisted code. Tooling: GSD-pi agent. All changes were tested to the same standard as human-written code (behavior-executing tests, full affected-suite run, clean-baseline comparison). The AI is not credited as an author or co-author.

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-pi/pr/785"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784511433&installation_id=134682257&pr_number=785&repository=open-gsd%2Fgsd-pi&return_to=https%3A%2F%2Fgithub.com%2Fopen-gsd%2Fgsd-pi%2Fpull%2F785&signature=bf2d34fb7aee41574275abb5088673c774156f089e97bb6d8c01be7c13b5d098"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/extensions/google-search/index.ts
+++ b/extensions/google-search/index.ts
@@ -174,7 +174,20 @@ function cacheKey(query: string): string {
 
 // ── Extension ────────────────────────────────────────────────────────────────
 
-export default function (pi: ExtensionAPI) {
+// Track which ExtensionAPI instances have already registered the tool, so a
+// real single-instance process registers once across repeated session_start
+// fires, while tests that construct a fresh pi per case each register cleanly.
+const googleSearchRegisteredFor = new WeakSet<object>();
+
+/**
+ * Register the google_search tool. Gated on credential availability by the
+ * session_start hook: the tool is never advertised to the model when it cannot
+ * authenticate (GEMINI_API_KEY or Google OAuth), so the agent does not reach
+ * for a tool that can only return an auth error. Idempotent per pi instance.
+ */
+function registerGoogleSearchTool(pi: ExtensionAPI): void {
+	if (googleSearchRegisteredFor.has(pi)) return;
+	googleSearchRegisteredFor.add(pi);
 	pi.registerTool({
 		name: "google_search",
 		label: "Google Search",
@@ -411,7 +424,9 @@ export default function (pi: ExtensionAPI) {
 			return new Text(text, 0, 0);
 		},
 	});
+}
 
+export default function (pi: ExtensionAPI) {
 	// ── Session cleanup ─────────────────────────────────────────────────────
 
 	pi.on("session_shutdown", async () => {
@@ -419,18 +434,33 @@ export default function (pi: ExtensionAPI) {
 		client = null;
 	});
 
-	// ── Startup notification ─────────────────────────────────────────────────
+	// ── Credential-gated registration ────────────────────────────────────────
+	// Only register google_search when it can actually authenticate. Without
+	// credentials the tool can only return an auth error, so presenting it to
+	// the model is pure confusion — the agent reaches for a tool that cannot
+	// work. Re-evaluated each session so a key or Google login added before a
+	// later session is picked up without a restart.
 
 	pi.on("session_start", async (_event, ctx) => {
-		if (process.env.GEMINI_API_KEY) return;
-
-		const hasOAuth = await ctx.modelRegistry.authStorage.hasAuth("google-gemini-cli");
-		if (!hasOAuth) {
-			ctx.ui.notify(
-				"Google Search: No authentication set. Log in via Google or set GEMINI_API_KEY to use google_search.",
-				"warning",
-			);
+		const hasGeminiKey = !!process.env.GEMINI_API_KEY;
+		let hasOAuth = false;
+		if (!hasGeminiKey) {
+			try {
+				hasOAuth = await ctx.modelRegistry.authStorage.hasAuth("google-gemini-cli");
+			} catch {
+				hasOAuth = false;
+			}
 		}
+
+		if (hasGeminiKey || hasOAuth) {
+			registerGoogleSearchTool(pi);
+			return;
+		}
+
+		ctx.ui.notify(
+			"Google Search: No authentication set. Log in via Google or set GEMINI_API_KEY to use google_search.",
+			"warning",
+		);
 	});
 }
 
--- a/src/resources/extensions/search-the-web/index.ts
+++ b/src/resources/extensions/search-the-web/index.ts
@@ -3,38 +3,74 @@
  *
  * Native Anthropic hooks stay eager. Heavy tool registration is deferred in
  * interactive mode so startup is not blocked on the full search tool stack.
+ *
+ * Provider gating: the Brave/Tavily/Ollama-backed tools (search-the-web,
+ * search_and_read) are only registered when a provider API key is actually
+ * configured (resolveSearchProvider() !== null). Without a key these tools can
+ * only return an auth error, so presenting them to the model is pure confusion
+ * — the agent reaches for a tool that cannot work. fetch_page is always
+ * registered because it works key-free via Jina Reader.
  */
 
 import { importExtensionModule, type ExtensionAPI } from "@gsd/pi-coding-agent";
 import { registerSearchProviderCommand } from "./command-search-provider.js";
 import { registerNativeSearchHooks } from "./native-search.js";
+import { resolveSearchProvider } from "./provider.js";
 
-let toolsPromise: Promise<void> | null = null;
+// fetch_page registration — always available (key-free via Jina Reader).
+let fetchToolPromise: Promise<void> | null = null;
+// search-the-web + search_and_read registration — gated on provider key.
+let searchToolsPromise: Promise<void> | null = null;
 let resetSearchLoopGuardStateRef: (() => void) | null = null;
 
-async function registerSearchTools(pi: ExtensionAPI): Promise<void> {
-  if (!toolsPromise) {
-    toolsPromise = (async () => {
-      const [
-        { registerSearchTool, resetSearchLoopGuardState },
-        { registerFetchPageTool },
-        { registerLLMContextTool },
-      ] = await Promise.all([
-        importExtensionModule<typeof import("./tool-search.js")>(import.meta.url, "./tool-search.js"),
-        importExtensionModule<typeof import("./tool-fetch-page.js")>(import.meta.url, "./tool-fetch-page.js"),
-        importExtensionModule<typeof import("./tool-llm-context.js")>(import.meta.url, "./tool-llm-context.js"),
-      ]);
-      resetSearchLoopGuardStateRef = resetSearchLoopGuardState;
-      registerSearchTool(pi);
+async function registerFetchTool(pi: ExtensionAPI): Promise<void> {
+  if (!fetchToolPromise) {
+    fetchToolPromise = (async () => {
+      const { registerFetchPageTool } = await importExtensionModule<typeof import("./tool-fetch-page.js")>(
+        import.meta.url,
+        "./tool-fetch-page.js",
+      );
       registerFetchPageTool(pi);
-      registerLLMContextTool(pi);
     })().catch((error) => {
-      toolsPromise = null;
+      fetchToolPromise = null;
       throw error;
     });
   }
 
-  return toolsPromise;
+  return fetchToolPromise;
+}
+
+/**
+ * Register the provider-backed search tools, but only when a search provider
+ * key is configured. Memoized so it registers at most once per process; the
+ * guard is re-evaluated on each session_start so a key added before a later
+ * session is picked up without a restart.
+ */
+async function registerProviderSearchTools(pi: ExtensionAPI): Promise<void> {
+  if (resolveSearchProvider() === null) return;
+  if (!searchToolsPromise) {
+    searchToolsPromise = (async () => {
+      const [
+        { registerSearchTool, resetSearchLoopGuardState },
+        { registerLLMContextTool },
+      ] = await Promise.all([
+        importExtensionModule<typeof import("./tool-search.js")>(import.meta.url, "./tool-search.js"),
+        importExtensionModule<typeof import("./tool-llm-context.js")>(import.meta.url, "./tool-llm-context.js"),
+      ]);
+      resetSearchLoopGuardStateRef = resetSearchLoopGuardState;
+      registerSearchTool(pi);
+      registerLLMContextTool(pi);
+    })().catch((error) => {
+      searchToolsPromise = null;
+      throw error;
+    });
+  }
+
+  return searchToolsPromise;
+}
+
+async function registerSearchTools(pi: ExtensionAPI): Promise<void> {
+  await Promise.all([registerFetchTool(pi), registerProviderSearchTools(pi)]);
 }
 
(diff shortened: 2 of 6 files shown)


aidanthewiz:
## ✅ Risk Verification — PASS

**Commit verified:** `918fe678081c81d85326a72a9432be173b0d194c` · **Date:** 2026-06-20 · **Verified by:** agent

| Agent Core | check for race conditions and loop exit paths in the agent loop | ✅ |
| AI Providers | confirm provider API contract and error handling are intact | ✅ |
| Tool System | verify tool call and response contracts are unchanged | ✅ |

### Evidence
| Check | Command | Result |
|-------|---------|--------|
| unit suite — PR commit | `npm test` (pretest+unit+integration+packages) | 10910/10934 pass (24 pre-existing failures, +3 vs baseline) |
| unit suite — main baseline | `npm test` on `main` | 10907/10931 pass (24 pre-existing failures) |
| typecheck | `pnpm run typecheck:extensions` (via pretest) | 0 errors |

### Root cause & scope
- **Root cause:** Extension registration was unconditional (tools registered at module load time regardless of credential availability). Active-set gating alone is insufficient because GSD auto-mode rebuilds the tool surface from `getAllTools()` (all *registered* tools), so a deactivated tool re-enters the model-facing surface on the next rebuild. The fix gates *registration itself* behind credential checks at `session_start`.
- **Scope:** Isolated to `extensions/google-search/index.ts`, `src/resources/extensions/search-the-web/index.ts`, and `src/resources/extensions/search-the-web/native-search.ts`. No changes to agent loop logic, tool call/response contracts, AI provider clients, or any other subsystem. Three new test files cover the gating behavior explicitly.
- **Downstream/stale deps:** `fetch_page` is split into its own always-on registration path (key-free via Jina Reader) — existing callers unaffected.

### Agent Core — race condition analysis
The `WeakSet<object>` guard in `google-search` is checked and updated synchronously within the async `session_start` handler; no await point separates the check from the add, so concurrent `session_start` fires on the same `pi` instance cannot both pass the guard. The `search-the-web` split into `fetchToolPromise`/`searchToolsPromise` uses two independent memoized promises — each has its own error-reset `catch`, and `Promise.all` on independent registrations introduces no shared mutable state. No new loop exit paths introduced; all changes are confined to extension registration hooks.

### Tool System — contract check
`pi.registerTool(...)` call signatures and all tool schemas/handlers are unchanged in both extensions. The only behavioral delta is *whether* registration occurs (gated on credentials), not how the registered tools behave.

### AI Providers — error handling
`authStorage.hasAuth()` in `google-search` is now wrapped in `try/catch` with a safe `false` fallback — previously it was called without error handling. `resolveSearchProvider()` reads environment keys at call time (not module load time), so it correctly reflects keys added before a later session without restart.

### Notes / caveats
- The 24 suite failures are all pre-existing on `main` (confirmed by identical failure set on baseline run). No failures introduced by this PR.
- Once tools are registered (credentials present at first `session_start`), they remain registered for the lifetime of that `pi` instance even if keys are later removed. This is intentional and documented in the extension comments.

Issue #648, opened 2026-06-10
[Bug]: Error: 400 {"error":{"message":"The use of the web search tool is not supported.","code":"unsupported_value"}}


mgerdov-igt:
### Bug type

Regression (worked before, now fails)

### Summary

upon GSD start initial model does not work with error:
Error: 400 {"error":{"message":"The use of the web search tool is not supported.","code":"unsupported_value"}}
Model has to be changed to any other value for this error to go away

### Steps to reproduce

1. Run GSD
2. ANy prompt fails with Error: 400 {"error":{"message":"The use of the web search tool is not supported.","code":"unsupported_value"}}
3.  Change model to any pother supported.
4. Error goes away

### Expected behavior

Error: 400 {"error":{"message":"The use of the web search tool is not supported.","code":"unsupported_value"}} should not appear on any model on startup

### Actual behavior

Any model fails with Error: 400 {"error":{"message":"The use of the web search tool is not supported.","code":"unsupported_value"}}

### GSD version

1.2.0

### Operating system

Windows 11

### Install method

npm global install

### Affected area

Auto-mode / dispatch loop

### Logs, screenshots, and evidence

```shell

```

jeremymcs:
### Issue evaluation - approved

Issue: [[Bug]: Error: 400 {"error":{"message":"The use of the web search tool is not supported.","code":"unsupported_value"}}](https://github.com/open-gsd/gsd-pi/issues/648)

Approved for automatic work: the evaluator could not disprove a concrete, surgical issue report and found no safety flags.

Confidence: high (82%)

**Safety flags**
- none

**Labels applied**
- `ready-for-agent`

jeremymcs:
⏳ **Issue work started** — beginning work on [#648 [Bug]: Error: 400 {"error":{"message":"The use of the web search tool is not supported.","code":"unsupported_value"}}](https://github.com/open-gsd/gsd-pi/issues/648).

- Repo: `open-gsd/gsd-pi`
- Issue: [#648 [Bug]: Error: 400 {"error":{"message":"The use of the web search tool is not supported.","code":"unsupported_value"}}](https://github.com/open-gsd/gsd-pi/issues/648)

jeremymcs:
❌ **Issue work failed** — [#648 [Bug]: Error: 400 {"error":{"message":"The use of the web search tool is not supported.","code":"unsupported_value"}}](https://github.com/open-gsd/gsd-pi/issues/648).

[...]

PR #446, opened 2026-06-03
Fix GSD workflow tool contract saves

jeremymcs:
## Linked issue

Closes #445

- [x] I have linked an issue above. I understand that PRs without a linked issue will be closed without review.

---

## TL;DR

**What:** Hardens GSD run-uat and validate-milestone save contracts so completed workflows submit complete structured payloads.
**Why:** UAT and milestone validation could pass their checks, then fail at save time because required presentation tools or planned verification rows were omitted.
**How:** Prompt/tool guidance now exposes complete canonical payload shapes, validators report all missing fields/rows together, and regression tests cover the contracts.

## What

- Updates the run-uat prompt to call `gsd_uat_result_save` with the full required presentation proof.
- Updates validate-milestone prompt context to preserve canonical planned verification class rows.
- Aggregates missing UAT presentation and validate-milestone verification-class errors.
- Tightens MCP/workflow tool schema guidance for the affected save calls.
- Adds regression coverage across prompt contracts, integration prompt rendering, MCP workflow tools, and validate-milestone write ordering.

## Why

The prompt/tool guidance could be looser than the validators. That allowed workflows to finish UAT or validation work, then get stuck in retry loops where each save call exposed one missing contract field or row at a time.

## How

The prompts now provide copyable canonical save/table shapes. The validators still enforce the contracts, but return complete missing-field or missing-row diagnostics so any remaining invalid call is actionable in one pass.

## Change type

- [ ] `feat` — New feature or capability
- [x] `fix` — Bug fix
- [ ] `refactor` — Code restructuring (no behavior change)
- [x] `test` — Adding or updating tests
- [ ] `docs` — Documentation only
- [ ] `chore` — Build, CI, or tooling changes

## Scope

- [ ] `pi-tui` — Terminal UI
- [ ] `pi-ai` — AI/LLM layer
- [ ] `pi-agent-core` — Agent orchestration
- [ ] `pi-coding-agent` — Coding agent
- [x] `gsd extension` — GSD workflow
- [ ] `native` — Native bindings
- [ ] `ci/build` — Workflows, scripts, config

## Breaking changes

- [x] No breaking changes
- [ ] Yes — described above

## Test plan

- [ ] CI passes
- [x] New/updated tests included
- [x] Manual testing — steps described below
- [ ] No tests needed — explained above

Local validation:

- [x] `pnpm install --frozen-lockfile --ignore-scripts`
- [x] `pnpm run build:pi`
- [x] `pnpm run build:rpc-client`
[...]

PR #792, opened 2026-06-20
fix(agent-core): forward resolved model in before_provider_request event

### LINKED ISSUE #791: [Bug]: Native Anthropic web search never activates — the resolved model is dropped from the `before_provider_request` event, so injection silently fails

## Summary

On a first-party Anthropic provider (`provider: anthropic`, `api: anthropic-messages`, e.g. `claude-opus-4-8`), native Anthropic web search is **never injected** into the request. The agent has no `web_search` tool, falls back to `fetch_page` for everything, and will tell the user it "doesn't have a web search tool." This happens even with `search_provider: native` and a fully native-capable model.

Root cause: the `onPayload` bridge in `gsd-agent-core/src/sdk.ts` receives the resolved model but discards it (`_model`) when calling `runner.emitBeforeProviderRequest(payload)`. The runner then builds `BeforeProviderRequestEvent` **without `model`**, even though the event type has always declared `model?: { provider, id, api? }`. The native-search hook reads `event.model` as its primary ("most reliable") provider-detection path, so with `model` undefined it can never confirm the provider and never injects `web_search_20250305`.

## Steps to reproduce

1. Configure an Anthropic API-key provider: `defaultProvider: anthropic`, `defaultModel: claude-opus-4-8`, `search_provider: native` (no Brave/Tavily key needed — native search is keyless).
2. Start GSD interactively.
3. Ask it to do something that needs the live web (e.g. "search the web for the latest stable Node.js version").
4. Observe the agent reason that it has no web-search tool and reach for `fetch_page` (or say it can't search).

## Expected behavior

The native `web_search_20250305` server tool is injected into the outgoing Anthropic request (it's keyless and server-side), the model calls it, and results come back inline — i.e. native web search works the way it does in Claude Code.

## Actual behavior

`web_search` is never in the request. Captured at runtime by logging the final `params.tools` immediately after the `onPayload` hook in `pi-ai`'s Anthropic provider, on `claude-opus-4-8` / `anthropic-messages`:

```
onPayload=true  tools: 141 -> 141  hasWebSearch=false      (x5 requests)
```

And logging native-search's own decision in `before_provider_request`:

```
bpr: hasEventModel=false  modelSelectFired=false  isAnthropicProvider=false
     payloadModelName=claude-opus-4-8  looksAnthropic=true  -> isAnthropic=false
```

So the hook runs (`onPayload=true`), but `event.model` is `undefined` (`hasEventModel=false`), `model_select` never fired (`modelSelectFired=false`), and the payload-name heuristic is correctly refused by the `#648` guard — so it hits the fail-safe `else` branch and never injects.

## Root cause

Two cooperating gaps; the first is the real defect:

1. **The model is dropped at the SDK bridge.** `packages/gsd-agent-core/src/sdk.ts` (`createAgentSession`):

   ```ts
   onPayload: async (payload, _model) => {           // model received…
     const runner = extensionRunnerRef.current;
     if (!runner?.hasHandlers("before_provider_request")) return payload;
     return runner.emitBeforeProviderRequest(payload); // …but never forwarded
   },
   ```

   `packages/pi-coding-agent/src/core/extensions/runner.ts` (`emitBeforeProviderRequest`) then builds the event without `model`:

   ```ts
   const event: BeforeProviderRequestEvent = {
     type: "before_provider_request",
     payload: currentPayload,
     // model is missing — but the type declares model?: { provider, id, api? }
   };
   ```

   `BeforeProviderRequestEvent` (extension-upstream-types.ts) has declared `model?: { provider: string; id: string; api?: string }` since the repo's initial import — the intent was always to carry it. `search-the-web/native-search.ts` documents `event.model` as its primary, "most reliable" detection path and only falls back to the `model_select` flag and the payload-name heuristic.

2. **`model_select` never fires to extensions in interactive mode.** The decision log shows `modelSelectFired=false` across every request, so the fallback flag is never armed either. This is a secondary latent issue — once `event.model` is present (fix #1), native-search's primary path works regardless — but it's worth tracking separately because other extensions may also rely on `model_select`.

Because `event.model` is absent **and** `model_select` never fires, native-search can only fall back to the model-name heuristic, which `#648` deliberately refuses (a Claude-named model served through copilot/openrouter must not get the Anthropic server tool). Net result: native search is unreachable on the one provider where it should always work.

## Regression / history

This is a **re-regression of an already-fixed bug**. The pre-import history lives in `gsd-2` (https://github.com/gsd-build/gsd-2), and the trail is exact:

1. `gsd-2` vendored Pi (`7d1e2cd27 feat: vendor Pi source into workspace monorepo`) with the buggy one-arg bridge — `onPayload: async (payload, _model)` → `emitBeforeProviderRequest(payload)`.
2. `gsd-2` **#444 / #446** (`654a9f5f1`, "fix: prevent web_search tool injection for non-Anthropic providers serving Claude models", 2026-03-14) **fixed it** — it threaded the model: `onPayload: async (payload, currentModel)` → `emitBeforeProviderRequest(payload, currentModel)`, added `model?` to the runner and to `BeforeProviderRequestEvent`. This is the *same* fix proposed here.
3. `gsd-pi` then went through the **ADR-010 "clean seam" restructure** (see `docs/dev/pi-upstream.md`): vendored Pi is re-pulled from upstream `earendil-works/pi` as *"upstream + allowlisted deltas"*, and GSD-specific session/SDK code is **extracted out of `pi-coding-agent` into the new `@gsd/agent-core` package**. This split #444's two halves apart.
4. The squashed initial-import commit (`#189`, 4146 files) carried the regressed state forward, so the one-arg form is all `gsd-pi` history shows.

### How the fix was lost: a cross-cutting patch bisected by the seam

gsd-2 #444 was a **single fix spanning two layers**: the vendored-Pi *plumbing* that supplies `event.model` (`pi-coding-agent` sdk.ts bridge + runner + event type) and the GSD-owned *consumer* that reads it (`search-the-web/native-search.ts`). The ADR-010 restructure preserved one half and reverted the other:

- **Plumbing half — reverted.** #444 was a *gsd-2-local delta*, not an upstream `earendil-works/pi` change. Re-vendoring "upstream + allowlisted deltas" reset `pi-coding-agent/runner.ts` and the event type to upstream's pre-#444 state, and the bridge was relocated from `pi-coding-agent/src/core/sdk.ts` (now **gone** in gsd-pi) into a freshly written `gsd-agent-core/src/sdk.ts` that came back in the pre-#444 shape (`_model` dropped). The #444 plumbing delta wasn't in the allowlist.
- **Consumer half — kept.** `native-search.ts` was migrated across intact; it still reads `event.model` as its "most reliable" path and still cites `#444` in its comments.

So the extension keeps asking for `event.model`; the rewired plumbing stopped supplying it.

### Why it stayed invisible

- **No contract test** asserted "the `before_provider_request` event carries the model" (this issue's fix adds one).
- **It fails *safe*, not loud.** The `#648` guard's job is to decline injection when it can't confirm the provider — so a missing model is indistinguishable from "correctly declining." No error, no 400, just a quiet fallback to `fetch_page`.
- **The comments imply otherwise.** `native-search.ts` still references `#444`, so a reader assumes it's in effect — but the wiring it depends on was cut two packages away.

This issue/PR re-applies #444's plumbing half to `gsd-pi`'s `gsd-agent-core` bridge and adds the missing contract test.

### Worth a broader audit (note for maintainers)

It's worth flagging that a shipped, tested fix (gsd-2 #444) silently failed to carry across the ADR-010 restructure. The failure mode — a **cross-cutting gsd-2-local delta** (vendored-Pi plumbing + a GSD-side consumer) where the re-vendor reset the plumbing half but kept the consumer — is not specific to web search. **Any other gsd-2 delta that spanned the `pi-coding-agent` → `@gsd/agent-core` (or `@gsd/agent-modes`) seam could have been bisected the same way and be silently degraded today.** A sweep of gsd-2's allowlisted/local deltas against the current vendored tree — especially anything touching the extension event surface, the SDK bridge, or provider gating — would be worth doing. The tell is exactly what we see here: gsd-pi code/comments still cite a gsd-2 PR number (e.g. `#444`) while the wiring that PR added is gone, and a unit test that mocks the missing input stays green.

## Fix

Thread the resolved model through the bridge and attach it to the event (matching the existing type):

- `gsd-agent-core/src/sdk.ts`: `return runner.emitBeforeProviderRequest(payload, model);`
- `pi-coding-agent/.../runner.ts`: accept `model?` and add `model` to the `BeforeProviderRequestEvent`.

Verified at runtime after the fix on `claude-opus-4-8` / `anthropic-messages`:

```
bpr: hasEventModel=true -> isAnthropic=true
req: onPayload=true  tools: 141 -> 142  hasWebSearch=true
```

…and the agent now calls `web_search` instead of falling back to `fetch_page`.

## GSD version

1.2.0

## Node.js version

v24.11.0

## Operating system

macOS (Apple Silicon)

## Install method

## Affected area

pi-coding-agent / agent-core (extension event plumbing); surfaced via the search-the-web native-search hook

## Model and provider

`claude-opus-4-8`, provider `anthropic`, api `anthropic-messages`

## Impact

Native Anthropic web search — a headline keyless capability — is completely non-functional for every Anthropic-provider session. Users get no web search at all (unless they configure and pay for Brave/Tavily and switch `search_provider`), and the agent confidently states it has no web search tool. Any other extension that gates on `event.model` is equally affected.

aidanthewiz:
## Linked issue

Closes #791

- [x] I have linked an issue above.

## TL;DR

**What:** Pass the resolved model into the `before_provider_request` extension event so it carries `event.model`.
**Why:** Native Anthropic web search never injected — the `onPayload` bridge dropped the model, so native-search's primary `event.model` detection always failed and the agent fell back to `fetch_page`.
**How:** Thread `model` through `gsd-agent-core` `onPayload` → `runner.emitBeforeProviderRequest(payload, model)` and attach it to the event (the event type already declares `model?`).

This is a **plumbing / agent-correctness** fix. No tool behavior changes — it only restores `event.model` on an extension event that already declared the field.

- Threads the resolved model from the `onPayload` bridge into the `before_provider_request` extension event, so `event.model` is populated.
- Re-enables native Anthropic `web_search` injection, which had been silently dead on every Anthropic-provider session (the agent fell back to `fetch_page` and reported having no web search).
- Adds a runner-level contract test asserting the event carries the model — the guard that was missing and let this regress unnoticed.
- No change to any extension that ignores `event.model`; no new public types.

## What changed — every change and why

| File | What changed | Why |
| --- | --- | --- |
| `packages/gsd-agent-core/src/sdk.ts` | The `onPayload` bridge in `createAgentSession` forwards the resolved model: `runner.emitBeforeProviderRequest(payload, model)` (was `(payload)`, model received as `_model` and discarded). | This is where the model was dropped. It is already available from the `pi-ai` `onPayload(params, model)` callback. |
| `packages/pi-coding-agent/src/core/extensions/runner.ts` | `emitBeforeProviderRequest` accepts an optional `model?: { provider; id; api? }` and attaches `model` to the `BeforeProviderRequestEvent`. | The event type has always declared `model?`; the runner just never populated it. |
| `packages/pi-coding-agent/test/extensions-runner.test.ts` | New regression test: registers a `before_provider_request` handler and asserts the runner forwards the resolved model into the event. | The missing contract test. The existing `native-search.test.ts` mocks `event.model` directly, so it stayed green while the runtime was broken — this guards the *producer* side. Verified red without the fix, green with it. |

## Why (the root problem)

On an Anthropic provider (`claude-opus-4-8` / `anthropic-messages`), native web search (`web_search_20250305`) was never injected. The search-the-web `native-search` hook reads `event.model` as its primary, "most reliable" provider-detection path; the event arrived with `model === undefined`, `model_select` never fired to arm the fallback flag, and the payload-name heuristic is correctly refused by the `#648` guard — so it hit the fail-safe and never injected. The agent had no `web_search` tool, fell back to `fetch_page`, and reported having no web search.

This was a latent wiring gap: `BeforeProviderRequestEvent` declared `model?` from the initial import, but the bridge never populated it.

**This re-applies a fix that already shipped upstream.** In `gsd-2` (https://github.com/gsd-build/gsd-2), PR **#444 / #446** (`654a9f5f1`, "fix: prevent web_search tool injection for non-Anthropic providers serving Claude models") added exactly this model threading. #444 was a cross-cutting fix: the vendored-Pi *plumbing* that supplies `event.model` **and** the GSD `native-search.ts` *consumer* that reads it. The ADR-010 "clean seam" restructure (re-vendor upstream `earendil-works/pi` + extract the SDK into `@gsd/agent-core`; see `docs/dev/pi-upstream.md`) bisected it: the consumer half was migrated intact (it still cites `#444`), but the plumbing half reverted to pre-#444 upstream and the relocated `gsd-agent-core/src/sdk.ts` bridge came back in the one-arg shape. With no test on the "event carries the model" contract and a fail-safe (`#648`) that hides the breakage, native web search went dark unnoticed. This change brings `gsd-pi` back in line with gsd-2 #444 and adds that missing contract test.

## How

The model is available in the `pi-ai` provider `onPayload(params, model)` callback and is passed up to `gsd-agent-core`'s `onPayload(payload, _model)`. We simply forward it:

```ts
// gsd-agent-core/src/sdk.ts
onPayload: async (payload, model) => {
  const runner = extensionRunnerRef.current;
  if (!runner?.hasHandlers("before_provider_request")) return payload;
  return runner.emitBeforeProviderRequest(payload, model);
},
```

```ts
// pi-coding-agent/.../runner.ts
async emitBeforeProviderRequest(
  payload: unknown,
  model?: { provider: string; id: string; api?: string },
): Promise<unknown> {
  ...
  const event: BeforeProviderRequestEvent = {
    type: "before_provider_request",
    payload: currentPayload,
    model,
  };
  ...
}
```

No new types, no behavior change for extensions that ignore `event.model`. Native-search's existing, documented `event.model` path now works as designed — no change needed in the search extension.

### Verification (runtime, before vs after)

Instrumented the Anthropic provider to log final `params.tools` after `onPayload`, and native-search's own decision:

| | `event.model` | `isAnthropic` | request tools | `web_search` present |
|---|---|---|---|---|
| before | `undefined` | `false` | 141 → 141 | ❌ |
| after  | `{anthropic, anthropic-messages}` | `true` | 141 → **142** | ✅ |

After the fix the agent calls `web_search` instead of `fetch_page`. (Instrumentation was temporary and is not part of this PR.)

### Notes for reviewers

- **`model_select` doesn't fire to extensions in interactive mode** (`modelSelectFired=false` in all observed requests), so native-search's fallback flag is never armed. This fix makes that moot for native search (the `event.model` path is the intended primary), but the non-delivery is worth a separate look since other extensions may rely on it.
- **This is a re-regression of a fix that already shipped (gsd-2 #444), and it may not be the only one.** The fix was lost because it was a *cross-cutting gsd-2-local delta* — vendored-Pi plumbing plus a GSD-side consumer — and the ADR-010 re-vendor reset the plumbing half while keeping the consumer. The same pattern could have bisected other gsd-2 deltas spanning the `pi-coding-agent` → `@gsd/agent-core`/`@gsd/agent-modes` seam, leaving them silently degraded. A sweep of gsd-2's allowlisted/local deltas against the current vendored tree may be worth doing; the tell is code that still cites a gsd-2 PR number while the wiring it added is gone (and a unit test that mocks the missing input stays green).

## Change type

- [x] `fix` — Bug fix

## Scope

- [x] `pi-agent-core` — Agent orchestration 
- [x] `pi-coding-agent` — Coding agent 

## Breaking changes

- [x] No breaking changes — additive optional parameter and an optional event field already present on the type.

## Test plan

- [x] New/updated tests included — `extensions-runner.test.ts` "emitBeforeProviderRequest forwards the resolved model into the event" (verified **red** with the fix reverted, **green** with it).
- [x] No regression — `native-search.test.ts` (the consumer suite) stays green (46/46); `tsc` clean on the changed files.
- [x] Manual testing — runtime-verified on `claude-opus-4-8` / `anthropic-messages` (before/after table above); `web_search` injected (`tools 141 → 142`) and used by the agent.
- [x] `pnpm run verify:merge` — ran against `e253f25`. Exits 1 **only** due to pre-existing, environment-driven unit failures (local `GSD_HOME` override, home-dependent provider-auth/path tests); the failure set is byte-identical to the `main` baseline (**10910 passed / 24 failed / 10 skipped** on both → **0 PR-attributable regressions**). PR-surface evidence is green: `extensions-runner` 29/29 incl. the new contract test, `gsd-agent-core` typecheck 0 errors, `pi-coding-agent` build exit 0.

## AI disclosure

- [x] This PR includes AI-assisted code. Root-caused and implemented with an AI coding agent (GSD). Verified by runtime request-tool capture on a live Anthropic session and a red/green regression test in `pi-coding-agent`.

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-pi/pr/792"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784576205&installation_id=134682257&pr_number=792&repository=open-gsd%2Fgsd-pi&return_to=https%3A%2F%2Fgithub.com%2Fopen-gsd%2Fgsd-pi%2Fpull%2F792&signature=35586a84c011275a9415052a99f23db5aa37e61402480221eb2b5020ef8f03cf"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/packages/gsd-agent-core/src/sdk.ts
+++ b/packages/gsd-agent-core/src/sdk.ts
@@ -372,12 +372,16 @@ export async function createAgentSession(options: CreateAgentSessionOptions = {}
 				headers,
 			});
 		},
-		onPayload: async (payload, _model) => {
+		onPayload: async (payload, model) => {
 			const runner = extensionRunnerRef.current;
 			if (!runner?.hasHandlers("before_provider_request")) {
 				return payload;
 			}
-			return runner.emitBeforeProviderRequest(payload);
+			// Thread the resolved model into the event so extensions can gate on
+			// provider/api (e.g. native web_search injection). Without this the
+			// before_provider_request event carries no model and provider-shape
+			// detection silently fails.
+			return runner.emitBeforeProviderRequest(payload, model);
 		},
 		onResponse: async (response, _model) => {
 			const runner = extensionRunnerRef.current;
--- a/packages/pi-coding-agent/src/core/extensions/runner.ts
+++ b/packages/pi-coding-agent/src/core/extensions/runner.ts
@@ -938,7 +938,10 @@ export class ExtensionRunner {
 		return currentMessages;
 	}
 
-	async emitBeforeProviderRequest(payload: unknown): Promise<unknown> {
+	async emitBeforeProviderRequest(
+		payload: unknown,
+		model?: { provider: string; id: string; api?: string },
+	): Promise<unknown> {
 		const ctx = this.createContext();
 		let currentPayload = payload;
 
@@ -951,6 +954,7 @@ export class ExtensionRunner {
 					const event: BeforeProviderRequestEvent = {
 						type: "before_provider_request",
 						payload: currentPayload,
+						model,
 					};
 					const handlerResult = await handler(event, ctx);
 					if (handlerResult !== undefined) {
--- a/packages/pi-coding-agent/test/extensions-runner.test.ts
+++ b/packages/pi-coding-agent/test/extensions-runner.test.ts
@@ -847,4 +847,33 @@ describe("ExtensionRunner", () => {
 			expect(runner.hasHandlers("agent_end")).toBe(false);
 		});
 	});
+
+	// Regression: the before_provider_request event must carry the resolved model.
+	// Extensions (e.g. native Anthropic web_search injection) gate on
+	// event.model.provider/api; previously the runner dropped the model, so the
+	// event arrived with model=undefined and provider-shape detection silently
+	// failed. See gsd-agent-core sdk.ts onPayload bridge.
+	it("emitBeforeProviderRequest forwards the resolved model into the event", async () => {
+		const extCode = `
+			export default function(pi) {
+				pi.on("before_provider_request", (event) => {
+					globalThis.__bprEventModel = event.model
+						? { provider: event.model.provider, api: event.model.api, id: event.model.id }
+						: null;
+					return event.payload;
+				});
+			}
+		`;
+		fs.writeFileSync(path.join(extensionsDir, "bpr-model.ts"), extCode);
+
+		const result = await discoverAndLoadExtensions([], tempDir, tempDir);
+		const runner = new ExtensionRunner(result.extensions, result.runtime, tempDir, sessionManager, modelRegistry);
+
+		const model = { provider: "anthropic", id: "claude-opus-4-8", api: "anthropic-messages" };
+		await runner.emitBeforeProviderRequest({ model: "claude-opus-4-8", tools: [] }, model);
+
+		const captured = (globalThis as unknown as { __bprEventModel?: unknown }).__bprEventModel;
+		expect(captured).toEqual(model);
+		delete (globalThis as unknown as { __bprEventModel?: unknown }).__bprEventModel;
+	});
 });


aidanthewiz:
## ✅ Risk Verification — PASS WITH NOTES

**Commit verified:** `e253f25dbc251c7ab86436e2e547f3ae35409a46` · **Date:** 2026-06-20 · **Verified by:** agent (pi)

| Extensions | verify extension event dispatch and hook ordering | ✅ |
| Event System | confirm event emission ordering and listener cleanup | ✅ |

### Evidence
| Check | Command | Result |
|-------|---------|--------|
| Extensions runner suite | `pnpm exec vitest run packages/pi-coding-agent/test/extensions-runner.test.ts` | 29/29 pass |
| New regression test | `vitest … -t "forwards the resolved model"` | 1 pass / 28 skipped |
| gsd-agent-core typecheck | `tsc --noEmit -p packages/gsd-agent-core/tsconfig.json` | 0 errors |
| pi-coding-agent build/typecheck | `tsc -p packages/pi-coding-agent/tsconfig.json` | exit 0 |

### Root cause & scope
- **Root cause:** Symptom — extensions gating on `before_provider_request` `event.model` (e.g. native web_search injection) silently no-op'd. True cause — the event's `model?` field (pre-existing, added in `cab50eeb`) was never populated: `ExtensionRunner.emitBeforeProviderRequest` omitted it and the `gsd-agent-core` SDK bridge discarded the resolved model (`_model`). This PR threads it end-to-end. No regressing commit; incomplete wiring since the field was introduced.
- **Scope:** Single production call site (the SDK bridge, updated). `model` param is optional → other caller (`test/suite/harness.ts`) stays backward-compatible. Event type pre-existing/untouched. No duplicated logic.
- **Downstream/stale deps:** The same-named harness `BeforeProviderRequestEvent` in `pi-agent-core` is a distinct type — unaffected. No stale consumers.

### Event dispatch & hook ordering confirmed
Runner loop iterates extensions in order → each extension's handlers in registration order → accumulates `currentPayload` forward; the only delta is the additive `model` field on the constructed event. Per-handler try/catch error isolation and listener registration/teardown (`hasHandlers`, provider register/unregister) are unchanged and covered by passing tests.

### Notes / caveats
- Build-order sensitivity (non-blocking): `gsd-agent-core` typechecks against `pi-coding-agent`'s built `dist/*.d.ts`. A stale dist yields `TS2554: Expected 1 arguments, but got 2` on the bridge; rebuilding `pi-coding-agent` clears it (0 errors). CI build order handles this — rebuild `pi-coding-agent` before typechecking `gsd-agent-core` locally.

PR #852, opened 2026-06-25
fix(issue): depth-verification gate for milestone CONTEXT.md is permanent block: user confirmation does not unlock it

### LINKED ISSUE #827: depth-verification gate for milestone CONTEXT.md is permanent block: user confirmation does not unlock it
## Problem
The depth-verification gate that blocks writes to `*-CONTEXT.md` files in `.gsd/milestones/<MID>/` triggers unconditionally and cannot be satisfied via the `ask_user_questions` confirmation flow. After the user explicitly confirms a write twice — selecting the "(Recommended)" option each time — the gate continues to block the same write with the same error message. There is no path through the gate from inside an agent session, leaving the agent unable to create CONTEXT.md files for newly-planned milestones. The gate's purpose is to require human depth verification before auto-generated milestone context is written, but the gate is currently a permanent blocker rather than a checkpoint.

## Root Cause

The gate fires on writes to `*.gsd/milestones/<MID>/<MID>-CONTEXT.md` paths. The blocking message is:

```
HARD BLOCK: Cannot write to milestone CONTEXT.md without depth verification.
This is a mechanical gate — you MUST NOT proceed, retry, or rationalize past this block.
Required action: call ask_user_questions with question id containing "depth_verification".
The user MUST select the "(Recommended)" confirmation option to unlock this gate.
If the user declines, cancels, or the tool fails, you must re-ask — not bypass.
```

The gate is enforced at the `write` tool boundary. The instruction requires the agent to call `ask_user_questions` with a question id containing "depth_verification" and have the user select the "(Recommended)" option. When the agent does this and the user selects the recommended option, the gate continues to block.

The bug is in the gate's confirmation-state tracking. Possible failure modes:

1. The gate checks the question's *options* array for a literal label string "(Recommended)" but the question-rendering layer strips that suffix before presenting to the user. The user sees a clean label like "Yes, write CONTEXT.md" and selects it, but the gate doesn't see the (Recommended) marker and rejects the confirmation.
2. The gate checks the user's selection against a server-side registry of "this user already approved" but the registry is keyed on the question id alone, and the question id from the agent's `ask_user_questions` doesn't match the gate's expected id pattern (the message says "containing 'depth_verification'" but the actual matcher may require a more specific format like `^depth_verification$` or `depth_verification_<step>`).
3. The gate is keyed on a transient session state that the `ask_user_questions` answer doesn't propagate to. The gate then re-checks the (non-existent) state on the next write attempt and finds nothing.

The third failure mode matches the observed behavior: the user has confirmed twice, but the gate's approval state is per-write-attempt and not propagated between attempts. Each write attempt re-runs the gate check from scratch.

Whatever the precise root cause, the practical symptom is: **a user-approved write to a `*-CONTEXT.md` file is permanently blocked, even after explicit confirmation**.

## Expected Behavior

A depth-verification gate should be a *checkpoint*, not a permanent block. After the user has explicitly confirmed via `ask_user_questions` with the recommended option, the gate should:

1. Recognize the confirmation state and allow the next write to a `*-CONTEXT.md` file to proceed.
2. The confirmation should be sticky for the duration of the agent session (or some sensible window), not per-attempt.
3. If the gate's matcher truly requires a specific question-id pattern, the gate's error message should state the exact pattern, not just "containing 'depth_verification'."
4. If the user declines, the gate should respect that and not block forever — it should allow the write to proceed with a recorded "user declined depth verification" annotation, or refuse the write without making it a permanent block.

The right durable fix is to use a clear, unambiguous question-id pattern that the gate matcher can recognize, and to have the confirmation state persist across multiple write attempts within a session. A simpler interim fix: when the gate is going to block, surface the exact required question-id and option-label format in the error message, so the agent can satisfy it on the first try.

## Reproduction

1. Plan a new milestone (any path through `gsd_plan_milestone` succeeds; the gate fires later on the write).
2. In the same session, attempt to write `<MILESTONE_DIR>/<MILESTONE_DIR_NAME>-CONTEXT.md` via the `write` tool.
3. Observe: the `write` tool returns the depth-verification gate block.
4. Call `ask_user_questions` with question id `depth_verification` and option label including `(Recommended)`.
5. User selects the recommended option.
6. Retry the same `write` call with the same content.
7. Observe: the gate still blocks, with the same error.
8. Repeat steps 4-6 with different question ids and option label phrasings. The gate continues to block on every attempt.

## Environment

- GSD version: 1.3.0
- OS: macOS
- Affected: any `write` tool call to `*.gsd/milestones/<MID>/<MID>-CONTEXT.md` after a fresh `gsd_plan_milestone` call

## Forensic Evidence

The user has confirmed the write twice in the same session:
- Confirmation 1: question id `depth_verification`, options included "Yes, write CONTEXT.md (Recommended)". User selected recommended.
- Confirmation 2: question id `depth_verification_confirm`, options included "Confirm: write it (Recommended)". User selected recommended.

Both `write` tool calls returned the same `HARD BLOCK: Cannot write to milestone CONTEXT.md without depth verification.` error.

The user's M009-kxo571 milestone (cluster offload for SLM selector, re-implementing the lost M002 work) is currently registered in the DB and has its `M009-kxo571-ROADMAP.md` file in place, but the `M009-kxo571-CONTEXT.md` write is permanently blocked despite the user explicitly approving it.

## Related

- This is the sixth GSD bug surfaced in the same forensic session. The first five were:
  - #821: doctor `stale_crash_lock` hint says "Run /gsd doctor" but the default invocation is read-only.
  - #822: `gsd worktree merge` crashes because the extension imports `packages/pi-ai/src/index.ts` which doesn't ship in the v1.3.0 tarball.
  - #823: `complete-milestone` blocks with "validation absent" when `validate-milestone` was skipped.
  - #825: `unmerged-milestone-guard` uses diff instead of ancestor check, producing false-positive blocks after a --no-ff merge.
  - #826: `unregistered-milestone` drift recovery hint points at `/gsd recover --confirm` but that command is a destructive full reimport, not a targeted import.
- The gate being a permanent blocker (rather than a checkpoint) is the same defect class as #825 (the diff-vs-ancestor check fails closed and blocks the user permanently) and #826 (the recovery hint points at a destructive operation without a targeted option). The common thread: gates that are supposed to protect the user end up trapping the user.

---

jeremymcs:
## Summary
- Fixed milestone depth-verification ID normalization and exact gate guidance; verified diff hygiene, with tests blocked by unavailable npm dependencies.

## Verification

## Related Issue
- Closes #827
- [#827 depth-verification gate for milestone CONTEXT.md is permanent block: user confirmation does not unlock it](https://github.com/open-gsd/gsd-pi/issues/827)

## User
- @AReid987

## Repo
- `open-gsd/gsd-pi`

## Branch
- `issue/827-depth-verification-gate-for-milestone-co-1782387768`

---

> [!NOTE]
> **Medium Risk**
> Changes the mechanical write gate and persisted verification state; scope is narrow (ID normalization) but incorrect behavior would still block or wrongly allow milestone CONTEXT writes.
> 
> **Overview**
> Fixes milestone **CONTEXT.md** staying blocked after the user confirms depth verification when milestone IDs differ only by casing (e.g. `depth_verification_m001_confirm` vs stored `M001`).
> 
> Adds **`normalizeMilestoneId`** and applies it across write-gate persistence and checks: loading/merging snapshots, marking depth verified (host and child adapters), extracting IDs from question IDs and CONTEXT paths, and **`isMilestoneDepthVerified`** lookups.
> 
> **`shouldBlockContextWrite`** block messages now name the exact gate id (`depth_verification_${milestone}_confirm`) and the first “(Recommended)” option. Tests cover lowercase question ids end-to-end via register-hooks and the updated error text.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 86bbc865b1b93a77c633a29cf322194d2e4593c5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-pi/pr/852"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a>

### DIFF
--- a/src/resources/extensions/gsd/bootstrap/write-gate.ts
+++ b/src/resources/extensions/gsd/bootstrap/write-gate.ts
@@ -25,6 +25,12 @@ export const MILESTONE_CONTEXT_RE = /M\d+(?:-[a-z0-9]{6})?-CONTEXT\.md$/;
 const CONTEXT_MILESTONE_RE = /(?:^|[/\\])(M\d+(?:-[a-z0-9]{6})?)-CONTEXT\.md$/i;
 const DEPTH_VERIFICATION_MILESTONE_RE = /depth_verification[_-](M\d+(?:-[a-z0-9]{6})?)/i;
 
+function normalizeMilestoneId(milestoneId: string): string {
+  const match = milestoneId.match(/^(M)(\d+)(?:-([a-z0-9]{6}))?$/i);
+  if (!match) return milestoneId;
+  return `M${match[2]}${match[3] ? `-${match[3].toLowerCase()}` : ""}`;
+}
+
 /**
  * Path segment that identifies .gsd/ planning artifacts.
  * Writes to these paths are allowed during queue mode.
@@ -228,7 +234,9 @@ function clearPersistedWriteGateSnapshot(basePath: string): void {
 function normalizeWriteGateSnapshot(value: unknown): WriteGateSnapshot {
   const record = value && typeof value === "object" ? value as Record<string, unknown> : {};
   const verified = Array.isArray(record.verifiedDepthMilestones)
-    ? record.verifiedDepthMilestones.filter((item): item is string => typeof item === "string")
+    ? record.verifiedDepthMilestones
+        .filter((item): item is string => typeof item === "string")
+        .map(normalizeMilestoneId)
     : [];
   const verifiedGates = Array.isArray(record.verifiedApprovalGates)
     ? record.verifiedApprovalGates.filter((item): item is string => typeof item === "string")
@@ -289,7 +297,7 @@ export function loadWriteGateSnapshot(basePath: string): WriteGateSnapshot {
  *     guard that previously protected only the tool_execution_start window.
  */
 function mergeSnapshotIntoState(state: InMemoryWriteGateState, disk: WriteGateSnapshot): void {
-  for (const milestone of disk.verifiedDepthMilestones) state.verifiedDepthMilestones.add(milestone);
+  for (const milestone of disk.verifiedDepthMilestones) state.verifiedDepthMilestones.add(normalizeMilestoneId(milestone));
   for (const gate of disk.verifiedApprovalGates ?? []) state.verifiedApprovalGates.add(gate);
   state.activeQueuePhase = disk.activeQueuePhase;
   state.pendingGateId = disk.pendingGateId;
@@ -395,7 +403,7 @@ export function isMilestoneDepthVerified(
 ): boolean {
   if (!milestoneId) return false;
   refreshWriteGateStateFromDisk(basePath);
-  return getWriteGateState(basePath).verifiedDepthMilestones.has(milestoneId);
+  return getWriteGateState(basePath).verifiedDepthMilestones.has(normalizeMilestoneId(milestoneId));
 }
 
 export function isMilestoneDepthVerifiedInSnapshot(
@@ -403,7 +411,7 @@ export function isMilestoneDepthVerifiedInSnapshot(
   milestoneId: string | null | undefined,
 ): boolean {
   if (!milestoneId) return false;
-  return snapshot.verifiedDepthMilestones.includes(milestoneId);
+  return snapshot.verifiedDepthMilestones.includes(normalizeMilestoneId(milestoneId));
 }
 
 export function isQueuePhaseActive(basePath: string = process.cwd()): boolean {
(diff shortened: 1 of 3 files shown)


jeremymcs:

Re-checked PR #852 against [#827 depth-verification gate for milestone CONTEXT.md is permanent block: user confirmation does not unlock it](https://github.com/open-gsd/gsd-pi/issues/827).

## Subtasks
- [x] **depth-verification gate for milestone CONTEXT.md is permanent block: user confirmation does not unlock it**

PR #938, opened 2026-06-27
fix(issue): [Bug]: Working-tree status is read twice per commit (`git status --porcelain` then `git status --short`)

### LINKED ISSUE #904: [Bug]: Working-tree status is read twice per commit (`git status --porcelain` then `git status --short`)
**Bug type**
Incorrect behavior or output

**Summary**
Each turn commit reads the working-tree status with `git status --porcelain` and then again with `git status --short`, doubling the most frequent git read on the hottest auto-mode path.

**Steps to reproduce**

1. Run `/gsd auto`.
2. Trace git invocations around each unit/task completion commit.
3. Observe `git status --porcelain` immediately followed by `git status --short` on the same tree in the same turn.

**Expected behavior**
One status read informs both the per-turn dirty report and the commit decision.

**Actual behavior**
`runTurnGitAction` calls `_resetHasChangesCache()` (`git-service.ts:1338`) then `collectRepositoryDirtyStatus` (`git-service.ts:1339`), which runs `runGit(repo.root, ["status", "--porcelain"])` for each repo (`git-service.ts:1270`). It then constructs `new GitServiceImpl(repo.root, gitPrefs).autoCommit(...)` (`git-service.ts:1311`); `autoCommit` (def `:925`) immediately calls `nativeHasChanges(this.basePath)` at its early-return guard (`git-service.ts:933`) — a second `git status --short` on the same tree. Because the cache was just reset at `:1338`, that second read genuinely re-spawns rather than hitting the ~10 s TTL cache. After staging, `nativeHasStagedChanges` (`git diff --cached --stat`, `:942`) runs too, but that is a legitimately distinct staged-vs-working check. So a single-repo commit issues `status --porcelain` + `status --short` (+ `diff --cached --stat`), where the first two are redundant.

**GSD version**
1.3.0 (local HEAD build at commit 451b7412, ~100+ commits ahead of the official 1.3.0 tag)

**Node.js version**
v26.3.1

**Operating system**
macOS 26.5.1 (Darwin 25.5.0, arm64)

**Install method**
local checkout / locally compiled build

**Affected area**
git/worktree isolation

**Logs, screenshots, and evidence**

- `src/resources/extensions/gsd/git-service.ts:1338-1339` — `_resetHasChangesCache()` then `collectRepositoryDirtyStatus(args.basePath)`.
- `src/resources/extensions/gsd/git-service.ts:1264-1270` — `collectRepositoryDirtyStatus` runs `runGit(repo.root, ["status", "--porcelain"])` (READ #1).
- `src/resources/extensions/gsd/git-service.ts:1311` — `new GitServiceImpl(...).autoCommit(...)`.
- `src/resources/extensions/gsd/git-service.ts:933` — `if (!nativeHasChanges(this.basePath)) return null;` → `git status --short` (READ #2, redundant; cache reset at `:1338` forces the re-spawn).
- `src/resources/extensions/gsd/git-service.ts:942` — `nativeHasStagedChanges` (`git diff --cached --stat`); `:962` — post-commit `_resetHasChangesCache()` (intentional, keep it).
- Amplifier: `src/resources/extensions/gsd/native-git-bridge.ts:19` — native libgit2 path is OFF by default (Issue #453, R1-C01), so `nativeHasChanges` is a CLI `git status --short` spawn.

**Impact**
LOW-MEDIUM severity, HIGH frequency (every unit/task commit). One redundant full `git status` per commit; on large trees `status` is the costliest read.

**Additional information**
Surgical fix: thread the already-computed dirty result from `collectRepositoryDirtyStatus` (`git-service.ts:1339`) into the commit path so `autoCommit` can skip its leading `nativeHasChanges` when the per-turn status is already known — implemented as an optional `knownDirty`/dirty-hint param on `autoCommit` (guard at `git-service.ts:933`) that defaults to today's behavior (re-check) so all other callers are unaffected. Keep the post-commit `_resetHasChangesCache()` (`:962`) so later checks still see a clean tree.

Do NOT flip `GSD_ENABLE_NATIVE_GSD_GIT` as the fix (broad blast radius).

---

### Reproduction evidence

- **Repro status:** partial — mechanism and the two distinct status spawns confirmed from source + the known fallback commands (derived from the constants `status --porcelain` / `status --short`). A full live `runTurnGitAction` turn was not driven (it requires a real GSD repo + preferences/registry + commit, beyond a cheap probe), so the count is trace-derived rather than measured live.
- **How tested:** read-only, non-destructive. Static call-graph trace of the commit (`action !== "status-only"`) path; confirmed the cache reset at `:1338` defeats the ~10 s `nativeHasChanges` TTL so the second read re-spawns. No project state, DB, git history, or worktree was mutated.
- **Raw result:**

  ```
  runTurnGitAction (commit path), single repo, one turn:
    line 1338  _resetHasChangesCache()         → invalidates the ~10s nativeHasChanges cache
    line 1339  collectRepositoryDirtyStatus    → runGit status --porcelain   [READ #1]
    line 1311  new GitServiceImpl(...).autoCommit(...)
    line 933   nativeHasChanges(this.basePath) → git status --short          [READ #2, redundant]
    line 942   nativeHasStagedChanges(...)      → git diff --cached --stat    (distinct: staged check)
    ⇒ status --porcelain + status --short are two full working-tree reads of the SAME tree
      in the same turn; the cache reset at 1338 guarantees READ #2 actually re-spawns.
  ```

- **Confidence & corrections:** The two reads are different git subcommands (`--porcelain` vs `--short`) producing equivalent dirty/clean signal — genuinely redundant for the dirty decision, though the second is not literally identical output. `nativeHasStagedChanges` is a legitimately distinct staged check and is NOT part of the redundancy. The `_resetHasChangesCache()` at `:1338` is intentional (force per-turn freshness for the dirty report); the bug is that `autoCommit` then re-reads instead of trusting that fresh result. Lowest severity of the area's findings, but highest frequency.

### Surgical-fix assessment

- **Proposed fix:** Optional dirty-hint param on `autoCommit` (guard at `git-service.ts:933`) fed by the result of `collectRepositoryDirtyStatus` (producer at `:1339`), defaulting to today's re-check behavior. Sites: `git-service.ts:1339` (producer) and `:933` (consumer).
- **Blast radius:** 1 file, 2 functions.
- **Touches a shared primitive?** Borderline — `autoCommit` is the central commit entrypoint, so the new param must default to current behavior (re-check) to avoid changing other callers.
- **Predicted bot verdict:** ready-for-agent if done as an additive optional hint param defaulting to today's behavior; would be needs-maintainer-review if it restructured `autoCommit`'s early-return contract.
- **Decomposition:** Not needed. Caveat to preserve: keep the post-commit `_resetHasChangesCache()` (`:962`). Deferred follow-up (do NOT bundle here): any change to `GSD_ENABLE_NATIVE_GSD_GIT` defaults.

jeremymcs:
## Summary
- Threaded turn dirty status into autoCommit and added a regression test; git diff --check passed, while runtime tests were blocked by missing deps/network.

## Verification

## Related Issue
- Closes #904
- [#904 [Bug]: Working-tree status is read twice per commit (`git status --porcelain` then `git status --short`)](https://github.com/open-gsd/gsd-pi/issues/904)

## User
- @astarktc

## Repo
- `open-gsd/gsd-pi`

## Branch
- `issue/904-bug-working-tree-status-is-read-twice-pe-1782525079`

---

> [!NOTE]
> **Low Risk**
> Small, localized git-service change with backward-compatible optional parameter and a targeted regression test; no auth or data-path changes.
> 
> **Overview**
> Turn **commit** actions already collect per-repo dirtiness once via `git status --porcelain` in `collectRepositoryDirtyStatus`. This change **threads that map into** `runPerRepositoryCommitAction` and **`autoCommit`’s new `knownDirty` argument**, so a repo marked dirty skips the extra `nativeHasChanges` probe (fallback path uses `git status --short`).
> 
> When `knownDirty` is **`false`**, `autoCommit` returns immediately without re-reading the tree; when **`undefined`**, behavior stays the same as before for callers that do not pass turn status.
> 
> A **regression test** wraps `git` and asserts a commit turn invokes `status --porcelain` only once and never `status --short`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 270474530b6cfe8bd4a37e16fa06ab45d1dea66c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-pi/pr/938"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a>

### DIFF
--- a/src/resources/extensions/gsd/git-service.ts
+++ b/src/resources/extensions/gsd/git-service.ts
@@ -921,16 +921,19 @@ export class GitServiceImpl {
    *
    * Returns the commit message on success, or null if nothing to commit.
    * @param extraExclusions Additional paths to exclude from staging (e.g. [".gsd/"] for pre-switch commits).
+   * @param knownDirty Fresh caller-provided dirty status; when omitted, autoCommit probes the tree.
    */
   autoCommit(
     unitType: string,
     unitId: string,
     extraExclusions: readonly string[] = [],
     taskContext?: TaskCommitContext,
+    knownDirty?: boolean,
   ): string | null {
     // Quick check: is there anything dirty at all?
     // Native path uses libgit2 (single syscall), fallback spawns git.
-    if (!nativeHasChanges(this.basePath)) return null;
+    if (knownDirty === false) return null;
+    if (knownDirty !== true && !nativeHasChanges(this.basePath)) return null;
 
     const scoped = taskContext
       ? this.scopedStageTaskFiles(taskContext, extraExclusions)
@@ -1282,6 +1285,7 @@ function runPerRepositoryCommitAction(args: {
   unitId: string;
   taskContext?: TaskCommitContext;
   targetRepositories?: string[];
+  dirtyRepositories?: Record<string, boolean>;
 }): {
   commitMessages: Record<string, string>;
   commitErrors: Record<string, string>;
@@ -1313,6 +1317,7 @@ function runPerRepositoryCommitAction(args: {
           args.unitId,
           [],
           args.taskContext,
+          args.dirtyRepositories?.[repo.id],
         ) ?? "";
       if (message) {
         commitMessages[repo.id] = message;
@@ -1360,7 +1365,7 @@ export function runTurnGitAction(args: {
       };
     }
 
-    const repoCommitResult = runPerRepositoryCommitAction(args);
+    const repoCommitResult = runPerRepositoryCommitAction({ ...args, dirtyRepositories });
     if (Object.keys(repoCommitResult.commitErrors).length > 0) {
       return {
         action: args.action,
--- a/src/resources/extensions/gsd/tests/uok-gitops-turn-action.test.ts
+++ b/src/resources/extensions/gsd/tests/uok-gitops-turn-action.test.ts
@@ -3,10 +3,20 @@
 
 import test from "node:test";
 import assert from "node:assert/strict";
-import { mkdtempSync, mkdirSync, realpathSync, rmSync, symlinkSync, writeFileSync } from "node:fs";
+import {
+  chmodSync,
+  mkdtempSync,
+  mkdirSync,
+  readFileSync,
+  realpathSync,
+  rmSync,
+  symlinkSync,
+  writeFileSync,
+} from "node:fs";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 import { execSync } from "node:child_process";
+import { GIT_NO_PROMPT_ENV } from "../git-constants.ts";
 import { handleTurnGitActionError, runTurnGitAction } from "../git-service.ts";
 
 function run(cmd: string, cwd: string): string {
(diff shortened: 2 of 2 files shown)


jeremymcs:

Re-checked PR #938 against [#904 [Bug]: Working-tree status is read twice per commit (`git status --porcelain` then `git status --short`)](https://github.com/open-gsd/gsd-pi/issues/904).

## Subtasks
- [x] **[Bug]: Working-tree status is read twice per commit (`git status --porcelain` then `git status --short`)**

PR #1079, opened 2026-06-30
fix(issue): [Bug]: memories_fts stays empty when FTS5 becomes available after schema V19 already ran without it

### LINKED ISSUE #1052: [Bug]: memories_fts stays empty when FTS5 becomes available after schema V19 already ran without it

**Summary:** When schema migration V19 runs on a SQLite build where FTS5 is unavailable, the `memories_fts` index is never populated, but the schema version still advances to 19 (and onward). If FTS5 later becomes available, a subsequent open creates an empty `memories_fts` virtual table via `CREATE VIRTUAL TABLE IF NOT EXISTS`, and only the insert/update/delete triggers keep it current going forward. Every memory written before FTS5 became available is never indexed. Keyword search then queries `memories_fts MATCH`, which returns zero rows for those historical memories without throwing, so the LIKE fallback (only reachable from the query's `catch`) never runs. Pre-FTS memories are silently dropped from search results with no error.

**Steps to reproduce:**

1. Open a GSD database on a SQLite build without FTS5 compiled in, so V19 records schema version 19 without backfilling `memories_fts` (`db/engine.ts` records each version unconditionally inside the `migrateSchema` guards).
2. Write memories while FTS5 is unavailable (they land only in the `memories` table).
3. Open the same database on a SQLite build with FTS5 available. `tryCreateMemoriesFts` (called from `initSchema`, `db/engine.ts:106-112`) creates an empty `memories_fts` external-content table (`content='memories'`); the version is already >= 19 so the V19 backfill is never reachable again (`migrateSchema` early-returns when `currentVersion >= SCHEMA_VERSION`).
4. Run a keyword search whose term appears in a pre-FTS memory.

**Expected behavior:** Keyword search returns all matching memories, including those written before FTS5 became available. When FTS5 first becomes available, the index is reconciled to cover the full `memories` table.

**Actual behavior:** Only memories written after FTS5 became available match. Pre-FTS memories are absent from results. The `MATCH` query returns zero rows for them without throwing, so the LIKE fallback is not triggered and there is no error or warning.

**GSD version:** 1.3.0 (local HEAD build at commit `658a36e5`, ~100+ commits ahead of the official 1.3.0 tag)

**Node.js version:** v26.3.1

**Operating system:** macOS 26.5.1 (Darwin 25.5.0, arm64)

**Install method:** local checkout / locally compiled build

**Affected area:** state/persistence/database

**Logs, screenshots, and evidence:** In-memory `node:sqlite` harness replicating the base-schema FTS create + triggers + keyword-search `MATCH` path:

```
fts table exists (isFtsAvailable=true): true
memories rows: 3  fts rows: 3
keyword 'auth' MATCH hits: ["m3"]
EXPECTED (all auth memories): m1, m3
RESULT: pre-FTS memory m1 silently MISSING from search => true
```

The harness also confirms `SELECT count(*) FROM memories_fts` returns 3 (the full `memories` row count) while only 1 row is actually indexed.

**Impact:** Memories written before FTS5 became available are permanently invisible to keyword search with no error and no self-heal, because the version-gated V19 backfill cannot run again.

**Additional information:** In `initSchema` after the base-schema objects are created (`db/engine.ts:106-112`), run `INSERT INTO memories_fts(memories_fts) VALUES('rebuild')` exactly once, driven by a one-shot `runtime_kv` marker keyed `memories_fts_rebuilt_at` (the `runtime_kv` table is created at `db/engine.ts:114` via `createRuntimeKvTableV25`): on open, if FTS5 is available (`isMemoriesFtsAvailable`) and the marker is absent, run the rebuild and set the marker. Do NOT gate this on a `count(*)` comparison: `memories_fts` is an external-content FTS5 table (`content='memories'`), so `count(*) FROM memories_fts` is answered from the content table and returns the full memory count even when the index is empty — a `count(*)` heuristic is provably dead here and would compile to a no-op guard.

### Reproduction evidence

- **Repro status:** Reproduced. Pre-FTS memory `m1` is silently absent from keyword results; only the post-FTS `m3` matches.
- **How tested:** In-memory `node:sqlite` harness replicating the base-schema FTS create, the sync triggers, and the `keywordSearch` `MATCH` path; seed memories with no FTS table, then create the FTS table and add a new memory, then query `MATCH`. Source lines re-confirmed at HEAD `658a36e5`.
- **Raw result:** `MATCH 'auth'` returned `["m3"]` only; `count(*) FROM memories_fts` returned 3 while only 1 row was indexed, confirming the external-content count behavior.
- **Confidence & corrections:** High. Confirmed `db/engine.ts:106-112` runs `tryCreateMemoriesFts` on every open and `createRuntimeKvTableV25` is invoked at `db/engine.ts:114`. No line drift from the reproduction report for this ticket.

jeremymcs:
## Summary
- Added one-shot FTS rebuild on schema open with runtime_kv marker and verified the focused FTS schema regression test.

## Verification

## Related Issue
- Closes #1052
- [#1052 [Bug]: memories_fts stays empty when FTS5 becomes available after schema V19 already ran without it](https://github.com/open-gsd/gsd-pi/issues/1052)

## User
- @astarktc

## Repo
- `open-gsd/gsd-pi`

## Branch
- `issue/1052-bug-memories-fts-stays-empty-when-fts5-b-1782824219`

---

> [!NOTE]
> **Low Risk**
> Scoped to startup when FTS exists; failures roll back and retry on next open; marker prevents repeat rebuilds.
> 
> **Overview**
> Fixes **empty `memories_fts`** when FTS5 was unavailable during migration V19 (or the external-content index was never populated for rows that already existed).
> 
> Adds **`rebuildMemoriesFtsSchemaOnce`**, which runs after schema init/migrations on DB open: if `memories_fts` exists and there is no `runtime_kv` marker (`memories_fts_rebuilt_at`), it runs FTS5 **`rebuild`** in a transaction, records the marker, and logs failures without blocking startup.
> 
> **Tests** cover the helper (rebuild + skip-if-marked) and an integration path where reopening the DB makes pre-existing memories searchable via `queryMemoriesRanked`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 179a9785d49363b468f224a698718a58d43290b8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-pi/pr/1079"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a>

### DIFF
--- a/src/resources/extensions/gsd/db-memory-fts-schema.ts
+++ b/src/resources/extensions/gsd/db-memory-fts-schema.ts
@@ -2,9 +2,16 @@
 // File Purpose: Memory FTS5 SQLite schema helpers for the GSD database facade.
 
 import type { DbAdapter } from "./db-adapter.js";
+import { createRuntimeKvTableV25 } from "./db-runtime-kv-schema.js";
+
+export const MEMORIES_FTS_REBUILT_KEY = "memories_fts_rebuilt_at";
 
 export interface MemoryFtsSchemaOptions {
   onUnavailable?: (message: string) => void;
+}
+
+export interface MemoryFtsRebuildOptions {
+  onRebuildFailed?: (message: string) => void;
 }
 
 function formatFtsUnavailableError(err: unknown): string {
@@ -62,5 +69,43 @@ export function isMemoriesFtsAvailableSchema(db: DbAdapter): boolean {
     return !!row;
   } catch {
     return false;
+  }
+}
+
+export function rebuildMemoriesFtsSchemaOnce(
+  db: DbAdapter,
+  options: MemoryFtsRebuildOptions = {},
+): void {
+  if (!isMemoriesFtsAvailableSchema(db)) return;
+
+  createRuntimeKvTableV25(db);
+  const marker = db.prepare(
+    "SELECT 1 as present FROM runtime_kv WHERE scope = 'global' AND scope_id = '' AND key = :key",
+  ).get({ ":key": MEMORIES_FTS_REBUILT_KEY });
+  if (marker) return;
+
+  const now = new Date().toISOString();
+  try {
+    db.exec("BEGIN");
+    db.exec("INSERT INTO memories_fts(memories_fts) VALUES('rebuild')");
+    db.prepare(
+      `INSERT INTO runtime_kv (scope, scope_id, key, value_json, updated_at)
+       VALUES ('global', '', :key, :value_json, :updated_at)
+       ON CONFLICT (scope, scope_id, key) DO UPDATE SET
+         value_json = excluded.value_json,
+         updated_at = excluded.updated_at`,
+    ).run({
+      ":key": MEMORIES_FTS_REBUILT_KEY,
+      ":value_json": JSON.stringify(now),
+      ":updated_at": now,
+    });
+    db.exec("COMMIT");
+  } catch (err) {
+    try {
+      db.exec("ROLLBACK");
+    } catch {
+      // Best effort: leave startup alive and retry the rebuild on next open.
+    }
+    options.onRebuildFailed?.(`FTS5 rebuild failed: ${(err as Error).message}`);
   }
 }
--- a/src/resources/extensions/gsd/db/engine.ts
+++ b/src/resources/extensions/gsd/db/engine.ts
@@ -47,7 +47,11 @@ import {
   applyMigrationV28MemoryLastHitAt,
   applyMigrationV29RepositoryTargets,
 } from "../db-migration-steps.js";
-import { isMemoriesFtsAvailableSchema, tryCreateMemoriesFtsSchema } from "../db-memory-fts-schema.js";
+import {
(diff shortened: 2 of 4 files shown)


jeremymcs:

Re-checked PR #1079 against [#1052 [Bug]: memories_fts stays empty when FTS5 becomes available after schema V19 already ran without it](https://github.com/open-gsd/gsd-pi/issues/1052).

## Subtasks
- [x] **[Bug]: memories_fts stays empty when FTS5 becomes available after schema V19 already ran without it**

PR #1123, opened 2026-07-01
fix(gsd): exempt browser automation tools from arg-independent loop cap

### LINKED ISSUE #1120: [Bug]: tool-call loop guard blocks a valid browser UAT — `browser_*` tools hit the arg-independent per-tool cap (6) during a legitimate multi-step verification

**Summary:** In `checkToolCallLoop` (`tool-call-loop-guard.ts`), Guard 2's per-tool-name cap counts calls to the same tool regardless of arguments; Browser Automation Contract tools (`browser_*`) are in neither the exempt set nor the repeatable-cap set, so they get the generic `PER_TOOL_DEFAULT_CAP = 6` and a normal multi-step browser UAT (many distinct `browser_evaluate`/`browser_act`/`browser_find` calls, each with different args) is falsely blocked as an "infinite loop," even though Guard 1's identical-signature streak never tripped.

**Steps to reproduce:**

1. Run a browser-executable UAT that legitimately calls one browser verb more than 6 times within a single agent turn with *different* arguments (e.g. `browser_evaluate` to read a validation message, then to count staged edits, then to inspect a control card, etc.).
2. On the 7th call, Guard 2 blocks with: `Tool loop detected (repeated tool): browser_evaluate called <N> times this turn (cap 6). Blocking to prevent infinite loop.`
3. The UAT cannot complete a valid PASS; auto-mode is pushed into the recovery/replan path despite the calls making real progress with distinct args.

**Expected behavior:** Distinct-argument browser-automation calls during a UAT are not treated as a loop. If a browser tool is capped at all, it should be by *signature* (identical args), which Guard 1 already enforces (cap 4) — not by a generic arg-independent count.

**Actual behavior:** Guard 2 counts `browser_*` calls arg-independently against `PER_TOOL_DEFAULT_CAP = 6` and blocks the 7th call regardless of whether the arguments differ, aborting a valid UAT.

**GSD version:** 1.4.0 (local HEAD build at commit `430a3a6c`, ~2 commits ahead of the v1.4.0 tag)

**Node.js version:** v26.3.1

**Operating system:** macOS 26.5.1 (Darwin 25.5.0, arm64)

**Install method:** local checkout / locally compiled build

**Affected area:** MCP/tools/extensions

**Logs, screenshots, and evidence:**

- Live transcript: `browser_evaluate` was called 9 times in one browser UAT turn with distinct scripts and was blocked — `Tool loop detected (repeated tool): browser_evaluate called 9 times this turn (cap 6). Blocking to prevent infinite loop.`
- `tool-call-loop-guard.ts` — `PER_TOOL_DEFAULT_CAP = 6`, `PER_TOOL_REPEATABLE_CAP = 15`, `PER_TOOL_CAP_EXEMPT_TOOLS = new Set(["find", "glob", "grep", "ls", "read", "search_and_read"])`. Guard 2 computes `perToolCount` independent of args and blocks when `perToolCount > perToolCap`.
- `browser_*` tools are in neither `PER_TOOL_CAP_EXEMPT_TOOLS` nor `INHERENTLY_REPEATABLE_TOOL_SET` (`core-session-tools.ts:30-44`, which lists core tools like `gsd_exec`/`multi_edit`/`search_and_read`), so they fall through to `PER_TOOL_DEFAULT_CAP = 6`.
- Guard 1 (identical-signature streak) already bounds a genuinely stuck browser loop: `MAX_CONSECUTIVE_IDENTICAL_CALLS = 4`, keyed on `hashToolCall(toolName, args)` — so *identical-arg* repeats are still caught.
- The Browser Automation Contract family is already recognized as a class elsewhere: `register-hooks.ts:190` `isBrowserTool(toolName) => hasBrowserContractPrefix(canonicalToolName(toolName))` (`hasBrowserContractPrefix` = `startsWith("browser_")`, `shared/browser-contract.ts:50`; `canonicalToolName`, `engine-hook-contract.ts:67`).
- The mutation-epoch decay does not help here: `recordToolCallLoopMutation` only bumps the epoch on `edit`/`write`/`multi_edit`/`notebook_edit`, none of which fire during a browser UAT, so the per-tool count never decays.

**Impact:** A browser-executable UAT that is legitimately verifying UI behavior (the exact workflow `/gsd auto` dispatches at slice closeout) is blocked mid-pass and forced into recovery/replan, wasting the run and preventing a valid PASS. Any browser verb used more than 6 times with varied args in one turn triggers it.

**Additional information:** In `checkToolCallLoop` (`tool-call-loop-guard.ts`), exempt Browser Automation Contract tools from Guard 2's arg-independent per-tool cap, exactly as the read/navigation tools are already exempted (Guard 1 still catches identical-arg loops). Extend the existing early return:

```ts
// existing:
if (PER_TOOL_CAP_EXEMPT_TOOLS.has(toolName)) {
  return { block: false, count: consecutiveCount };
}
```

to also skip browser-contract tools:

```ts
if (PER_TOOL_CAP_EXEMPT_TOOLS.has(toolName) || hasBrowserContractPrefix(canonicalToolName(toolName))) {
  return { block: false, count: consecutiveCount };
}
```

adding `import { hasBrowserContractPrefix } from "../../shared/browser-contract.js";` and `import { canonicalToolName } from "../engine-hook-contract.js";`. Rationale mirrors the existing exempt-set comment ("Read-only navigation tools are normal context gathering; Guard 1 still catches true reread loops with identical arguments"): browser UAT is inherently many distinct-arg calls, and identical-arg browser loops remain bounded by Guard 1's streak (cap 4). Preserve the invariant that a non-exempt tool (e.g. an arbitrary workflow tool) called past `PER_TOOL_DEFAULT_CAP` with varied args still blocks, and that an identical-arg `browser_evaluate` streak still trips Guard 1.

### Reproduction evidence

- **Repro status:** reproduced (live).
- **How tested:** Observed directly in a live browser UAT transcript (`browser_evaluate` ×9, distinct scripts, blocked at cap 6). Confirmed against source: `browser_*` is absent from both `PER_TOOL_CAP_EXEMPT_TOOLS` and `INHERENTLY_REPEATABLE_TOOL_SET`, so Guard 2 applies `PER_TOOL_DEFAULT_CAP = 6` arg-independently; Guard 1's identical-signature streak (cap 4) is the only signature-specific bound and did not fire because the args varied.
- **Raw result:** derived from constants `PER_TOOL_DEFAULT_CAP = 6` and `MAX_CONSECUTIVE_IDENTICAL_CALLS = 4`: the 7th varied-arg `browser_evaluate` call in a turn blocks; the observed run blocked at the 7th+ call and reported 9.
- **Confidence & corrections:** High. All cited `file:line` confirmed at HEAD `430a3a6c`.

astarktc:
## TL;DR

**What:** Exempt Browser Automation Contract tools (`browser_*`) from the tool-call loop guard's arg-independent per-tool cap (Guard 2).
**Why:** A legitimate multi-step browser UAT makes many distinct-arg `browser_*` calls in one turn and was falsely blocked as an "infinite loop."
**How:** Extend the existing read/navigation exemption to also skip browser-contract tools; Guard 1's identical-signature streak still bounds a genuinely stuck loop.

## What

`checkToolCallLoop` in `src/resources/extensions/gsd/bootstrap/tool-call-loop-guard.ts` runs two independent guards:

1. **Guard 1** — identical-signature streak (`MAX_CONSECUTIVE_IDENTICAL_CALLS = 4`), keyed on `hashToolCall(toolName, args)`.
2. **Guard 2** — per-tool-name cap counted **independent of args** (`PER_TOOL_DEFAULT_CAP = 6`, or `PER_TOOL_REPEATABLE_CAP = 15` for tools in `INHERENTLY_REPEATABLE_TOOL_SET`).

`browser_*` tools were in neither `PER_TOOL_CAP_EXEMPT_TOOLS` nor `INHERENTLY_REPEATABLE_TOOL_SET`, so they got the generic cap of 6. This change adds browser-contract tools to the early exemption from Guard 2, reusing the existing `hasBrowserContractPrefix(canonicalToolName(...))` helpers — the same recognition `register-hooks.ts` and `uat-policy.ts` already use. A regression test is added to `tool-call-loop-guard.test.ts`.

## Why

Closes #1120.

A browser-backed UAT — the exact workflow `/gsd auto` dispatches at slice closeout — legitimately calls one browser verb more than six times in a single turn with **different** arguments (read a validation message, count staged edits, inspect a control card, …). Because the args vary, Guard 1's identical-signature streak never trips, but Guard 2's arg-independent count hits the cap of 6 and blocks the 7th call:

```
Tool loop detected (repeated tool): browser_evaluate called 9 times this turn (cap 6). Blocking to prevent infinite loop.
```

The UAT can't reach a valid PASS and auto-mode is pushed into recovery/replan despite the calls making real progress. Read-only navigation tools (`find`/`glob`/`grep`/`ls`/`read`/`search_and_read`) were already exempted from Guard 2 for exactly this reason; browser UAT is the same shape — inherently many distinct-arg calls.

## How

Extend the existing Guard 2 exemption early-return:

```ts
if (
  PER_TOOL_CAP_EXEMPT_TOOLS.has(toolName) ||
  hasBrowserContractPrefix(canonicalToolName(toolName))
) {
  return { block: false, count: consecutiveCount };
}
```

with two imports (`hasBrowserContractPrefix` from `../../shared/browser-contract.js`, `canonicalToolName` from `../engine-hook-contract.js`). The exemption runs **after** Guard 1, so identical-arg browser loops are still bounded by the signature streak (cap 4), and MCP-prefixed spellings (`mcp__agent-browser__browser_evaluate`) are handled because `canonicalToolName` strips the prefix before the `browser_` check.

Preserved invariants (asserted by tests): (1) a non-exempt, non-browser tool called past `PER_TOOL_DEFAULT_CAP` with varied args still blocks via Guard 2; (2) an identical-arg `browser_evaluate` streak still trips Guard 1.

**Alternatives considered:** adding `browser_*` to `INHERENTLY_REPEATABLE_TOOL_SET` (cap 15) — rejected because a browser UAT still isn't a "loop" at 16 distinct-arg calls, and it would conflate browser tools with mutating/execution tools. The exemption mirrors the read/nav precedent exactly and keeps blast radius to one function.

## Change type

- [x] `fix` — Bug fix

## Testing

Test-first: the new regression case fails before the change (7th distinct `browser_evaluate` blocked at cap 6) and passes after. It imports and exercises `checkToolCallLoop` (not a source-grep test). Run:

```
node --import ./src/resources/extensions/gsd/tests/resolve-ts.mjs --experimental-strip-types --test src/resources/extensions/gsd/tests/tool-call-loop-guard.test.ts
```

`pnpm run typecheck:extensions` is clean. `pnpm run verify:merge` is being run before this PR is marked Ready for review.

---

_AI-assisted: this change was prepared with AI assistance and reviewed and understood before submission._

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-pi/pr/1123"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1785533116&installation_id=134682257&pr_number=1123&repository=open-gsd%2Fgsd-pi&return_to=https%3A%2F%2Fgithub.com%2Fopen-gsd%2Fgsd-pi%2Fpull%2F1123&signature=6acb3caec397ee9cfa5c197a11f0f01fc4cef0f502613a25589c1c0169c8c7e6"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/src/resources/extensions/gsd/bootstrap/tool-call-loop-guard.ts
+++ b/src/resources/extensions/gsd/bootstrap/tool-call-loop-guard.ts
@@ -25,6 +25,8 @@
 
 import { createHash } from "node:crypto";
 import { INHERENTLY_REPEATABLE_TOOL_SET } from "./core-session-tools.js";
+import { hasBrowserContractPrefix } from "../../shared/browser-contract.js";
+import { canonicalToolName } from "../engine-hook-contract.js";
 
 const MAX_CONSECUTIVE_IDENTICAL_CALLS = 4;
 
@@ -130,8 +132,16 @@ export function checkToolCallLoop(
   perToolLastMutationEpoch.set(toolName, mutationEpoch);
 
   // Read-only navigation tools are normal context gathering; Guard 1 still
-  // catches true reread loops with identical arguments.
-  if (PER_TOOL_CAP_EXEMPT_TOOLS.has(toolName)) {
+  // catches true reread loops with identical arguments. Browser Automation
+  // Contract tools (browser_*) are the same shape: a browser-backed UAT makes
+  // many distinct-arg calls (read a message, count edits, inspect a card, …),
+  // so the arg-independent per-tool cap misfires on legitimate verification;
+  // Guard 1's identical-signature streak still catches a genuinely stuck
+  // browser loop.
+  if (
+    PER_TOOL_CAP_EXEMPT_TOOLS.has(toolName) ||
+    hasBrowserContractPrefix(canonicalToolName(toolName))
+  ) {
     return { block: false, count: consecutiveCount };
   }
 
--- a/src/resources/extensions/gsd/tests/tool-call-loop-guard.test.ts
+++ b/src/resources/extensions/gsd/tests/tool-call-loop-guard.test.ts
@@ -385,3 +385,66 @@ console.log('\n── Loop guard: block reasons tell model to respond in text �
 }
 
 // ═══════════════════════════════════════════════════════════════════════════
+// Distinct-arg browser-automation calls are a legitimate UAT, not a loop (#1120)
+// ═══════════════════════════════════════════════════════════════════════════
+
+console.log('\n── Loop guard: distinct-arg browser automation calls are not capped by per-tool count (#1120) ──');
+
+{
+  // A browser-backed UAT legitimately calls one browser verb many times with
+  // DIFFERENT arguments (read a validation message, count staged edits, inspect
+  // a control card, …). Guard 1's identical-signature streak never trips because
+  // the args vary; Guard 2's arg-independent per-tool cap must NOT block it.
+  for (const toolName of ['browser_evaluate', 'browser_act', 'browser_find']) {
+    resetToolCallLoopGuard();
+    for (let i = 1; i <= 12; i++) {
+      const result = checkToolCallLoop(toolName, { script: `step ${i}` });
+      assert.ok(result.block === false, `distinct ${toolName} call ${i} should be allowed`);
+    }
+  }
+
+  // MCP-prefixed browser tools canonicalize to `browser_*` and are exempt too.
+  resetToolCallLoopGuard();
+  for (let i = 1; i <= 12; i++) {
+    const result = checkToolCallLoop('mcp__agent-browser__browser_evaluate', { script: `step ${i}` });
+    assert.ok(result.block === false, `distinct MCP-prefixed browser_evaluate call ${i} should be allowed`);
+  }
+}
+
+// ═══════════════════════════════════════════════════════════════════════════
+// Identical-arg browser loops are STILL caught by Guard 1 (#1120)
+// ═══════════════════════════════════════════════════════════════════════════
+
+console.log('\n── Loop guard: identical-arg browser loops are still caught by Guard 1 (#1120) ──');
+
+{
+  resetToolCallLoopGuard();
+  for (let i = 1; i <= 4; i++) {
+    const result = checkToolCallLoop('browser_evaluate', { script: 'document.title' });
+    assert.ok(result.block === false, `identical browser_evaluate call ${i} should be allowed`);
+  }
+  const blocked = checkToolCallLoop('browser_evaluate', { script: 'document.title' });
+  assert.ok(blocked.block === true, '5th identical browser_evaluate call should be blocked by Guard 1');
+  assert.ok(blocked.reason?.includes('identical args'), 'identical browser loops should be caught by Guard 1, not Guard 2');
+}
+
+// ═══════════════════════════════════════════════════════════════════════════
+// The browser exemption must NOT widen to other non-exempt tools (#1120)
+// ═══════════════════════════════════════════════════════════════════════════
+
+console.log('\n── Loop guard: browser exemption does not widen to other tools (#1120) ──');
+
+{
+  // Negative invariant: an arbitrary non-exempt, non-browser tool called past
+  // the default cap with varied args must still be blocked by Guard 2.
+  resetToolCallLoopGuard();
+  for (let i = 1; i <= 6; i++) {
+    const result = checkToolCallLoop('some_workflow_tool', { arg: `v${i}` });
+    assert.ok(result.block === false, `non-exempt tool call ${i} should be allowed`);
+  }
+  const blocked = checkToolCallLoop('some_workflow_tool', { arg: 'v7' });
+  assert.ok(blocked.block === true, '7th varied-arg non-exempt call must still trip Guard 2');
+  assert.ok(blocked.reason?.includes('repeated tool'), 'non-exempt overflow should be caught by Guard 2');
+}
+
+// ═══════════════════════════════════════════════════════════════════════════


astarktc:
### Scope verification (per the risk report follow-up)

Ran the requested scope assessment before/after the change.

**Root cause (not symptom).** Guard 2 of the tool-call loop guard counts calls per tool name *independent of args* against `PER_TOOL_DEFAULT_CAP = 6`. Browser Automation Contract tools (`browser_*`) are in neither `PER_TOOL_CAP_EXEMPT_TOOLS` nor `INHERENTLY_REPEATABLE_TOOL_SET`, so they fall through to that generic cap. That fall-through is the defect — a browser UAT is inherently many *distinct-arg* calls, exactly what the existing read/navigation exemption already recognizes. Guard 1 (identical-signature streak, cap 4) is the correct bound for a genuinely stuck browser loop and is unchanged.

**Other call sites / duplicated logic.** None. The per-tool cap logic exists only in `tool-call-loop-guard.ts`, and `checkToolCallLoop` has a single caller — the `tool_call` hook at `register-hooks.ts:1316`. The other `browser_*` recognition site (`isBrowserTool` in `register-hooks.ts`) governs tool-surface dropping, not loop capping, so it does not share the bug.

**Downstream consumers.** The only consumer returns the block verbatim (`return { block: true, reason: ... }`) with no browser-specific handling, so nothing depends on `browser_*` being capped at 6. The caller already passes a canonicalized name (`canonicalToolName(event.toolName)`); the guard's added `canonicalToolName(toolName)` is idempotent, so existing non-browser paths are byte-for-byte unchanged.

**Changes beyond the immediate file/function.** None — the diff is +14/-2 in `checkToolCallLoop` (two imports + one extended condition) plus the test.

**GSD Workflow state transitions.** No workflow state transition is touched: the guard is a pre-execution `tool_call` gate that only decides block/allow; it does not read or mutate milestone/slice/task state, the write gate, or any DB/projection. Net effect: a browser-backed UAT that previously aborted at the 7th distinct-arg call now runs to a normal PASS/FAIL, and auto-mode is no longer forced into a spurious recovery/replan.

---

### Local verification

Per `CONTRIBUTING.md` ("run `pnpm run verify:merge` before requesting review"). Reporting honestly rather than claiming a clean sweep, because this local checkout has pre-existing failures unrelated to the change:

- **Regression test (test-first):** the added case in `tool-call-loop-guard.test.ts` fails before the fix (7th distinct-arg `browser_evaluate` blocked at cap 6) and passes after. Two negative invariants are also asserted — a non-exempt/non-browser tool past the default cap with varied args still blocks (Guard 2), and an identical-arg `browser_evaluate` streak still trips Guard 1.
- **`typecheck:extensions`:** clean. **`build:core` + web-host build:** pass. Loop-guard suites (`tool-call-loop-guard`, `register-hooks-loop-guard-auto`) pass.
- **`test:unit`:** `11490 passed, 53 failed`. The 53 failures are a **byte-identical set to clean `origin/main`** (verified by diffing the sorted failure lists) — pre-existing/environmental on this checkout (e.g. macOS `/var` vs `/private/var` realpath, plus some permission-rule/hook/worktree suites), **not caused by this change**; none touch the loop guard.
- **`validate-pack`:** fails only at the global-install hoisting-repair smoke (packs the tarball, global `npm install` into a temp prefix with a 15s timeout; no stdout/stderr emitted) — environment-sensitive and independent of a 2-line guard edit. All packaging checks that source can affect (isolated install, workspace resolution, standalone host, `gsd -v`, undici, agent-core re-exports) passed.

CI in a clean environment is the authoritative gate for those two environmental items. Happy to adjust anything on request.

PR #1147, opened 2026-07-02
fix(git): start workflow and quick-task branches from main, not HEAD

404prefrontalcortexnotfound:
## Problem

Workflow templates (`commands-workflow-templates.ts`, both the legacy and plugin paths) and quick tasks (`quick.ts`) create their working branch with a bare `git checkout -b <branch>`, which branches from whatever is currently checked out.

When a workflow finishes, GSD leaves its `gsd/<template>/<slug>` branch checked out. The next workflow therefore branches on top of it, and the one after that on top of both. In practice this silently produces a stacked chain: every new `gsd/*` branch — and any PR opened from it — drags along all commits from every previous workflow. We hit this in a real repo where a docs-only workflow PR ended up containing 50 files from three earlier workflows.

## Fix

New shared helper `taskBranchArgs()` in `git-service.ts`, used at all three call sites. Design constraints, in order:

1. **Redirection is gated to leftover GSD task branches.** The start-point override only fires when the current branch matches `WORKFLOW_BRANCH_RE` / `QUICK_BRANCH_RE` / `SLICE_BRANCH_RE` — i.e. a branch GSD itself created and abandoned. On a user's own branch (feature branch, worktree base, the main branch itself) the legacy branch-from-HEAD behavior is unchanged: basing new work on the branch the user chose to stand on is intentional there.
2. **The base comes from `GitServiceImpl.getMainBranch()`**, passed in by the callers, so the explicit `main_branch` preference, milestone integration branch, and worktree base branch all outrank repo-default detection. The helper never detects on its own.
3. **No new failure paths.** A base branch with no local ref (e.g. `origin/HEAD` naming a branch never checked out locally) and a tree still dirty after the pre-switch auto-commit both fall back to branch-from-HEAD with a logged warning — branch creation cannot fail anywhere it previously could not.
4. **Quick tasks stay coherent with their closeout.** A redirected quick branch records the base as its squash-merge return target, so main-rooted quick work is not merged back into the stale task branch it started on.

The pre-switch `autoCommit` snapshot is untouched — dirty state is still committed to the branch it was made on before switching away. The base is the local ref, deliberately without a fetch, so branch creation stays offline-safe.

## Tests

Added to `tests/integration/git-service.test.ts`:
- anti-stacking: a second workflow branch created after a finished one points at the base, with none of the first workflow's commits;
- manual feature branch: no redirect;
- base branch with no local ref: falls back to branch-from-HEAD;
- dirty working tree on a task branch: falls back to branch-from-HEAD.

```
node --import ./src/resources/extensions/gsd/tests/resolve-ts.mjs --experimental-strip-types --test src/resources/extensions/gsd/tests/integration/git-service.test.ts
```

75 pass / 2 fail — the 2 failures (autoCommit pre-commit-retry tests) fail identically on an unmodified checkout in my environment and are unrelated to this change.

### DIFF
--- a/src/resources/extensions/gsd/commands-workflow-templates.ts
+++ b/src/resources/extensions/gsd/commands-workflow-templates.ts
@@ -23,7 +23,7 @@ import {
 } from "./workflow-templates.js";
 import { loadPrompt } from "./prompt-loader.js";
 import { gsdRoot } from "./paths.js";
-import { createGitService, runGit } from "./git-service.js";
+import { createGitService, runGit, taskBranchArgs } from "./git-service.js";
 import { isAutoActive, isAutoPaused } from "./auto.js";
 import { getErrorMessage } from "./error-utils.js";
 import { resolvePlugin, type WorkflowPlugin } from "./workflow-plugins.js";
@@ -464,7 +464,7 @@ export async function handleStart(
         try {
           git.autoCommit("workflow-template", templateId, []);
         } catch { /* nothing to commit */ }
-        runGit(basePath, ["checkout", "-b", branchName]);
+        runGit(basePath, taskBranchArgs(basePath, branchName, current));
         branchCreated = true;
       }
     } catch (err) {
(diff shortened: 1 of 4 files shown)


jeremymcs:
@404prefrontalcortexnotfound please open an issue before your PRs. Ask your agent to use CONTRIB

404prefrontalcortexnotfound:
> @404prefrontalcortexnotfound please open an issue before your PRs. Ask your agent to use CONTRIB

Apologies, we'll do so going forward. 

Thanks for a fantastic tool by the way.

PR #1266, opened 2026-07-05
fix(issue): [Bug]: auto-start's DB gate only checks that the database opened, not that it is writable, so a read-only handle passes and fails minutes later with an opaque error

### LINKED ISSUE #1234: [Bug]: auto-start's DB gate only checks that the database opened, not that it is writable, so a read-only handle passes and fails minutes later with an opaque error
**Bug type**
Incorrect behavior or output

**Summary**
The auto-start DB gate (`auto-start.ts:1741-1780`) aborts only when the provider is *unavailable* (`existsSync(gsdDbPath) && !isDbAvailable()`); it performs no write. On a schema-current DB every migration/init step is a no-op, so a handle that opened successfully but is not writable (read-only file, WAL/SHM permission mismatch, or a stale/moved handle) passes the gate and only fails much later at the first real write.

**Steps to reproduce**

1. Make `gsd.db` open successfully but be non-writable at the OS/handle level (e.g. read-only perms, or a handle whose backing file was replaced → `SQLITE_READONLY_DBMOVED`).
2. Run `/gsd auto`.
3. The open-only gate passes; auto-start proceeds through its bootstrap and then dies at the first fatal-allowed write (the `uok-kernel-enter` audit) with an opaque "readonly database" error.

**Expected behavior**
Auto-start should detect a non-writable database at open time and abort with a clear, actionable "database is not writable" message, instead of proceeding and failing minutes later with an opaque audit-flavored error.

**Actual behavior**
The gate checks only that the provider loaded and the file opened. A write-degraded handle sails through and the failure surfaces later as:

```
Error: Auto-start failed: DB authoritative audit write failed: attempt to write a readonly database
```

**GSD version**
1.5.0 (local HEAD build at commit `48fc294`, ~2751 commits on HEAD)

**Node.js version**
v26.3.1

**Operating system**
macOS 26.5.1 (Darwin 25.5.0, arm64) — analysis host; the incident was observed on the reporter's WSL2 machine at the same commit.

**Install method**
local checkout / locally compiled build

**Affected area**
state/persistence/database

**Logs, screenshots, and evidence**

- The open-only gate, `src/resources/extensions/gsd/auto-start.ts:1741-1780`:

  ```ts
  const initialDbOpen = openWorkflowDatabase(s.basePath);
  // …retry…
  // Gate: abort bootstrap if the DB file exists but the provider is still unavailable…
  if (existsSync(gsdDbPath) && !isDbAvailable()) {
    // …builds phaseHint from getWorkflowDatabaseStatus()…
    ctx.ui.notify(`SQLite database exists but failed to open: ${gsdDbPath}. …`, "error");
    return releaseLockAndReturn();
  }
  ```

  `isDbAvailable()` reflects only that the provider loaded and `open` succeeded; nothing here issues a write, so a read-only/moved handle reports available.

**Impact**
A recoverable, common failure class (read-only DB, WAL/SHM perms, moved handle) is not caught at the natural checkpoint. Instead auto-start runs its whole bootstrap and then aborts with a message that names "audit write" rather than the real cause (the DB is not writable), making the failure hard to diagnose.

**Additional information**

Fix (smallest behavior-preserving change): after the DB opens and before clearing the gate in `auto-start.ts` (right after the `existsSync(gsdDbPath) && !isDbAvailable()` check at `:1758`), add a **page-write** probe and abort with a clear message on failure:

```ts
// Confirm the handle is actually writable, not just open. A schema-current
// DB does zero writes during open, so a read-only / DBMOVED handle otherwise
// passes silently and fails later at the first authoritative write.
if (existsSync(gsdDbPath) && isDbAvailable()) {
  const writable = probeDbWritable();   // must dirty a page, then roll back
  if (!writable.ok) {
    ctx.ui.notify(`SQLite database is not writable: ${gsdDbPath}. ${writable.detail}`, "error");
    return releaseLockAndReturn();
  }
}
```

The probe **must force a real page write** — e.g. `PRAGMA user_version = <current>` (idempotent: read it first, write the same value) or `BEGIN IMMEDIATE; <no-op update to an existing row>; ROLLBACK`. A bare `BEGIN IMMEDIATE; ROLLBACK` is not sufficient: a moved handle (`SQLITE_READONLY_DBMOVED`) does not fail until a page is actually dirtied, so it would false-pass the probe.

### Reproduction evidence

- **Repro status:** reproduced (static gate is code-proven; the degraded-handle behavior verified with a micro-repro).
- **How tested:** read-only trace at HEAD `48fc294` of the gate at `auto-start.ts:1741-1780` (no write performed); separately, a Node `node:sqlite` micro-harness replaced the DB file under an open handle and confirmed a bare `BEGIN IMMEDIATE; ROLLBACK` succeeds while the next page write fails with `attempt to write a readonly database` (errcode 1032 = `SQLITE_READONLY_DBMOVED`).
- **Raw result:** open-only gate passes a non-writable handle; a page-dirtying probe is required to catch it.
- **Confidence & corrections:** high; the probe-must-dirty-a-page requirement is the key correction over a naive transaction probe.

jeremymcs:
## Summary
- Added a page-writing `probeDbWritable()` and an auto-start gate that aborts with a clear "database is not writable" message when the DB opens read-only/moved; verified with new probe tests, the existing gate test, and a clean extensions typecheck.

## Verification

## Related Issue
- Closes #1234
- [#1234 [Bug]: auto-start's DB gate only checks that the database opened, not that it is writable, so a read-only handle passes and fails minutes later with an opaque error](https://github.com/open-gsd/gsd-pi/issues/1234)

## User
- @astarktc

## Repo
- `open-gsd/gsd-pi`

## Branch
- `issue/1234-bug-auto-start-s-db-gate-only-checks-tha-1783262666`

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-pi/pr/1266"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a>

---

> [!NOTE]
> **Low Risk**
> Bootstrap-only guard with a rolled-back probe write; no change to normal DB write paths beyond the new startup check.
> 
> **Overview**
> Auto-start used to treat a successful DB open as enough to continue, but a schema-current database can open without writing anything—so read-only files, WAL permission problems, or stale **DBMOVED** handles still passed and only failed later with an opaque readonly error.
> 
> This PR adds **`probeDbWritable()`**, which rewrites the current `PRAGMA user_version` inside `BEGIN IMMEDIATE` and rolls back, forcing a real page write that catches those cases. **`bootstrapAutoSession`** now runs that probe after the existing open gate and aborts with a clear “database is not writable” message (including SQLite detail) instead of starting auto-mode. Regression tests cover writable DB, no open handle, and simulated read-only write failure.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1129b8a162749dbe4c4579408c1ab0c31aeb0157. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/resources/extensions/gsd/auto-start.ts
+++ b/src/resources/extensions/gsd/auto-start.ts
@@ -76,7 +76,7 @@ import { initRoutingHistory } from "./routing-history.js";
 import { restoreHookState, resetHookState, reconcileRestoredHookDispatch } from "./post-unit-hooks.js";
 import { resetProactiveHealing, setLevelChangeCallback } from "./doctor-proactive.js";
 import { snapshotSkills } from "./skill-discovery.js";
-import { isDbAvailable, getMilestone, getAllMilestones, insertMilestone, updateMilestoneStatus } from "./gsd-db.js";
+import { isDbAvailable, probeDbWritable, getMilestone, getAllMilestones, insertMilestone, updateMilestoneStatus } from "./gsd-db.js";
 import {
   getWorkflowDatabaseStatus,
   openExistingWorkflowDatabase,
@@ -1788,6 +1788,23 @@ export async function bootstrapAutoSession(
         "error",
       );
       return releaseLockAndReturn();
+    }
+
+    // Gate: confirm the handle is actually writable, not just open. A
+    // schema-current DB does zero writes during open, so a read-only /
+    // DBMOVED handle otherwise passes the check above and only fails much
+    // later at the first authoritative write (the uok-kernel-enter audit)
+    // with an opaque "readonly database" error (#1234).
+    if (existsSync(gsdDbPath) && isDbAvailable()) {
+      const writable = probeDbWritable();
+      if (!writable.ok) {
+        const detail = writable.detail ? ` (${writable.detail})` : "";
+        ctx.ui.notify(
+          `SQLite database is not writable: ${gsdDbPath}.${detail} Check file/WAL permissions or reopen a stale handle before running auto-mode.`,
+          "error",
+        );
+        return releaseLockAndReturn();
+      }
     }
 
     // Initialize metrics
--- a/src/resources/extensions/gsd/db/engine.ts
+++ b/src/resources/extensions/gsd/db/engine.ts
@@ -860,3 +860,39 @@ export function getDb(): DbAdapter {
 export function getDbOrNull(): DbAdapter | null {
   return currentDb;
 }
+
+export interface DbWritableProbeResult {
+  ok: boolean;
+  detail?: string;
+}
+
+/**
+ * Confirm the open handle can actually write, not just that it opened.
+ *
+ * A schema-current database performs zero writes during open, so a handle that
+ * opened successfully but is not writable (read-only file, WAL/SHM permission
+ * mismatch, or a stale/moved handle → SQLITE_READONLY_DBMOVED) otherwise passes
+ * the open-only availability check and only fails much later at the first real
+ * write. The probe forces a genuine page write by re-writing the current
+ * `PRAGMA user_version` value back inside an IMMEDIATE transaction: a bare
+ * `BEGIN IMMEDIATE; ROLLBACK` is not sufficient because a moved handle does not
+ * fail until a page is actually dirtied. The value is unchanged, so the probe
+ * is idempotent, and the transaction is rolled back so nothing persists.
+ */
+export function probeDbWritable(): DbWritableProbeResult {
+  const db = currentDb;
+  if (!db) return { ok: false, detail: "No database is open." };
+  try {
+    const row = db.prepare("PRAGMA user_version").get() as { user_version?: number } | undefined;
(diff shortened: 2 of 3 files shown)


jeremymcs:

Re-checked PR #1266 against [#1234 [Bug]: auto-start's DB gate only checks that the database opened, not that it is writable, so a read-only handle passes and fails minutes later with an opaque error](https://github.com/open-gsd/gsd-pi/issues/1234).

## Subtasks
- [x] **[Bug]: auto-start's DB gate only checks that the database opened, not that it is writable, so a read-only handle passes **

PR #1340, opened 2026-07-07
fix(gsd): reject path-traversal keys in compat marker projections

jeremymcs:
## TL;DR

**What:** Reject path-traversal keys in the gsd-core compat marker (`.gsd/.compat.json`) during marker validation, before the reconcile drift detectors read them.
**Why:** The marker is repo-controlled content, but validation only checked entry *values* (`sha`, `entities`) and never the projection-map *keys* — relative paths that get `join`ed under `.gsd/`/`.planning/` and `readFileSync`'d — so a hostile key like `../../../.ssh/known_hosts` caused arbitrary out-of-project file reads on every reconcile pass.
**How:** Add `isSafeProjectionKey` and enforce it in `isValidProjectionMap`, the single chokepoint every marker and both projection maps route through; one escaping key invalidates the whole marker so `readCompatMarker` fails safe to the empty marker.

## What

- `src/resources/extensions/gsd/compat/compat-marker.ts` — new `isSafeProjectionKey(key)` predicate (rejects empty, backslash, NUL, absolute, drive-letter, and any `normalize`d form that escapes with `..`); `isValidProjectionMap` now validates keys in addition to values.
- `src/resources/extensions/gsd/tests/compat-marker.test.ts` — 7 new tests: four escaping-key cases (`../outside.md`, `/etc/hosts`, `C:/x.md`, `..\x.md`), a legitimate-nested-key round-trip, a `planning.passthrough` escape case, and a hostile-marker→drift-detector integration test with an out-of-project sentinel.
- `plans/README.md` — status row for plan 029; recorded the write-side sibling finding (raw entity-ID path joins in `workflow-projections.ts` / `migrate/writer.ts`, currently latent because ID generators are slug-safe) as deferred defense-in-depth.

No behavior change for legitimate markers: normal projection keys (roadmap/plan/phase paths) still round-trip unchanged.

## Why

`.gsd/.compat.json` lives inside the user's repository, so a malicious or compromised repo controls its contents entirely (the only authenticity check is `lastWriter === "gsd-pi"`, a plain JSON string). The `projections` map keys are relative file paths, and marker validation checked only the entry values — keys were never sanitized. The reconciliation drift detectors (`external-markdown-edit.ts`, `external-planning-edit.ts`) then do `join(basePath, ".gsd", projectionPath)` and `readFileSync` on every key.

A key such as `"../../../../Users/<user>/.ssh/known_hosts"` therefore made gsd-pi read arbitrary files outside the project on the next reconcile pass — which runs before every auto-mode dispatch. Combined with the repair path (which re-imports drifted markdown into the DB and re-projects it into repo-visible files), this is an arbitrary-file read + existence probe with a plausible disclosure route. The write layer already has the containment guard this mirrors (`db-writer.ts` throws when a projection path escapes `.gsd/`); the marker read side never got it.

## How

The fix is a single chokepoint at marker validation rather than per-detector guards (keeping it out of the way of the in-flight scope-external-edit-reimport change). Both drift detectors consume the marker via `readCompatMarker`, so they inherit the guard.

`isSafeProjectionKey` mirrors the write-side guard in `db-writer.ts` (`relative()` then reject `..`-prefixed or absolute). Enforcement lives in `isValidProjectionMap`, through which `isValidMarker` and `isValidPlanningMarker` route for `projections`, `passthrough`, and the planning marker.

Key decision — **whole-marker rejection vs dropping bad entries:** one escaping key marks the entire marker invalid. A marker containing traversal keys is hostile or corrupt, and `readCompatMarker` already falls back to the empty marker on invalid input (quarantining the file). The fail-safe direction is to trust none of it: detectors then see no entries, so nothing is read or repaired for that pass. If gsd-core interop ever needs graceful partial acceptance, drop entries instead — but keep the containment predicate.

Verified locally: `pnpm run typecheck:extensions` exits 0; the new test file passes 18/18; stashing the source fix flips the four traversal assertions to failing (proving the tests exercise the guard); and the drift/reconciliation suites (`external-markdown-edit`, `external-planning-edit`, `compat-marker-invalidation`, `planning-marker`, `planning-round-trip-property`, `planning-projection-hook`) all pass standalone.

## Change type

- [x] `fix` — Bug fix (security: path-traversal containment)
- [x] `test` — Adds tests for the new guard

No public API, CLI, or config changes. AI-assisted.

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-pi/pr/1340"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a>

---

> [!NOTE]
> **Medium Risk**
> Touches reconcile/drift reads (hostile marker content), SQLite claiming under concurrency, gate ledger atomicity, and derive-state caching in auto-mode—important correctness paths, with the compat-marker change being the main security-sensitive piece.
> 
> **Overview**
> This PR lands **Plan 028** (six small state/DB fixes) and **Plan 029** (compat-marker read-side path containment), with matching tests and plan index updates.
> 
> **Plan 029 — security:** `readCompatMarker` now rejects projection-map keys that could escape `.gsd/` / `.planning/` (`isSafeProjectionKey` in `isValidProjectionMap`). One bad key invalidates the whole marker so drift detectors fail safe to the empty marker instead of `join` + `readFileSync` outside the repo. Tests cover hostile keys, legitimate nested keys, `planning.passthrough`, and the external-markdown drift detector.
> 
> **Plan 028 — correctness:** `skipSliceCascade` uses `isClosedStatus` (excluding re-skip of `skipped`) so `closed` slices are not downgraded. `insertTask` stamps `completed_at` via `toStatus` for all complete aliases, not `done`/`complete` only. `saveGateResult` wraps the gate UPDATE and `insertGateRun` in one `transaction`. `claimNextCommand` uses `immediateTransaction` to avoid WAL `SQLITE_BUSY_SNAPSHOT` races. Auto-mode `captureMilestoneLockEnv` / `restoreMilestoneLockEnv` call `invalidateStateCache` when `GSD_MILESTONE_LOCK` flips. New tests cover derive-cache vs lock, gate atomicity rollback, lease force-release, and cascade behavior.
> 
> `plans/README.md` marks plans 028 and 029 DONE and notes deferred write-side ID path hardening.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cbe66b3564aed6a4eef9fd066483ae7f62472db3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


jeremymcs:

Merge conflicts were detected while updating this branch from the base branch.

Conflicted files:
- `plans/README.md`

An automatic resolution attempt is running. If it cannot complete safely, this PR will be marked for manual follow-up.

PR #1342, opened 2026-07-08
fix(gsd): serialize write-gate snapshot read-merge-write across processes

jeremymcs:
## TL;DR

**What:** Serialize the write-gate snapshot's cross-process read-merge-write behind a fail-open advisory lock so an armed HARD-BLOCK gate can't be silently dropped.
**Why:** The host and MCP child share `write-gate-state.json` via a whole-file snapshot; atomic rename prevents torn files but not lost updates, so an unprotected interleave could overwrite `pendingGateId`/`activeQueuePhase` and let an agent proceed past an unconfirmed gate.
**How:** Wrap both read-merge-write seams in a `write-gate.lock` built on the existing `sync-lock` primitive, fail-open on contention.

## What

- `sync-lock.ts` — generalized `acquireSyncLock`/`releaseSyncLock` to take an optional lock name (default `sync.lock`, unchanged for the existing worktree-sync caller). Also hardened `acquireSyncLock` so a persistently-uncreatable lock path (dangling-symlink parent, `ENOTDIR`, read-only FS) **fails open instead of spinning its retry loop forever** and blocking the event loop.
- `bootstrap/write-gate.ts` — new `withWriteGateLock` wraps `mutateWriteGateState` and `childMutate` (the two read-merge-write seams). It materializes `.gsd/` first (so the lock engages under an external-state symlink instead of silently failing open), acquires the lock, and releases it in a `finally`. Added a test-only `_setWriteGateInterleaveHookForTest` seam.
- Tests — `write-gate-seam.test.ts` and `sync-lock.test.ts` extended.

## Why

The write-gate HARD-BLOCKs an agent from proceeding past unconfirmed depth/approval gates. Its state is shared between two OS processes with no cross-process lock; the merge assigns `pendingGateId`/`activeQueuePhase` last-writer-wins. The unprotected interleave: the child arms a gate and persists while the host is mid-mutation having read the snapshot *before* the child's write — the host's persist lands last and writes `pendingGateId = null`, dropping the armed gate. (Distinct from the already-fixed "host re-arm clobbered child verification" incident, which the `verified-wins-over-pending` rule covers.)

## How

The lock is **fail-OPEN by contract**: on contention it logs once and proceeds exactly as before (racy but functional) — it never blocks, throws, or refuses a gate mutation. It shrinks the lost-update window to lock-contention-fail-open, a deliberate residual (the next step, if audits ever show the warning firing, is a monotonic sequence number with refuse-stale-overwrite on the two fields).

The `acquireSyncLock` hardening was surfaced during verification: the added caller exposed a latent infinite-loop (spin at ~96% CPU) when the lock path's parent can't be created — it also reached the existing `workflow-reconcile` caller, so fixing it in the primitive protects both.

Verified-wins-over-pending (`dropVerifiedPendingGate`) and the "delete the snapshot file to clear the HARD BLOCK" escape hatch are unchanged.

## Change type

- [x] `fix` — Bug fix

## Scope

## Breaking changes

- [x] No breaking changes

## Test plan

- [x] New/updated tests included — interleave hook pins that the lock brackets the critical section (fails if the wrapping is ever removed); fail-open-under-held-lock and uncreatable-path-fails-open regressions; existing delete-file escape hatch and verified-wins rules still pass.
- [x] 256 tests green across sync-lock, write-gate, write-gate-seam, predicates, planning-unit, worktree-write-gate, register-hooks-depth-verification, workflow-reconcile, and the MCP-server bridge suites; `typecheck:extensions` clean.
- [x] CI passes

---

> [!NOTE]
> **Medium Risk**
> Changes concurrency around persisted HARD-BLOCK gate state shared by two processes; fail-open on lock contention leaves a small residual race, but behavior on contention matches the pre-fix path.
> 
> **Overview**
> Fixes a cross-process race where the extension host and workflow MCP child both update `write-gate-state.json`: atomic rename avoids torn files but **last-writer-wins** on `pendingGateId` / `activeQueuePhase` could let the host persist over a child-armed gate and drop a **HARD BLOCK**.
> 
> **Write gate:** `mutateWriteGateState` and `childMutate` now run inside `withWriteGateLock`, using `write-gate.lock` via the existing advisory lock. `.gsd/` is ensured first (including external-state symlink targets) so locking actually engages. On contention the path **fails open**—logs once and proceeds like before—so mutations are never refused or blocked.
> 
> **Sync lock:** `acquireSyncLock` / `releaseSyncLock` accept an optional lock name (default `sync.lock` unchanged). Non-`EEXIST` create failures **fail open** instead of spinning forever on uncreatable paths (e.g. dangling `.gsd` symlink).
> 
> **Tests:** Interleave hook proves the lock brackets the critical section; fail-open under held lock; uncreatable-path regression. Plan **030** marked DONE in `plans/README.md`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c4c0482228b917342d079cac804ccdf8182f9823. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-pi/pr/1342"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a>

### DIFF
--- a/plans/README.md
+++ b/plans/README.md
@@ -45,6 +45,7 @@ commit `58dc840f`. Plans 018–024 planned against commit `7cca07ae`.
 | 027 | Scope external-edit drift repair status authority to drifted entities | P1 | M | — | DONE |
 | 028 | Six small state/DB correctness fixes (status aliases, gate atomicity, claim locking, derive-cache lock, lease test) | P1 | S | — | DONE |
 | 029 | Reject path-traversal keys in the compat marker before drift detectors read them | P1 | S | — | DONE |
+| 030 | Stop the write-gate's cross-process read-merge-write from losing an armed gate | P2 | M | — | DONE |
 
 All of 009–017 implemented, tested, and committed on branch `chore/audit-fixes-009-017`
 (2026-07-01). Notable deviations recorded in the plan files: 009 skipped the
--- a/src/resources/extensions/gsd/bootstrap/write-gate.ts
+++ b/src/resources/extensions/gsd/bootstrap/write-gate.ts
@@ -10,6 +10,7 @@ import { loadJsonFileOrNull } from "../json-persistence.js";
 import { getIsolationMode } from "../preferences.js";
 import { compileSubagentPermissionContract, type ToolsPolicy } from "../unit-context-manifest.js";
 import { logWarning } from "../workflow-logger.js";
+import { acquireSyncLock, releaseSyncLock } from "../sync-lock.js";
 import { isGsdWorktreePath, resolveWorktreeProjectRoot } from "../worktree-root.js";
 import { worktreesDirs } from "../worktree-placement.js";
 import { bashReferencesProjectRootOutsideWorktree } from "../worktree-shell-guard.js";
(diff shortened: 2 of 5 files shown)


jeremymcs:

Merge conflicts were detected while updating this branch from the base branch.

Conflicted files:
- `plans/README.md`

An automatic resolution attempt is running. If it cannot complete safely, this PR will be marked for manual follow-up.

PR #1351, opened 2026-07-08
fix(issue): Skill loading instructions cause agents to construct paths instead of using <available_skills> location field

### LINKED ISSUE #1348: Skill loading instructions cause agents to construct paths instead of using <available_skills> location field
## Summary

The engine correctly discovers skills across all configured directories and emits an `<available_skills>` block with an exact `<location>` path for each skill. However, the hook instructions that tell the agent to load a skill use ambiguous language that allows the agent to construct its own path instead of extracting the `<location>` value. This produces ENOENT failures when the constructed path doesn't match where the skill actually lives.

## Evidence

Testing with the `execution-discipline` hook, the instruction emitted is:

```
- load `code-implementation` from `<available_skills>` using `read`
```

The agent sometimes interprets "from `<available_skills>`" as "confirm the skill exists in the list" rather than "extract the `<location>` child element and pass it verbatim to `read`". The agent then constructs a path based on convention (`~/.gsd/agent/skills/<name>/SKILL.md`) — which fails when the skill lives in a different directory (project-local, user-global, or Claude-compat).

The engine already knows the correct absolute path at hook-render time. `preferences-skills.js` `resolveAllSkillReferences()` resolves every `always_use_skills` and `prefer_skills` entry to an absolute path. That information is available but not passed through to the hook text.

## Root Cause

The instruction `load X from <available_skills> using read` is ambiguous to an LLM:

- **Interpretation A** (correct): Look up `<skill><name>X</name><location>/abs/path</location></skill>`, extract `<location>`, pass that exact path to `read`.
- **Interpretation B** (wrong): Confirm X exists in the list, then construct `~/.gsd/agent/skills/X/SKILL.md` and `read` that.

The LLM sometimes chooses interpretation B — typically defaulting to `~/.gsd/agent/skills/` because it's the most populous directory and appears first in precedence order.

## Proposed Fix

**Preferred approach**: Inline the resolved absolute path into the hook text at render time. `preferences-skills.js` already computes the path for every skill reference. The hook renderer could replace:

```
- load `code-implementation` from `<available_skills>` using `read`
```

with:

```
- read the skill file at `/abs/path/to/code-implementation/SKILL.md`
```

This removes the LLM path-construction step entirely — no XML parsing, no ambiguity.

**Alternative**: Keep the `<available_skills>` lookup but tighten the instruction language:

```
- find `code-implementation` in `<available_skills>`, copy its `<location>` value, and pass that exact path to `read`
```

## Scope

- **Component**: `preferences-skills.js` + hook text rendering
- **Affected hooks**: `execution-discipline` (skills block), `milestone-council` (planning skill loads), `slice-council` (planning skill loads)
- **Severity**: Medium — the failure is intermittent (depends on which interpretation the LLM chooses on a given turn), but when it fires it wastes a tool call and context tokens on an avoidable ENOENT.

## Verification

- Skills from project-local directories (`.agents/skills/`) are loaded without constructing `~/.gsd/agent/skills/` paths
- Skills from user-global directories (`~/.agents/skills/`) are loaded without constructing project-local paths
- Skills from Claude-compat directories (`~/.claude/skills/`) are loaded correctly
- No regressions in bundled skill loading (`~/.gsd/agent/skills/`)

jeremymcs:
## Summary
- Rendered skill activation now uses absolute skill file paths; diff check passed and focused tests were blocked by missing TypeScript dependency/network.

## Verification

## Related Issue
- Closes #1348
- [#1348 Skill loading instructions cause agents to construct paths instead of using <available_skills> location field](https://github.com/open-gsd/gsd-pi/issues/1348)

## User
- @mikenorgate

## Repo
- `open-gsd/gsd-pi`

## Branch
- `issue/1348-skill-loading-instructions-cause-agents--1783550817`

---

> [!NOTE]
> **Low Risk**
> Prompt-text and test-only changes in skill activation; no auth, data, or runtime API surface beyond clearer paths in generated prompts.
> 
> **Overview**
> **Skill activation** no longer tells agents to infer paths from `<available_skills>`; it now emits **concrete read instructions** using each installed skill’s `filePath`.
> 
> `formatSkillActivationBlock` takes a name→path map built from `getInstalledSkills`, and for each safe skill name outputs `Read the skill file at \`<escaped path>\``. Paths are passed through a new **`escapeXml`** helper so unusual characters in paths cannot break or inject into the prompt XML. If a path is missing, the fallback still points at `<available_skills>` but asks the model to **copy the `<location>` value** instead of guessing a path.
> 
> Unit and integration tests were updated to assert the new wording (including full expected paths under `.agents/skills` or test `skills/` dirs) across `buildSkillActivationBlock`, research/complete-slice prompts, and parallel subagent dispatch.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 632b0aba299a5d1c3a3d51bbc14650847aff2e18. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-pi/pr/1351"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a>

### DIFF
--- a/src/resources/extensions/gsd/skill-activation.ts
+++ b/src/resources/extensions/gsd/skill-activation.ts
@@ -131,10 +131,23 @@ function resolvePreferredSkillNames(
  *  to prevent prompt injection via crafted directory names. */
 const SAFE_SKILL_NAME = /^[a-z0-9][a-z0-9-]*$/;
 
-function formatSkillActivationBlock(skillNames: string[]): string {
+function escapeXml(str: string): string {
+  return str
+    .replace(/&/g, "&amp;")
+    .replace(/</g, "&lt;")
+    .replace(/>/g, "&gt;")
+    .replace(/"/g, "&quot;")
+    .replace(/'/g, "&apos;");
+}
+
+function formatSkillActivationBlock(skillNames: string[], skillPaths: Map<string, string>): string {
   const safe = skillNames.filter(name => SAFE_SKILL_NAME.test(name));
   if (safe.length === 0) return "";
-  const reads = safe.map(name => `Read the installed '${name}' skill file from <available_skills>`).join(". ");
+  const reads = safe.map(name => {
+    const path = skillPaths.get(name);
+    if (path) return `Read the skill file at \`${escapeXml(path)}\``;
+    return `Find '${name}' in <available_skills>, copy its <location> value, and pass that exact path to read`;
+  }).join(". ");
   return `<skill_activation>${reads}.</skill_activation>`;
 }
 
@@ -203,6 +216,7 @@ export function buildSkillActivationBlock(params: {
   const visibleSkills = loaded;
   const manifestScopedSkills = filterSkillsByManifest(visibleSkills, params.unitType);
   const installedNames = new Set(visibleSkills.map(skill => normalizeSkillReference(skill.name)));
+  const installedSkillPaths = new Map(visibleSkills.map(skill => [normalizeSkillReference(skill.name), skill.filePath]));
   warnIfManifestHasMissingSkills(params.unitType, installedNames);
   const avoided = new Set(resolvePreferenceSkillNames(prefs?.avoid_skills ?? [], params.base));
   const matched = new Set<string>();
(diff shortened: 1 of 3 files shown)


jeremymcs:

Re-checked PR #1351 against [#1348 Skill loading instructions cause agents to construct paths instead of using <available_skills> location field](https://github.com/open-gsd/gsd-pi/issues/1348).

## Subtasks
- [x] **Skill loading instructions cause agents to construct paths instead of using <available_skills> location field**

PR #1352, opened 2026-07-08
fix(issue): Auto-mode destructive command guard cannot complete its own confirmation flow — synchronous pause prevents ask_user_questions

### LINKED ISSUE #1347: Auto-mode destructive command guard cannot complete its own confirmation flow — synchronous pause prevents ask_user_questions
## Version

`@opengsd/gsd-pi` v1.9.0

## Bug

The destructive command guard confirmation flow is self-defeating in auto-mode. When a destructive bash command is blocked, the guard:

1. Returns instructions telling the model to call `ask_user_questions` with `destructive_confirm`
2. **Immediately calls `pauseAuto()` synchronously inside the same `tool_call` hook handler**

Because `pauseAuto()` fires before the model can issue a follow-up `ask_user_questions`, the confirmation dialog never appears. The user sees only:

> Destructive-command confirmation is waiting for your answer — pausing auto-mode.

...without ever having been asked a question. On auto-mode resume, the orchestrator re-dispatches the unit, hits `_isPauseOriginCancelledResult` in `unit-phase.js`, and cycles through `pause-during-setup` cancellations indefinitely. The confirmation can never be resolved because the model never gets a turn to present it.

## Where

`register-hooks.ts`, safety harness `tool_call` hook:

```typescript
if (classification.destructive) {
  // ...
  requestDestructiveConfirmation(command, guardBasePath);
  // Returns block reason telling model to call ask_user_questions...
  // ...then immediately pauses before the model can do it:
  await maybePauseAutoForApprovalGate(
    ctx,
    pi,
    isAutoActive(),
    "Destructive-command confirmation is waiting for your answer — pausing auto-mode.",
  );
  return { block: true, reason };
}
```

## Why the confirmation flow breaks

The confirmation flow requires two tool calls across one or two turns:

1. `ask_user_questions` — present the dialog
2. `bash` — re-issue the confirmed command (consumed by `consumeDestructiveConfirmation`)

`pauseAuto()` in the guard handler closes the turn window before the model can emit step 1. The guard tells the model to do something, then prevents it from doing it.

## Steps to reproduce

1. Auto-mode execute-task issues a destructive-classified bash command (e.g. `rm -f /tmp/some-file`)
2. Guard blocks it, returns "Call ask_user_questions with destructive_confirm..."
3. `pauseAuto()` fires synchronously — model cannot issue `ask_user_questions`
4. User sees "waiting for your answer" notification with no dialog
5. `/gsd auto` resume → unit start → `_isPauseOriginCancelledResult` → `pause-during-setup` → `unit-end cancelled`
6. Repeats on every subsequent dispatch

## Secondary: `rm -f` false positive as "recursive delete"

The regex `/\brm\s+(-[^\s]*[rfRF][^\s]*\s+|.*\s+-[^\s]*[rfRF])/` matches `rm -f single-file.md` as "recursive delete." The `f` in the flags triggers the pattern even when `-r`/`-R` is absent. Deleting a single file with `-f` is not destructive.

## Expected behavior

The model should be allowed to issue `ask_user_questions` in response to the block. The pause should only happen if the confirmation is not resolved within the turn, not before the model has a chance to present it.

## Proposed fix

**Guard timing:** Defer the pause to `agent_end` (similar to `deferApprovalGate`) instead of calling `maybePauseAutoForApprovalGate` synchronously in the `tool_call` handler.

**Regex fix:**

```diff
- { pattern: /\brm\s+(-[^\s]*[rfRF][^\s]*\s+|.*\s+-[^\s]*[rfRF])/, label: "recursive delete" },
+ { pattern: /\brm\s+(-[^\s]*[rR][^\s]*\s+|.*\s+-[^\s]*[rR])/, label: "recursive delete" },
```

## Feature request: disable/weaken control

No env var, preference key, or config option exists to disable or weaken the guard. A preference like:

```yaml
safety:
  destructive_command_guard: warn  # off | warn | block (default: block)
```

...would let users opt out and auto mode to continue uninterrupted.

jeremymcs:
## Summary
- Fixed destructive guard pause timing and rm -f false positives, with classifier tests passing and integration verification blocked by missing dependencies.

## Bugs Addressed
- [x] **Auto-mode pauses before destructive confirmation dialog appears**
- [x] **rm -f is incorrectly classified as recursive delete**

## Verification

## Related Issue
- Closes #1347
- [#1347 Auto-mode destructive command guard cannot complete its own confirmation flow — synchronous pause prevents ask_user_questions](https://github.com/open-gsd/gsd-pi/issues/1347)

## User
- @mikenorgate

## Repo
- `open-gsd/gsd-pi`

## Branch
- `issue/1347-auto-mode-destructive-command-guard-cann-1783550813`

---

> [!NOTE]
> **Medium Risk**
> Changes auto-mode pause timing and destructive bash classification in the safety harness; behavior is covered by new integration tests but affects a critical guard path.
> 
> **Overview**
> Fixes auto-mode getting stuck when a destructive bash command is blocked: **auto-mode no longer pauses synchronously inside the bash `tool_call` hook**, so the agent can still run `ask_user_questions` in the same turn. Pause is **deferred to `agent_end`** (mirroring depth approval gates) via a per–`basePath` flag that only triggers when a pending destructive command is still waiting.
> 
> The destructive-confirm lifecycle is tightened: declining or non-affirmative answers call **`resetDestructiveConfirmation`**, and deferred pause state is cleared on session boundaries and when the gate is answered.
> 
> **`classifyCommand`** no longer treats `rm -f` / `--force` as recursive delete; `rm --recursive` and force-only `rm` cases are covered by new tests, plus an integration test that destructive blocks do not emit the pause notification from the bash hook while auto is active.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d9b1e70b65e28202700d0f6023922c0a0c0c299b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-pi/pr/1352"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a>

### DIFF
--- a/src/resources/extensions/gsd/bootstrap/register-hooks.ts
+++ b/src/resources/extensions/gsd/bootstrap/register-hooks.ts
@@ -51,7 +51,9 @@ import {
   confirmDestructiveCommand,
   consumeDestructiveConfirmation,
   isDestructiveConfirmGateId,
+  peekPendingDestructiveCommand,
   requestDestructiveConfirmation,
+  resetDestructiveConfirmation,
 } from "../safety/destructive-confirmation.js";
 import { logWarning as safetyLogWarning, setStderrLoggingEnabled } from "../workflow-logger.js";
 import { isUnitCloseoutTool, runInteractiveUnitCloseout } from "../unit-closeout.js";
@@ -187,6 +189,7 @@ function suppressWelcomeHeader(ctx: ExtensionContext): void {
  * are bounded — cleared on activation, session boundaries, and verification.
  */
 const deferredApprovalGates = new Map<string, string>();
+const deferredDestructiveConfirmationPauses = new Set<string>();
 
 export const MINIMAL_GSD_TOOL_NAMES = [
   "gsd_exec",
@@ -618,6 +621,23 @@ function clearDeferredApprovalGate(basePath?: string): void {
   }
 }
 
+function deferDestructiveConfirmationPause(basePath: string): void {
+  deferredDestructiveConfirmationPauses.add(basePath);
+}
+
+function clearDeferredDestructiveConfirmationPause(basePath?: string): void {
+  if (!basePath) {
+    deferredDestructiveConfirmationPauses.clear();
+  } else {
+    deferredDestructiveConfirmationPauses.delete(basePath);
+  }
+}
+
+function isDestructiveConfirmationBlocking(basePath: string): boolean {
+  return deferredDestructiveConfirmationPauses.has(basePath)
+    && Boolean(peekPendingDestructiveCommand(basePath));
+}
+
 function deferApprovalGate(gateId: string, basePath: string): void {
   // Verified-on-disk wins (same adapter policy as activation/re-arm): if the
   // workflow MCP child already verified this gate, deferring would block
@@ -1015,6 +1035,7 @@ export function registerHooks(
     await applyToolCallLoopGuardConfig(basePath);
     approvalQuestionAbortInFlight = false;
     clearDeferredApprovalGate();
+    clearDeferredDestructiveConfirmationPause();
     await resetAskUserQuestionsTurnCache();
     await syncServiceTierStatus(ctx);
     await applyDisabledModelProviderPolicy(ctx);
@@ -1106,6 +1127,7 @@ export function registerHooks(
     resetToolCallLoopGuard();
     await applyToolCallLoopGuardConfig(basePath);
     clearDeferredApprovalGate();
+    clearDeferredDestructiveConfirmationPause();
     await resetAskUserQuestionsTurnCache();
     clearDiscussionFlowState(basePath);
     // /clear or /new destroys the conversation holding a discuss interview, so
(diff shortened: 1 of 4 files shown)


jeremymcs:

Re-checked PR #1352 against [#1347 Auto-mode destructive command guard cannot complete its own confirmation flow — synchronous pause prevents ask_user_questions](https://github.com/open-gsd/gsd-pi/issues/1347).

## Subtasks
- [x] **Destructive guard pauses before confirmation dialog can be shown**
- [x] **rm -f is incorrectly classified as recursive delete**

Issue #1292, opened 2026-07-06
[Bug]: marker-key mismatch triggers repeated whole-tree re-imports + leaked `.gsd-backups/migrate-*` dirs — reproduced on a real (non-symlinked) `.gsd`


evolv3ai:
## Summary

Re-imports against this project repeatedly performed a whole-tree re-import and left behind a `.gsd-backups/migrate-<epoch-ms>` directory each time. Over a single build day that accumulated **14 leaked backup directories** (13:25–18:37 local); audit cross-reference shows their timestamps cluster **in pairs at unit-dispatch boundaries**, so the trigger is the pre-dispatch reconcile/import pass (which also runs at process start), not process start alone. This is the marker-key false positive previously analyzed in the PR #1228 discussion (layers 2–3 of that root-cause write-up: `toArtifactPath` realpath/symlink producing dual marker keys) — but this reproduction is on a **real, non-symlinked `.gsd` directory**, so the symlink-specific fixes do not cover it.

Explicitly: **not fixed by #1254 / #1255** (those addressed the broken relative artifact path and the `SQLITE_READONLY_DBMOVED` through-symlink open, respectively), and the #1212 cleanup fix does not prevent the leak — the dirs still accumulate.

## Environment

- Observed 2026-07-05 on a local build of a main-lineage fork including the #1254/#1255 fixes; behavior spans v1.5.x–v1.6.0 code.
- Linux (WSL2), Node 24. Project `.gsd` is a plain directory (no symlink anywhere in the path).
- Re-verified 2026-07-06: no import-marker changes landed between `15a286bb` (v1.6.0) and `8b8c0f79` (v1.7.0). Distinct from closed #1209/#1212 (flat-phase migration backup leak on verification failure) — here the leak driver is the marker-key mismatch re-triggering full re-imports at dispatch boundaries.

## Observed

```
$ ls .gsd-backups/
migrate-1783275934833   (Jul 5 13:25)
migrate-1783279538104   (Jul 5 14:25)
migrate-1783279647855   (Jul 5 14:27)
migrate-1783279903295   (Jul 5 14:31)
migrate-1783281405659   (Jul 5 14:56)
migrate-1783281604437   (Jul 5 15:00)
migrate-1783284121571   (Jul 5 15:42)
migrate-1783284307356   (Jul 5 15:45)
migrate-1783288919562   (Jul 5 17:01)
migrate-1783289175458   (Jul 5 17:06)
[...]

PR #1378, opened 2026-07-09
fix(issue): [Bug]: compat marker-key mismatch still fires whole-tree re-import walks at every dispatch; writer/checker key derivation unshared, poisoned markers never healed (follow-up to #1292)

### LINKED ISSUE #1359: [Bug]: compat marker-key mismatch still fires whole-tree re-import walks at every dispatch; writer/checker key derivation unshared, poisoned markers never healed (follow-up to #1292)
## Summary

Follow-up to #1292, which was closed on #1305 (backup-dir reuse). #1305 plus the subsequent #1337 (scoped, status-preserving re-import) and #1340 (traversal-key rejection) fix the *consequences* of the marker-key mismatch — the disk leak, the status blast radius, and the arbitrary-read hole — but the mismatch itself (the root cause reported in #1292) remains unaddressed, and none of the three fix directions suggested there has landed. Verified on `main` @ `c855b965` (post-v1.9.0).

Requested in the closing discussion of #1292 ("happy to have this split into a follow-up issue… as long as the marker path itself stays tracked" — no response before closure), so splitting it out here.

## What still happens on current main

When a `.gsd/.compat.json` projection entry's recorded sha doesn't match the file on disk — including the permanent-mismatch case from #1292 where the writer refreshes a different key form than the one that goes stale — every reconcile pass at every dispatch boundary:

1. fires `external-markdown-edit` drift for the stale key (`state-reconciliation/drift/external-markdown-edit.ts`), and
2. runs a whole-tree `migrateHierarchyToDb` walk per drifted record (the repair comment itself notes it "walks the whole tree"; status authority is now correctly scoped per #1337, but the walk + backup handling still run).

Data is safe post-#1228/#1303/#1337; the residual cost is the #1292 cadence — a full-tree import walk at every unit dispatch, forever, on any project whose marker holds a key the writer no longer refreshes.

## Diagnosability regression (now worse than at #1292 time)

- The drift record carries `expectedSha`/`actualSha`, but nothing logs them (or the offending key) anywhere in `state-reconciliation/` — a field user cannot tell the mismatch is happening. With #1305 stopping the `migrate-*` dir accumulation, the mismatch has lost its only visible evidence, exactly the regression flagged in the #1292 discussion.
- New interaction with #1340: a marker that contains one traversal-form key (the dual-key symlinked-`.gsd` case from the #1228 layer-2 analysis — same physical plan file recorded under both a clean key and a `../../…` form) now invalidates the **whole** marker, and `readCompatMarker` fail-safes to the empty marker. `detectExternalMarkdownEdit` then returns no records — external-edit drift detection is silently disabled for that project, and the poisoned key is never healed. Fail-safe is right for a hostile marker, but for a self-inflicted writer-side key it means: no detection, no healing, no diagnostic.

## Root-cause pointer (unchanged)

`toArtifactPath` (`src/resources/extensions/gsd/markdown-renderer.ts:110`) still derives marker keys with raw `relative()` against `gsdProjectionRoot`/`gsdRoot` — no realpath canonicalization, and no single shared derivation used by both the writer paths and the marker-refresh paths. The #1292 reproduction was on a plain non-symlinked `.gsd`, so key-form drift is not limited to realpath/symlink divergence.

## Suggested fixes (same three as the #1292 discussion, still open)

1. **Log expected vs. found marker key + shas on mismatch** (one debug line in `detectExternalMarkdownEdit`) — restores field diagnosability now that the backup-dir evidence is gone.
2. **Derive marker keys through one shared function on the normalized (realpath'd) project root**, used by every writer and refresh site — removes the dual-key/format-drift class at the source; also lets `readCompatMarker` *heal* a stale-but-derivable key instead of (post-#1340) discarding the whole marker.
3. **Treat "marker mismatch but DB fully populated" as marker-refresh, not re-import** — bounds the worst case to a one-line marker write even if key logic regresses again.

## Environment

- `main` @ `c855b965` (v1.9.0 lineage), Linux (WSL2), Node 24.
- Prior evidence: #1292 report (14 `migrate-*` dirs/day, pair-clustered at dispatch boundaries, non-symlinked `.gsd`) and the #1228 layer-2/3 analysis (symlinked dual-key store forensics).

jeremymcs:
## Summary
- Fixed compat marker key healing and stale-marker refresh logic, with dependency-free checks passing and focused tests blocked by missing dependencies/network.

## Verification

## Related Issue
- Closes #1359
- [#1359 [Bug]: compat marker-key mismatch still fires whole-tree re-import walks at every dispatch; writer/checker key derivation unshared, poisoned markers never healed (follow-up to #1292)](https://github.com/open-gsd/gsd-pi/issues/1359)

## User
- @evolv3ai

## Repo
- `open-gsd/gsd-pi`

## Branch
- `issue/1359-bug-compat-marker-key-mismatch-still-fir-1783637731`

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-pi/pr/1378"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a>

---

> [!NOTE]
> **Medium Risk**
> Touches ADR-017 reconcile and marker persistence; incorrect DB-vs-disk matching could skip a real external edit, though scope is limited to artifact lookups and dry-run disables writes.
> 
> **Overview**
> Stops **repeated whole-tree re-imports** caused by `.gsd/.compat.json` projection keys that don’t match what the renderer writes, and by marker SHAs that lag disk even when content already matches the DB.
> 
> **Shared key derivation:** `deriveCompatProjectionKey` (realpath-aware, root-relative) replaces ad hoc `relative()` logic in the markdown renderer so writer and drift checker use the same keys.
> 
> **Marker healing:** `readCompatMarker` can rewrite derivable unsafe keys (e.g. traversal forms) to safe `.gsd`/`.planning` paths and persist the fix; quarantine/healing honor `dryRun` via `healInvalidKeys` / `quarantineInvalid`.
> 
> **Stale SHA without real drift:** `external-markdown-edit` compares mismatched files to DB artifact content (by path and scoped entity queries); when the on-disk file already matches canonical DB content, it **refreshes the marker SHA** instead of emitting external-edit drift. Planning drift detection uses the same read options.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d770ac3cda988568232f5d4842c2fb8814b618d5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/resources/extensions/gsd/compat/compat-marker.ts
+++ b/src/resources/extensions/gsd/compat/compat-marker.ts
@@ -6,8 +6,8 @@
 // (drift to import). gsd-core is oblivious to this file and ignores it.
 
 import { createHash } from "node:crypto";
-import { existsSync, readFileSync, renameSync, unlinkSync, writeFileSync, mkdirSync } from "node:fs";
-import { dirname, join, isAbsolute, normalize } from "node:path";
+import { existsSync, readFileSync, realpathSync, renameSync, unlinkSync, writeFileSync, mkdirSync } from "node:fs";
+import { dirname, join, isAbsolute, normalize, relative, resolve } from "node:path";
 
 /** Current marker schema version. Bump on breaking format changes + migrate. */
 export const COMPAT_MARKER_SCHEMA = 2;
@@ -54,6 +54,13 @@ export interface CompatMarker {
   piVersion: string;
 }
 
+export interface ReadCompatMarkerOptions {
+  /** Rewrite invalid-but-derivable marker keys back to safe root-relative keys. */
+  healInvalidKeys?: boolean;
+  /** Quarantine malformed/unsafe markers. Disable only for read-only previews. */
+  quarantineInvalid?: boolean;
+}
+
 /** Marker returned when no marker exists yet (fresh project, first gsd-pi run). */
 export const EMPTY_MARKER: CompatMarker = {
   schema: COMPAT_MARKER_SCHEMA,
@@ -81,6 +88,37 @@ export function computeProjectionSha(content: string): string {
   return createHash("sha256").update(normalizeForHash(content)).digest("hex").slice(0, 16);
 }
 
+function normalizeRealPath(p: string): string {
+  try {
+    return realpathSync.native(p);
+  } catch {
+    return resolve(p);
+  }
+}
+
+function rootRelativeKey(rootPath: string, absPath: string): string | null {
+  const root = normalizeRealPath(rootPath);
+  const abs = normalizeRealPath(absPath);
+  const rel = relative(root, abs);
+  if (!rel || rel === ".." || rel.startsWith(`..${sepForRelative(rel)}`) || isAbsolute(rel)) {
+    return null;
+  }
+  return rel.replace(/\\/g, "/");
+}
+
+function sepForRelative(rel: string): string {
+  return rel.includes("\\") ? "\\" : "/";
+}
+
+export function deriveCompatProjectionKey(absPath: string, roots: readonly string[]): string {
+  for (const root of roots) {
+    const key = rootRelativeKey(root, absPath);
+    if (key) return key;
+  }
+  const fallbackRoot = roots[roots.length - 1] ?? dirname(absPath);
+  return relative(fallbackRoot, absPath).replace(/\\/g, "/");
+}
+
 /**
  * Read & validate the marker. A missing marker → EMPTY_MARKER (treat every
  * projection as external on next reconcile). A malformed marker is quarantined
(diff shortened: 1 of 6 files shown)


jeremymcs:

Re-checked PR #1378 against [#1359 [Bug]: compat marker-key mismatch still fires whole-tree re-import walks at every dispatch; writer/checker key derivation unshared, poisoned markers never healed (follow-up to #1292)](https://github.com/open-gsd/gsd-pi/issues/1359).

## Subtasks
- [x] **[Bug]: compat marker-key mismatch still fires whole-tree re-import walks at every dispatch; writer/checker key derivatio**

PR #1380, opened 2026-07-10
fix(gsd): latch rejected ScheduleWakeup continuations

### LINKED ISSUE #1379: [Bug]: post-#1148, a rejected ScheduleWakeup still silently stalls unattended auto-mode on 1.9.0 — no arm-verification / stall heartbeat
Follow-up to #1148 (fixed via #1174) and #43. On **1.9.0** — which ships #1174 (merged 2026-07-02; 1.9.0 published 2026-07-08) — an unattended auto-mode / auto-orchestrator run **still silently stalled**, because its continuation `ScheduleWakeup` was rejected and the failure was neither surfaced nor recovered.

## Problem

During an unattended auto-mode run, the agent scheduled a continuation ("resume in ~15 min") via `ScheduleWakeup`. The call was **rejected** with:

```
`prompt` is required when `stop` is not true
```

The agent nonetheless reported to the operator that it had "scheduled a 15-minute wakeup." Nothing fired. The run then sat idle indefinitely — **no check-in, no progress, no error surfaced, no notification** — until a human intervened ~15+ minutes later.

## Why this is distinct from #1148 / #43 (not a duplicate)

#1174 fixed interactive `ScheduleWakeup` dispatch and cancelled-unit wakeup cleanup. This is a **different failure path** and a **residual systemic gap** that survives on 1.9.0:

- The wakeup here failed **schema validation** — a malformed call missing the required `prompt` — not the `auto_mode_inactive` path #1148 addressed.
- Regardless of *why* the call fails, the underlying resilience gap remains: **a failed `ScheduleWakeup` is reported to the operator as success, and nothing verifies the wakeup was actually armed or detects the resulting silent stall.** #1148's own secondary defect — *"reports success before the wakeup is guaranteed to be consumed"* — is the same theme, still reachable via this path.

## Expected

An unattended run must not be able to silently stall on a failed continuation:

- a rejected/errored `ScheduleWakeup` should **not** be reported as success (fail loud, or retry);
- the orchestrator should **verify** the wakeup was armed before treating itself as "paused pending resume"; and/or
- a **stall-detecting heartbeat/watchdog** should surface "run went silent without completing or erroring" — a safety net independent of any single tool call.

## Environment

- `@opengsd/gsd-pi` **1.9.0** (published 2026-07-08; includes #1174, merged 2026-07-02)
- Driver: Claude Code, auto-orchestrator (unattended) mode

## Evidence

- Tool result on the `ScheduleWakeup` call: `` `prompt` is required when `stop` is not true ``.
- ~15+ minutes idle with no check-in, no error, and no notification before manual recovery.

## Suggested fix (any one helps; together they're robust)

1. **Fail loud on a failed wakeup** — never report success when the `ScheduleWakeup` call errored; retry or surface it to the operator.
2. **Verify the arm** — confirm the continuation was accepted before the loop considers itself safely paused.
3. **Heartbeat / watchdog** — if a run is neither completed nor errored and hasn't progressed within its scheduled interval + grace, emit an operator notification. This catches silent stalls regardless of the underlying cause.

jeremymcs:
## TL;DR

**What:** Treat the exact rejected Claude Code `ScheduleWakeup` continuation as a latched auto-mode failure.
**Why:** A later successful read or diagnostic tool could otherwise erase the rejection before the unit boundary, leaving an unattended run silent with no armed wakeup.
**How:** Scope classification to the exact tool and validation text, retain that failure until `ScheduleWakeup` succeeds or post-unit handling pauses, and cover the universal external hook end to end.

## What

This change handles the exact Claude Code rejection:

```
`prompt` is required when `stop` is not true
```

Only `ScheduleWakeup` receives this special treatment. The failure survives unrelated successful and failed tool calls in the same turn. A successful `ScheduleWakeup` clears it; otherwise the existing post-unit path emits an error, pauses auto-mode, and does not schedule a verification retry.

The regression drives the real universal `tool_execution_end` hook used by external engines and verifies the downstream pause/notification behavior.

## Why

The initial classifier-only fix had two gaps:

- its conditional-required-field regex could classify similarly worded business validation from unrelated tools;
- any later successful tool result cleared `lastToolInvocationError`, so the failed continuation could still disappear before post-unit handling.

This revision closes both gaps without adding a general heartbeat policy.

Closes #1379

## How

- Match the exact missing-prompt error only when the tool is `ScheduleWakeup`.
- Centralize tool-error state updates so a latched continuation cannot be overwritten by another tool error.
- Pass successful tool identity into error clearing, allowing only successful `ScheduleWakeup` to clear the latch mid-unit.
- Preserve existing behavior for all other invocation, policy, user-skip, business, and network errors.

## Verification

- [x] RED: hook regression proved unrelated success erased the wakeup failure before the fix
- [x] RED: negative regression proved the broad regex classified identical text from another tool
- [x] GREEN: focused classifier, hook, runtime-state suites pass (48/48)
- [x] Mutation checks: disabling clear protection and overwrite protection each fails the hook regression
- [x] `pnpm run typecheck:extensions`
- [x] `pnpm run verify:fast` from a clean detached worktree
- [x] GitHub full CI

## Impact analysis

- **Blast radius:** Auto-mode tool-error retention and the two native/universal success hook paths.
- **Affected workflow:** Claude Code external-engine continuation scheduling during an active unit.
- **Compatibility:** Generic conditional business validation is not classified. Existing deterministic failures retain their prior clear-on-success behavior.

## Change type

- [x] `fix` — Bug fix

### DIFF
--- a/src/resources/extensions/gsd/auto-runtime-state.ts
+++ b/src/resources/extensions/gsd/auto-runtime-state.ts
@@ -3,11 +3,10 @@ import { AutoSession } from "./auto/session.js";
 import type { CurrentUnit } from "./auto/session.js";
 import type { SourceObservationStore } from "./source-observations.js";
 import {
-  isDeterministicPolicyError,
-  isQueuedUserMessageSkip,
-  isToolInvocationError,
   markToolEnd as markTrackedToolEnd,
   markToolStart as markTrackedToolStart,
+  shouldClearToolInvocationErrorAfterSuccess,
+  updateToolInvocationError,
 } from "./auto-tool-tracking.js";
 // Re-exported as a pure pass-through. Must stay UNGATED (no autoSession.active
 // argument, unlike markToolStart at the bottom of this file) so it is true in
@@ -99,13 +98,19 @@ export function markToolEnd(toolCallId: string): void {
 
 export function recordToolInvocationError(toolName: string, errorMsg: string): void {
   if (!autoSession.active) return;
-  if (isToolInvocationError(errorMsg) || isQueuedUserMessageSkip(errorMsg) || isDeterministicPolicyError(errorMsg)) {
-    autoSession.lastToolInvocationError = `${toolName}: ${errorMsg}`;
-  }
+  autoSession.lastToolInvocationError = updateToolInvocationError(
+    autoSession.lastToolInvocationError,
+    toolName,
+    errorMsg,
+  );
 }
 
-export function clearToolInvocationError(): void {
+export function clearToolInvocationError(successfulToolName?: string): void {
   if (!autoSession.active) return;
+  if (!shouldClearToolInvocationErrorAfterSuccess(
+    autoSession.lastToolInvocationError,
+    successfulToolName,
+  )) return;
   autoSession.lastToolInvocationError = null;
 }
 
--- a/src/resources/extensions/gsd/auto-tool-tracking.ts
+++ b/src/resources/extensions/gsd/auto-tool-tracking.ts
@@ -130,7 +130,7 @@ export function clearInFlightTools(): void {
  * from the tool handler. When these errors occur, retrying the same unit will
  * produce the same failure, so the retry loop must be broken.
  */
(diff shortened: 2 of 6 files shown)


jeremymcs:
Auto Engine risk verification completed:

- Root cause is limited to external-tool validation wording: Claude Code’s native `ScheduleWakeup` can return `` `prompt` is required when `stop` is not true ``, which the deterministic classifier did not recognize.
- The universal `tool_execution_end` hook already records classified external-engine failures, and the post-unit path already pauses auto-mode when `lastToolInvocationError` is set. This patch changes only the classifier predicate.
- Focused regression was observed RED, GREEN, and mutation-sensitive. Adjacent engine-hook/runtime/post-unit suites passed 60/60.
- Clean GitHub CI passed the full blocking build, including 12k+ unit tests, package/cloud tests, integration tests, and built-binary e2e.
- Existing assertions confirm normal business-logic and network errors remain outside the deterministic path.

No additional trigger, dispatch, or loop-termination code is changed.

Issue #758, opened 2026-06-16
MCP server: readiness gate, stdin idle watchdog, and PID registry for orphan cleanup


jeremymcs:
## Problem

When Claude Code launches the GSD workflow MCP server, there is a startup window where tools are not yet available. The system optimistically passes through `"pending"` server status (tool-surface-readiness.ts:62), letting tool calls fire before the server is ready. This produces cascading "No such tool available" failures across memory queries, capture_thought, gsd_exec, milestone_status, and UAT tools.

Additionally, orphaned MCP server processes accumulate when Claude Code exits ungracefully (crash, force-quit, OOM). The child process has SIGTERM/SIGINT/stdin-end handlers but nothing for when the parent simply vanishes — the process spins on a dead stdio pipe indefinitely (observed: PID at 97.6% CPU running since Sunday).

Web mode already has PID files, an instance registry (`~/.gsd/web-instances.json`), and stale-instance cleanup. MCP mode has **none of this**.

## Scope

Three independent, composable fixes:

### 1. MCP readiness gate — block on "pending" status

**File:** `src/resources/extensions/gsd/tool-surface-readiness.ts`

Change `getToolSurfaceReadinessError()` to return a `TOOL_SURFACE_NOT_READY` error when the workflow MCP server status is `"pending"` instead of returning `null` (pass-through).

This triggers the existing recovery flow:
- **Auto-mode** (`auto-post-unit.ts`): classifies as `tool-unavailable` → 3 retries with escalating delay (1s, 2s, 3s), then pause
- **Stream-adapter** (`stream-adapter.ts:2060`): aborts the unit before the first model turn with a transient error → retry at the dispatch layer

The key change is removing the optimistic pass-through at line 62 for `"pending"` status and instead treating it as a transient readiness failure.

### 2. Stdin idle watchdog in MCP server

**File:** `packages/mcp-server/src/cli.ts`

Add a periodic check (every 60s) that tracks last stdin activity. If no stdin data received for 5 minutes and no tool calls are in-flight, self-terminate with a clean shutdown via the existing `cleanup()` function.

This kills orphaned processes where the parent died without closing stdin/sending SIGTERM.

Implementation:
- Track `lastStdinActivity` timestamp, updated on every `process.stdin` data event
- `setInterval` every 60s: check `Date.now() - lastStdinActivity > 300_000` and no in-flight requests
- Call existing `cleanup()` for graceful shutdown
[...]

PR #1391, opened 2026-07-10
fix(issue): [Bug]: Orphaned MCP servers busy-loop at ~100% CPU for hours (SIGTERM-proof) — orphan cleanup from #758/#783 can't fire because the event loop is blocked

### LINKED ISSUE #1384: [Bug]: Orphaned MCP servers busy-loop at ~100% CPU for hours (SIGTERM-proof) — orphan cleanup from #758/#783 can't fire because the event loop is blocked
## Summary

During **autonomous GSD** runs, `@opengsd/mcp-server` processes are being **orphaned** (reparented to `launchd`, `PPID 1`) and left **busy-looping at ~100% CPU each, for hours**, after their parent Claude Code session exits. They are **not reaped** by the stdin idle watchdog / PID registry added in #758, and they **ignore `SIGTERM`** — only `SIGKILL` (`kill -9`) stops them.

Today I found **three** of them running simultaneously (~96–118% CPU each ≈ 3 full cores) — two alive for **7h45m**, one for **13h20m**. On a laptop this is a serious battery + thermal drain, and it compounds with every parallel autonomous session.

This looks like the same class of failure as the already-closed #783 ("orphaned MCP servers spin at 100% CPU") and the busy-wait/event-loop-blocking bugs #882, #884, #901 — but the orphan-cleanup mechanism from #758 does not save us in practice. See root-cause below for why.

## Environment

- `@opengsd/mcp-server` **1.2.0** (`~/.nvm/.../lib/node_modules/@opengsd/mcp-server`)
- Node **v22.18.0**
- macOS **26.3** (Darwin 25.3.0), Apple Silicon
- Usage pattern: multiple concurrent GSD sessions + auto-mode / autonomous phases, many short-lived Claude Code child sessions spawning MCP servers.

## Evidence

```
  PID  PPID  %CPU     ELAPSED  COMMAND
70855     1  96.0    13:20:49  node .../@opengsd/mcp-server/bin/gsd-mcp-server.js
 7352     1  97.0    07:45:24  node .../@opengsd/mcp-server/bin/gsd-mcp-server.js
 7523     1  96.5    07:45:14  node .../@opengsd/mcp-server/bin/gsd-mcp-server.js
```

- `PPID 1` → parent session already dead; process reparented to `launchd`.
- Meanwhile every **live-session** MCP server (child of an active `claude` PID) sat at **0.0% CPU** — so this is specific to the orphaned state, not normal operation.
- `kill <pid>` (SIGTERM) had **no effect** — processes stayed in state `R`. Only `kill -9` cleared them.

## Reproduction

Not deterministic, but reliably recurs under autonomous GSD. Rough recipe:
1. Run GSD auto-mode / multiple parallel sessions so many MCP servers spawn.
2. Let parent Claude sessions exit uncleanly (crash, kill, session churn, laptop sleep).
3. Some spawned `gsd-mcp-server` processes survive with `PPID 1` and pin a core indefinitely.

## Root cause (why #758's cleanup doesn't fire)

I inspected the installed 1.2.0 build:

- `dist/cli.js` **does** register `process.on('SIGTERM')` / `SIGINT` handlers.
- `dist/session-manager.js` **does** contain the idle/watchdog mechanism.

Yet the orphans were in kernel state **`R`** and **ignored SIGTERM**. That only happens when **the Node event loop is blocked by a synchronous busy-loop**. When the loop is blocked:

- The JS-level `process.on('SIGTERM')` handler **can never be dispatched** → SIGTERM is effectively swallowed → only SIGKILL works. ✔ matches observation.
- Any **timer-based idle watchdog never fires** → the orphan-cleanup path from #758 never runs → the process survives for hours. ✔ matches observation.
- The stdin `end`/`close` callback that would detect parent death **never gets dispatched** either. ✔

In other words: **the cleanup logic lives inside the very runtime it is supposed to clean up.** Once a synchronous busy-loop pegs the loop (cf. the busy-wait family #882 session-lock, #884 notification-store, #901 auto-loop-no-backoff), every in-process safety net — signal handler, idle watchdog, stdin-close detector — is disarmed at exactly the moment it's needed.

## Impact

- ~1 full CPU core burned **per orphan**, indefinitely (hours observed) → battery drain + fan/thermal throttling on laptops.
- Compounds linearly with parallelism — 3 today, and autonomous GSD spawns many.
- Silent: nothing surfaces it; users only notice via fans/battery and have to hunt it down manually.

## Suggested direction

The real fix probably can't live inside the blockable event loop:

1. **Out-of-process / event-loop-independent reaper.** Detect parent death without relying on the main loop — e.g. a tiny detached supervisor, a worker thread that watches `getppid()==1` (or the original parent PID) and calls `process.exit`/hard-exits, or spawn with a death-signal so the OS kills the child when the parent dies. This survives a blocked loop.
2. **Bound / remove the synchronous busy-waits** (#882, #884, #901 territory) so the loop can't be pegged unbounded in the first place — which would also let the existing SIGTERM handler + #758 watchdog actually run.
3. **Persisted PID registry + startup sweep**, so the next session (or a `gsd doctor`) can `SIGKILL` any registered MCP PID whose parent is gone. A user-facing `gsd doctor --reap-orphans` would help too.

## Current workaround

Manually reaping only orphaned (PPID 1), high-CPU gsd servers — never touches ones bound to a live session:

```bash
ps -Ao pid,ppid,%cpu,command | awk '$2==1 && /gsd-mcp-server/ && $3>20 {print $1}' | xargs -r kill -9
```

jeremymcs:
## Summary
- Added a worker-thread MCP orphan monitor with a busy-loop orphan regression and verified it with diff checks plus a dependency-free reaping harness.

## Verification

## Related Issue
- Closes #1384
- [#1384 [Bug]: Orphaned MCP servers busy-loop at ~100% CPU for hours (SIGTERM-proof) — orphan cleanup from #758/#783 can't fire because the event loop is blocked](https://github.com/open-gsd/gsd-pi/issues/1384)

## User
- @SashaMarchuk

## Repo
- `open-gsd/gsd-pi`

## Branch
- `issue/1384-bug-orphaned-mcp-servers-busy-loop-at-10-1783716564`

---

> [!NOTE]
> **Medium Risk**
> Changes process lifecycle and uses SIGKILL from a worker; behavior is gated on parent loss plus stdin idle, with test-only main-thread fallback when `isOrphaned` is mocked.
> 
> **Overview**
> Fixes orphaned MCP servers that **busy-loop at ~100% CPU** and ignore prior main-thread orphan/idle shutdown because the **event loop never runs**.
> 
> **Production path** replaces the ref-held main-thread orphan interval with a **`worker_threads` monitor** that shares stdin last-activity time via **`SharedArrayBuffer` / `Atomics`**. On a separate event loop it periodically checks **parent loss + stdin idle past the timeout**, then **`SIGKILL`s the server PID** instead of calling graceful `cleanup()`.
> 
> Tests keep the **main-thread orphan monitor** when `isOrphaned` is injected. Timeout/check intervals are **configurable** on `runMcpServerCli` for faster tests.
> 
> Adds a **POSIX integration test** that spawns a real child MCP server that enters an infinite loop after connect, drops the parent, and asserts the child process is reaped within a few seconds.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 599ac621ca14354337204d494f2ae4a32d2ac397. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-pi/pr/1391"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a>

### DIFF
--- a/packages/mcp-server/src/cli-runner.test.ts
+++ b/packages/mcp-server/src/cli-runner.test.ts
@@ -55,6 +55,68 @@ function spawnMcpServer(projectDir: string, gsdHome: string): ChildProcessWithou
   });
 }
 
+function spawnBusyLoopingMcpServerParent(projectDir: string, gsdHome: string): ChildProcessWithoutNullStreams {
+  const runnerUrl = new URL('./cli-runner.js', import.meta.url).href;
+  const childCode = `
+    import { runMcpServerCli } from ${JSON.stringify(runnerUrl)};
+    await runMcpServerCli({
+      sweepProjectOrphanMcpServers() {},
+      stdinIdleTimeoutMs: 100,
+      orphanParentLossCheckIntervalMs: 25,
+      createMcpServer: async () => ({
+        server: {
+          connect: async () => {
+            process.stderr.write('BUSY_READY\\n');
+            while (true) {}
+          },
+          close: async () => {},
+        },
+      }),
+      importStdioServerTransport: async () => ({ StdioServerTransport: class {} }),
+      warmWorkflowToolBridges() {},
+    });
+  `;
+  const parentCode = `
+    import { spawn } from 'node:child_process';
+    const child = spawn(process.execPath, ['--input-type=module', '--eval', ${JSON.stringify(childCode)}], {
+      cwd: ${JSON.stringify(projectDir)},
+      env: { ...process.env, GSD_HOME: ${JSON.stringify(gsdHome)} },
+      stdio: ['pipe', 'ignore', 'pipe'],
+    });
+    if (!child.pid) throw new Error('missing child pid');
+    process.stdout.write(String(child.pid) + '\\n');
+    let done = false;
+    child.stderr.on('data', (chunk) => {
+      const text = String(chunk);
+      process.stderr.write(text);
+      if (!done && text.includes('BUSY_READY')) {
+        done = true;
+        process.exit(0);
+      }
+    });
+    child.once('exit', (code, signal) => {
+      if (done) return;
+      done = true;
+      process.stderr.write('busy child exited before parent loss: code=' + code + ' signal=' + signal + '\\n');
+      process.exit(1);
+    });
+    setTimeout(() => {
+      if (done) return;
+      done = true;
+      process.stderr.write('timed out waiting for busy child readiness\\n');
+      process.exit(2);
+    }, 5000);
+  `;
+  return spawn(process.execPath, ['--input-type=module', '--eval', parentCode], {
+    cwd: projectDir,
+    env: {
+      ...process.env,
+      GSD_HOME: gsdHome,
+    },
+    stdio: ['ignore', 'pipe', 'pipe'],
+  });
+}
+
 async function waitForRegistryPid(gsdHome: string, pid: number | undefined, timeoutMs = 5_000): Promise<void> {
   assert.ok(pid, 'spawned child must have a pid');
   const registryPath = join(gsdHome, 'mcp-instances.json');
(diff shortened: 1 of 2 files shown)


jeremymcs:

Re-checked PR #1391 against [#1384 [Bug]: Orphaned MCP servers busy-loop at ~100% CPU for hours (SIGTERM-proof) — orphan cleanup from #758/#783 can't fire because the event loop is blocked](https://github.com/open-gsd/gsd-pi/issues/1384).

## Subtasks
- [x] **[Bug]: Orphaned MCP servers busy-loop at ~100% CPU for hours (SIGTERM-proof) — orphan cleanup from #758/#783 can't fire **

jeremymcs:
**Verifying** — checking the applied changes.
**Resolved** — addressed in commit `599ac62`.

jeremymcs:

jeremymcs:
Addressed in commit `599ac62`.

> Responding to [comment by @blacksmith-sh[bot]](https://github.com/open-gsd/gsd-pi/pull/1391#issuecomment-4939519500):
> 

Fixed the MCP orphan monitor worker crash by replacing CommonJS `require` in the eval worker source with ESM imports, allowing the busy-looped orphan server test to pass. Also corrected the spawned-process test helper types so package test compilation succeeds.

PR #1461, opened 2026-07-13
fix(gsd): keep milestone leases alive during unit execution

### LINKED ISSUE #1460: fix(auto): keep milestone lease alive during long unit execution
## Summary

Auto-mode refreshes its worker and milestone lease only at the top of each orchestration loop. A provider-backed unit can remain in flight longer than the 60-second lease TTL, so an otherwise healthy task loses settlement authority even when no other worker takes over.

## Reproduction

1. Start a canonical `execute-task` unit with a held milestone lease.
2. Let the provider call remain pending longer than the milestone lease TTL.
3. Allow the task to finish successfully.
4. Observe settlement fail with `workflow attempt requires the current held lease or current replacement lease`.

## Expected behavior

While auto-mode is actively awaiting a unit, it should periodically refresh the same worker and milestone lease. A stale or replaced token must still fail closed.

## Actual behavior

The loop is blocked awaiting the provider call, so its loop-top heartbeat never runs. The lease expires and the database fencing trigger correctly rejects the terminal Attempt transition.

## Approved approach

Keep the database fencing trigger unchanged. Run the existing heartbeat operation periodically only while a unit execution boundary is in flight, at half the lease TTL, and always clear the timer in `finally`. If a refresh misses because the lease was replaced or released, clear the session token and preserve fail-closed settlement.

The project developer approved this approach for the current refactor run.

## Acceptance criteria

- A real-DB auto-loop regression proves a pending task call refreshes the same worker/token lease before expiry.
- The task can stage and settle successfully after its original expiry point without a peer takeover.
- The interval is cleared when the unit exits, including errors.
- Refresh misses retain the existing fail-closed behavior.
- The database fencing trigger and replacement-lease recovery contract are unchanged.

## Impact

Severity: high for long-running tasks. A healthy execution can be stranded solely because provider latency exceeds one minute, blocking autonomous work until recovery is performed.

jeremymcs:
## TL;DR

**What:** Keep the active worker and milestone lease alive while an automatic unit is awaiting a long provider call.
**Why:** A healthy task could outlive the 60-second lease TTL and then be rejected by the database when it tried to settle.
**How:** Wrap both automatic execution paths in a half-TTL heartbeat that starts immediately and always stops when the unit resolves or rejects.

## What

- Adds a scoped async heartbeat boundary for long-running units.
- Uses the same behavior in built-in and custom-engine auto execution.
- Adds a real-database regression that expires the original lease while a Task is pending, advances the heartbeat, and proves successful settlement.
- Adds rejection-path coverage proving the interval is cleared after execution errors.

## Why

Auto-mode previously refreshed leases only at the top of its orchestration loop. While it awaited a provider call, the loop could not refresh again. Calls longer than the lease TTL therefore lost settlement authority even when no peer worker took over.

Closes #1460

## How

The existing worker heartbeat is run immediately before the unit boundary and every half-TTL while it remains pending. Cleanup lives in the async wrapper's `finally` block. A missed refresh still clears the session token, so stale or replaced workers continue to fail closed through the existing database fencing trigger.

The database trigger, replacement-lease recovery, and retry policy are unchanged.

## Verification

- RED: real-DB pending Task left the forced-expired lease unchanged on prior behavior.
- GREEN: original worker/token lease renews, Task settles succeeded, dispatch completes, and no refresh occurs after exit.
- Sabotage: delaying the interval beyond the test window fails the renewal assertion.
- Error sabotage: removing cleanup produces extra heartbeat calls after rejection (`4 !== 2`).
- `pnpm run typecheck:extensions`
- `pnpm run test:changed:src` — 139 passed
- `pnpm run verify:fast`
- Independent correctness and simplicity reviews — no findings after follow-up

The local merge-equivalent run passed build, web build, pack/install validation, coverage, and branch-specific tests before its aggregate unit phase encountered unrelated macOS path assertions that reproduce on detached `origin/main`; hosted CI is the clean full-suite gate for this draft.

## Change type

- [x] `fix` — Bug fix

## Breaking changes

None.

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-pi/pr/1461"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786553568&installation_id=134682257&pr_number=1461&repository=open-gsd%2Fgsd-pi&return_to=https%3A%2F%2Fgithub.com%2Fopen-gsd%2Fgsd-pi%2Fpull%2F1461&signature=82f44e7221e85e1f065067e612c34274486cb3d04951793324838b6aabab4dfc"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

---

> [!NOTE]
> **Medium Risk**
> Touches auto-loop coordination and milestone lease fencing during long-running dispatches; behavior is scoped and fail-closed on refresh miss, but incorrect timing or cleanup could still affect settlement authority.
> 
> **Overview**
> Fixes auto-mode losing milestone lease authority when a unit waits on a long provider call longer than the lease TTL (~60s), because heartbeats previously ran only at the top of each orchestration loop.
> 
> Adds **`runWithWorkerHeartbeat`**, which runs an immediate heartbeat and then **`maintainWorkerHeartbeat`** on a **half-TTL** interval (`milestoneLeaseTtlSeconds() * 500` ms) for the duration of an async unit run, with **`clearInterval`** in a `finally` block so the timer stops on success or rejection. Both **built-in** and **custom-engine** paths wrap task execution (`runWithTaskExecutionAttempt` → `runUnitPhaseViaContract`) in this boundary; loop-top heartbeat logic is unchanged but **`workerHeartbeatDeps`** is shared.
> 
> Tests add a real-DB auto-loop case that forces lease expiry while a task is pending, ticks the interval, and asserts renewal and successful settlement, plus a unit test that no further heartbeats run after execution rejects.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1978d3ba5a6ad621e961bdbc7c7717417826694a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/resources/extensions/gsd/auto/loop.ts
+++ b/src/resources/extensions/gsd/auto/loop.ts
@@ -43,7 +43,12 @@ import {
   getRecentUnitKeysForProjectRoot,
   markLatestActiveForWorkerCanceled,
 } from "../db/unit-dispatches.js";
-import { claimMilestoneLease, refreshMilestoneLease, forceReleaseLeasesForWorker } from "../db/milestone-leases.js";
+import {
+  claimMilestoneLease,
+  refreshMilestoneLease,
+  forceReleaseLeasesForWorker,
+  milestoneLeaseTtlSeconds,
+} from "../db/milestone-leases.js";
 import { heartbeatAutoWorker, getAutoWorker, markWorkerCrashed } from "../db/auto-workers.js";
 import { getRuntimeKv, setRuntimeKv } from "../db/runtime-kv.js";
 import { resolveUokFlags } from "../uok/flags.js";
@@ -83,7 +88,7 @@ import { createWorkflowPhaseReporter } from "./workflow-phase-reporter.js";
 import { createWorkflowTurnReporter } from "./workflow-turn-reporter.js";
 import { validateWorkflowSessionLock } from "./workflow-session-lock.js";
 import { dequeueSidecarItem } from "./workflow-sidecar-queue.js";
-import { maintainWorkerHeartbeat } from "./workflow-worker-heartbeat.js";
+import { maintainWorkerHeartbeat, runWithWorkerHeartbeat } from "./workflow-worker-heartbeat.js";
 import { gsdRoot } from "../paths.js";
 import {
   measureMemoryPressure,
@@ -196,6 +201,7 @@ function resolveCompletionStopFromOutcome(
 const STUCK_RECOVERY_ATTEMPTS_KEY = "stuck_recovery_attempts";
 const MAX_CONSECUTIVE_ALREADY_ACTIVE_SKIPS = 3;
 const MAX_CONSECUTIVE_ORCHESTRATION_SKIPS = 3;
+const WORKER_HEARTBEAT_INTERVAL_MS = milestoneLeaseTtlSeconds() * 500;
 const ORCHESTRATION_MISSING_REASON =
   "Auto Orchestration Module is not wired; cannot dispatch built-in GSD Unit.";
 const TASK_EXECUTION_CUTOVER_DEPS = {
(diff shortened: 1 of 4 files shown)


jeremymcs:
Critical Auto Engine follow-up completed:

- Root cause confirmed at the loop-top-only heartbeat versus a provider call longer than the 60-second lease TTL.
- Both built-in and custom-engine execution boundaries use the same scoped wrapper.
- Auto-mode trigger conditions, retry selection, dispatch claims, finalization, and loop termination branches are unchanged.
- The timer starts only while a worker-backed unit boundary is in flight and is cleared on both resolution and rejection.
- Refresh misses still clear the session fencing token; the database settlement trigger remains unchanged and fail-closed.
- Real-DB success/cleanup regression, rejection cleanup regression, two targeted sabotage runs, 139 changed-source tests, independent reviews, full hosted build, Node 22 smoke, and CI gate all pass.

PR #677, opened 2026-06-11
fix(issue): [Bug]: TUI streaming renderer duplicates the prose line before a code fence (mid-buffer reflow misclassified as tail append; scrollback-clamp re-emits boundary line)

jeremymcs:
## Summary
- Fixed streaming TUI duplicate-line bug by falling back to `fullRender(true)` when a mid-buffer insertion reaches into committed scrollback (`firstChanged < previousContentViewportTop && appendedLines`), and added a passing regression test that confirms the fix path is taken.

## Verification
- Completed in the repository worktree before push.

## Related Issue
- Closes #592
- [#592 [Bug]: TUI streaming renderer duplicates the prose line before a code fence (mid-buffer reflow misclassified as tail append; scrollback-clamp re-emits boundary line)](https://github.com/open-gsd/gsd-pi/issues/592)

## User
- @diakula

## Repo
- `open-gsd/gsd-pi`

## Branch
- `issue/592-bug-tui-streaming-renderer-duplicates-th-1781184541`

---

> [!NOTE]
> **Low Risk**
> Localized TUI redraw logic change with a targeted regression test; no auth, data, or API surface impact.
> 
> **Overview**
> Fixes **#592** duplicate-line glitches during streaming when markdown reflow inserts lines inside committed scrollback while the buffer grows (e.g. a code-fence border appearing above existing prose).
> 
> In `packages/pi-tui/src/tui.ts`, differential redraw now **bails out to `fullRender(true)`** when `firstChanged < previousContentViewportTop` **and** `appendedLines` is true, instead of the scrollback-clamp path that could leave a boundary line in scrollback and redraw it in the live region.
> 
> Adds a **regression test** in `tui-render.test.ts` for mid-buffer insertion in a short viewport (9→10 lines, insert at index 2).
> 
> Also includes **`pnpm-lock.yaml`** / **`web/package.json`** housekeeping (postcss devDep range, lockfile resolution churn).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4feb6854c340ccf704543677dd62acffed71433c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

---
[...]

PR #1487, opened 2026-07-17
fix(issue): fix(pi-tui): #677's scrollback-clamp guard only covers appendedLines=true; same-length reflow still duplicates a boundary line

### LINKED ISSUE #1486: fix(pi-tui): #677's scrollback-clamp guard only covers appendedLines=true; same-length reflow still duplicates a boundary line
## Problem

PR #677 (closing #592) fixed TUI streaming duplicate-line bugs, but only for one of two symmetric cases. The guard it added is gated on `appendedLines === true`; the sibling case — a reflow that shifts a line without changing the total line count — has no guard and reproduces the identical verbatim-duplicate symptom.

## Root Cause

`packages/pi-tui/src/tui.ts`, in `doRender()`:

```ts
if (firstChanged < previousContentViewportTop) {
    if (appendedLines) {
        // #677's fix — only this branch is guarded
        logRedraw(`firstChanged < viewportTop + buffer grew (...) — full repaint to avoid duplicate`);
        fullRender(true);
        return;
    }
    // Unguarded sibling, unchanged since before #592:
    const newViewportTop = getViewportTop(newLines.length);
    const clampedFirst = Math.max(0, Math.min(previousContentViewportTop, newViewportTop));
    logRedraw(`firstChanged < viewportTop (...) — repaint from ${clampedFirst}`);
    firstChanged = clampedFirst;
    lastChanged = Math.max(lastChanged, newLines.length - 1);
    clampedToViewport = true;
}
```

`#592`'s original report and `#677`'s fix framed the trigger as a markdown code-fence border *inserting* a line mid-buffer (buffer grows → `appendedLines: true`). The fix and its regression test (`tui-render.test.ts`, "mid-buffer insertion... reaching into committed scrollback") only exercise that path.

But the same duplicate-emission mechanism — a changed line landing above `previousContentViewportTop` (already flushed to real terminal scrollback, unreachable by cursor movement) — is not exclusive to insertions. `Text.render()` (`packages/pi-tui/dist/components/text.js:47`, via `wrapTextWithAnsi`) re-wraps the entire cached message from scratch on every streamed token. Plain prose (no code fence) can hit a word-wrap boundary shift where content moves between lines without changing the total line count — `appendedLines` stays `false`. When that shift lands on a line index already above the viewport, the code falls into the unguarded `else` branch: it clamps `firstChanged` up to the viewport top and repaints downward, but the line at its *old* row is still sitting in already-committed scrollback from the previous frame — producing the same verbatim duplicate #677 was meant to eliminate, just triggered by a same-length reflow instead of a growing one.

## Expected Behavior

Any reflow whose earliest change lands above `previousContentViewportTop` should trigger a clean repaint, regardless of whether `appendedLines` is `true` or `false` — the displaced-line risk is identical in both cases.

## Concrete Fix

Drop the `if (appendedLines)` condition gating the guard, so `fullRender(true)` fires whenever `firstChanged < previousContentViewportTop`:

```diff
 if (firstChanged < previousContentViewportTop) {
-    if (appendedLines) {
-        logRedraw(`firstChanged < viewportTop + buffer grew (${firstChanged} < ${previousContentViewportTop}) — full repaint to avoid duplicate`);
-        fullRender(true);
-        return;
-    }
-    const newViewportTop = getViewportTop(newLines.length);
-    const clampedFirst = Math.max(0, Math.min(previousContentViewportTop, newViewportTop));
-    logRedraw(`firstChanged < viewportTop (${firstChanged} < ${previousContentViewportTop}) — repaint from ${clampedFirst}`);
-    firstChanged = clampedFirst;
-    lastChanged = Math.max(lastChanged, newLines.length - 1);
-    clampedToViewport = true;
+    logRedraw(`firstChanged < viewportTop (${firstChanged} < ${previousContentViewportTop}) — full repaint to avoid duplicate`);
+    fullRender(true);
+    return;
 }
```

If a full repaint on every such reflow proves too disruptive for the shrink-without-growth case specifically, the clamp-and-repaint path could instead be made scrollback-aware (detect that the pre-clamp `firstChanged` row is already off-screen and skip re-emitting it) — but that's more invasive than reusing the existing `fullRender(true)` escape hatch already proven safe by #677.

## Test

Regression scenario mirroring `tui-render.test.ts`'s existing `#677` test, but with `appendedLines: false`: render N lines in a short viewport so some lines are already committed to scrollback (`previousContentViewportTop > 0`), then re-render the SAME line count with a change at an index below `previousContentViewportTop` (a same-length reflow, not an insertion). Assert `fullRedraws` increases and the changed line does not appear twice across scrollback + viewport.

## Environment

- GSD (gsd-pi): 1.11.0 (`@opengsd/gsd-pi`)
- Component: `packages/pi-tui` (`src/tui.ts` / compiled `dist/tui.js` ~line 1213-1230)
- AI runtime: Claude Code 2.1.198, model claude-sonnet-5

## Reproduction Context

Observed live during an interactive chat session: a wrapped prose paragraph's last line was printed twice on-screen, with no second `╭─ GSD` header (ruling out the tool-interleaved-turn bubble-duplication bugs like #393/#654) and no code fence involved (ruling out #592's exact trigger). The duplicated content was verified to appear only once in the actual message/session content — confirming it as a pure render artifact in the same family as #592, not a model or adapter issue.

## Duplicate Check

Searched `open-gsd/gsd-pi` issues for `appendedLines`, `mid-buffer reflow`, `scrollback clamp`, `verbatim duplicate`, `clampedToViewport`, `reflow same-length duplicate` — only #592 (closed, fixed by #677) and #1152 (a different bug: orphaned tool components re-added without dedup at message-end rebuild) come up. Neither describes this specific gap: #677's guard was scoped only to the `appendedLines: true` half of the `firstChanged < previousContentViewportTop` branch.

---
*Filed following a live investigation during an interactive session; traced against the installed 1.11.0 `dist/tui.js` and the `#677` diff/test on GitHub.*

jeremymcs:
## Summary
- Fixed same-length TUI scrollback reflows to force full repaint and added regression coverage; whitespace verification passed while focused tests were blocked by missing dependencies/network.

## Verification

## Related Issue
- Closes #1486
- [#1486 fix(pi-tui): #677's scrollback-clamp guard only covers appendedLines=true; same-length reflow still duplicates a boundary line](https://github.com/open-gsd/gsd-pi/issues/1486)

## User
- @diakula

## Repo
- `open-gsd/gsd-pi`

## Branch
- `issue/1486-fix-pi-tui-677-s-scrollback-clamp-guard--1784304230`

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-pi/pr/1487"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a>

---

> [!NOTE]
> **Low Risk**
> Localized change to pi-tui differential redraw heuristics with a targeted regression test; no auth, data, or API surface impact.
> 
> **Overview**
> Fixes **duplicate boundary lines** when markdown or word-wrap reflows move content across the scrollback/viewport edge **without** growing the buffer—behavior that previously only triggered a full repaint when `appendedLines` was true.
> 
> When `firstChanged` is above committed scrollback (`firstChanged < previousContentViewportTop`), differential rendering now **always** calls `fullRender(true)` instead of clamping the repaint start to the viewport. The old clamp path could leave a line frozen in scrollback while repainting it in the live region.
> 
> Related cleanup removes the `clampedToViewport` branch for ghost-line clearing and `maxLinesRendered` adjustment. A regression test covers same-length reflow that moves a boundary line from scrollback into the viewport.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6d90e9e002aaea71b9ec8c61354313ff80052cab. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/pi-tui/src/tui.ts
+++ b/packages/pi-tui/src/tui.ts
@@ -1485,27 +1485,16 @@ export class TUI extends Container {
 		}
 
 		const previousContentViewportTop = getViewportTop(this.previousLines.length);
-		let clampedToViewport = false;
 		if (firstChanged < previousContentViewportTop) {
-			if (appendedLines) {
-				// A mid-buffer insertion (e.g. a markdown code-fence border materialising)
-				// shifted a line across the scrollback/viewport boundary.  Clamping would
-				// leave the displaced line frozen in scrollback AND re-emit it in the live
-				// region, producing a verbatim duplicate.  Fall back to a clean repaint.
-				logRedraw(
-					`firstChanged < viewportTop + buffer grew (${firstChanged} < ${previousContentViewportTop}) — full repaint to avoid duplicate`,
-				);
-				fullRender(true);
-				return;
-			}
-			const newViewportTop = getViewportTop(newLines.length);
-			const clampedFirst = Math.max(0, Math.min(previousContentViewportTop, newViewportTop));
+			// A mid-buffer reflow (for example markdown code-fence borders or prose word-wrap)
+			// shifted a line across the scrollback/viewport boundary. Clamping would leave
+			// the displaced line frozen in scrollback and re-emit it in the live region,
+			// producing a verbatim duplicate. Fall back to a clean repaint.
 			logRedraw(
-				`firstChanged < viewportTop (${firstChanged} < ${previousContentViewportTop}) — repaint from ${clampedFirst}`,
+				`firstChanged < viewportTop (${firstChanged} < ${previousContentViewportTop}) — full repaint to avoid duplicate`,
 			);
-			firstChanged = clampedFirst;
-			lastChanged = Math.max(lastChanged, newLines.length - 1);
-			clampedToViewport = true;
+			fullRender(true);
+			return;
 		}
 
 		let buffer = this.useSynchronizedOutput ? "\x1b[?2026h" : "";
@@ -1560,7 +1549,7 @@ export class TUI extends Container {
 		// Track where cursor ended up after rendering
 		let finalCursorRow = renderEnd;
 
-		if (this.previousLines.length > newLines.length && !clampedToViewport) {
+		if (this.previousLines.length > newLines.length) {
 			const renderEndScreenRow = renderEnd - viewportTop;
 			const ghostLinesVisible = renderEndScreenRow < height - 1;
 			if (ghostLinesVisible) {
@@ -1617,11 +1606,7 @@ export class TUI extends Container {
 		// hardwareCursorRow tracks actual terminal cursor position (for movement)
 		this.cursorRow = Math.max(0, newLines.length - 1);
 		this.hardwareCursorRow = finalCursorRow;
-		if (clampedToViewport && newLines.length < this.maxLinesRendered) {
-			this.maxLinesRendered = newLines.length;
-		} else {
-			this.maxLinesRendered = Math.max(this.maxLinesRendered, newLines.length);
-		}
+		this.maxLinesRendered = Math.max(this.maxLinesRendered, newLines.length);
 		this.previousViewportTop = getViewportTop(this.maxLinesRendered);
 
 		// Position hardware cursor for IME
--- a/packages/pi-tui/test/tui-render.test.ts
+++ b/packages/pi-tui/test/tui-render.test.ts
@@ -355,6 +355,63 @@ describe("TUI mid-buffer reflow", () => {
 
 		tui.stop();
 	});
+
+	it("triggers full repaint when a same-length reflow shifts a change into committed scrollback", async () => {
+		// 9 lines in a height-5 terminal leaves indices 0-3 in committed scrollback.
+		const terminal = new VirtualTerminal(40, 5);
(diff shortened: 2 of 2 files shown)


jeremymcs:

Re-checked PR #1487 against [#1486 fix(pi-tui): #677's scrollback-clamp guard only covers appendedLines=true; same-length reflow still duplicates a boundary line](https://github.com/open-gsd/gsd-pi/issues/1486).

## Subtasks
- [x] **fix(pi-tui): #677's scrollback-clamp guard only covers appendedLines=true; same-length reflow still duplicates a boundar**

PR #1597, opened 2026-08-04
fix(issue): task-recovery-abort loop after verify timeout: gsd_task_recovery_resume requires a recoveryActionId that is never surfaced anywhere

### LINKED ISSUE #1593: task-recovery-abort loop after verify timeout: gsd_task_recovery_resume requires a recoveryActionId that is never surfaced anywhere
### Summary

When a per-task verify command exceeds `DEFAULT_COMMAND_TIMEOUT_MS` (120s) enough times to exhaust the `verification-failed` retry budget, auto-mode mints a terminal, agent-owned `abort` recovery action and breaks the session. Every subsequent `/gsd auto` dispatch then breaks in ~100ms with `task-recovery-abort` **before any agent turn starts** — indefinitely.

The engine does provide a forward path: `gsd_task_recovery_resume`, whose guard (`requireResumableAbortScope`) the stuck state fully satisfies. But the tool requires the exact `recoveryActionId`, and that ID is never surfaced on any operator-facing surface — not in the break message, journal, status, doctor, runtime unit files, or the dispatch prompt. The result presents as a permanent deadlock even though the state machine is resumable by design.

**Version:** `@opengsd/gsd-pi` 1.12.0. File:line references below are into the installed dist under `extensions/gsd/`.

### Failure chain

1. `constants.js:8` — `DEFAULT_COMMAND_TIMEOUT_MS = 120_000`.
2. `verification-gate.js:566-573` — a timed-out `spawnSync` is recorded as **`exitCode = 127`** (indistinguishable from "command not found"); VERIFY.json gets `passed: false`, `durationMs ≈ 120001`. Any verify chain slower than 120s always trips this, even when the underlying checks pass when run manually.
3. `recovery-policy.js:39-41,70-72` — the `verification-failed` budget (`maxUses: 2`) exhausts → `terminalDecision("abort")`, agent-owned.
4. `auto-verification.js:844-849` → `auto/finalize.js:181-184` — the session breaks with `verification-abort`; the task lifecycle is left `in_progress` while the settled attempt carries the terminal abort.
5. `auto/task-execution-cutover.js:240-248` — on every later dispatch, `runWithTaskExecutionAttempt` reads the settled predecessor, sees the agent-owned abort with `!resumeAuthorized`, and returns `{action: "break", reason: "task-recovery-abort"}` at line 247 — **before `claimTaskAttempt` at line 263**. No attempt is claimed, no executor runs, the dispatch exits in ~100ms. This repeats on every dispatch until the abort is resumed.

### The actual defect: the resume key is unobtainable

The resume path is externally callable and correctly guarded:

- `bootstrap/db-tools.js:1268-1289` registers `gsd_task_recovery_resume`; `resumeTaskRecovery` (`task-recovery-domain-operation.js:449-475`) is guarded only by `requireResumableAbortScope(recoveryActionId)` — a pure SQL predicate over DB state (latest attempt, current agent-owned abort, no open blocker, current fail verdict, not already resumed). No session/ownership condition; the stuck state satisfies every clause.

But the required `recoveryActionId` is hidden from every surface:

- It is minted with `randomUUID()` (`db/writers/task-recovery.js:238`) — not derivable.
- The by-task recovery reader **deliberately returns `null` for an abort that has not been resumed yet** (`task-recovery-domain-operation.js:161-163`), so the dispatch-prompt block that would print `**Recovery action:** <id>` (`auto-prompts.js:154`) never renders for exactly the state that needs it.
- The cutover break at `task-execution-cutover.js:247` fires even though `readTaskRecoveryRoute` (`task-recovery-domain-operation.js:476-506`) has just handed the cutover the `recoveryActionId` ungated — the engine is holding the ID at the exact moment it emits the opaque break, and discards it.
- It appears in no journal event, no runtime unit file (the break happens before the unit file is written), and no status/doctor output — doctor reports 0 issues for a project wedged in this state.

The tool's own guideline says "Use the exact recoveryActionId returned by the current abort" — but the current abort never returns it. In practice, recovery required a manual read-only SQL query against a copy of `gsd.db`, which the docs (correctly) forbid as a normal operation.

Notably, in the observed occurrence the task's deliverable was complete and every dispatch had logged `artifactVerified: true` — the wedge is purely recovery bookkeeping, not unfinished work, which makes the silent, unrecoverable-looking break especially costly.

### Repro

1. Plan a task whose `verify` command reliably takes > 120s (e.g. a browser UAT plus a full test suite).
2. Run `/gsd auto`; let the verify time out until the retry budget (2) is exhausted.
3. The session breaks with `verification-abort`; the task stays `in_progress`.
4. Re-run `/gsd auto`: instant `task-recovery-abort` break, no agent turn — every time.
5. Attempt recovery: `gsd_task_recovery_resume` cannot be called without the ID; `gsd_task_reopen` refuses (task is not `complete`); `gsd_doctor` reports nothing wrong.

### Suggested fix

**Primary (minimal):** include `recoveryActionId` in the `task-recovery-abort` break payload / journal event at `task-execution-cutover.js:246-247` — the `readTaskRecoveryRoute` result already carries it — and/or in the `verification-abort` finalize message when the abort is minted.

**Secondary hardening:**
- Let the by-task reader (`task-recovery-domain-operation.js:161-163`) return un-resumed aborts so the durable-recovery prompt section renders and shows the ID.
- Treat a terminal verify-stage abort as a surfaced paused/blocked state (as the validate-milestone path already does at `auto-verification.js:376-406`) instead of a silent break loop.
- In `verification-gate.js`, distinguish a `spawnSync` timeout from exit 127 so a slow verify is reported as "timed out after 120s" rather than "command not found" — the misleading exit code sent diagnosis in the wrong direction.

jeremymcs:
## Summary
- Surfaced the abort's recoveryActionId in the task-recovery-abort break reason so it reaches the journal/ledger, verified with typecheck and 216 passing recovery/cutover/loop tests.

## Verification

## Related Issue
- Closes #1593
- [#1593 task-recovery-abort loop after verify timeout: gsd_task_recovery_resume requires a recoveryActionId that is never surfaced anywhere](https://github.com/open-gsd/gsd-pi/issues/1593)

## User
- @jcisc0

## Repo
- `open-gsd/gsd-pi`

## Branch
- `issue/1593-task-recovery-abort-loop-after-verify-ti-1785811096`

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-pi/pr/1597"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a>

### DIFF
--- a/src/resources/extensions/gsd/auto-verification.ts
+++ b/src/resources/extensions/gsd/auto-verification.ts
@@ -239,6 +239,10 @@ function routeHostTechnicalFailure(
   attempt: VerificationAttemptSnapshot,
   verdict: FailedVerdictIdentity,
   failureKind: "verification-failed" | "verification-drift" = "verification-failed",
+  // A terminal abort is resumable by design via `gsd_task_recovery_resume`, but that
+  // tool needs the exact recoveryActionId. Hand it back so the caller can surface it
+  // on the finalize break reason instead of discarding it here (#1593).
+  recordAbort?: (recoveryActionId: string) => void,
 ): "retry" | "abort" {
   if (!attempt.resultId) throw new Error("Host verification Attempt Result is missing");
   const routeInput = {
@@ -273,8 +277,11 @@ function routeHostTechnicalFailure(
     case "remediate":
     case "replan":
       return "retry";
-    case "abort":
-      return recovery.status === "replayed" && recovery.resumeAuthorized ? "retry" : "abort";
+    case "abort": {
+      if (recovery.status === "replayed" && recovery.resumeAuthorized) return "retry";
+      recordAbort?.(recovery.recoveryActionId);
+      return "abort";
+    }
     default:
       throw new Error(`Unsupported agent recovery action ${recovery.action}`);
   }
@@ -639,6 +646,14 @@ export async function runPostUnitVerification(
     return "continue";
   }
 
+  // Capture the recoveryActionId of a terminal abort so the finalize break reason can
+  // print it — it is the only key `gsd_task_recovery_resume` accepts (#1593). Reset per
+  // verification pass so a stale id from an earlier unit can never be reported.
+  s.lastTaskRecoveryAbortId = null;
+  const recordAbort = (recoveryActionId: string) => {
+    s.lastTaskRecoveryAbortId = recoveryActionId;
+  };
+
   let recoverableAttempt: VerificationAttemptSnapshot | null = null;
   let recoverableAuthority: TaskVerificationAuthority | null = null;
   let canonicalVerdictWriteStarted = false;
@@ -664,7 +679,7 @@ export async function runPostUnitVerification(
         verdictId: replayedVerdict.verdictId,
         evidenceId: replayedVerdict.evidenceId,
         verdict: replayedVerdict.verdict,
-      }, replayedVerdict.supersedesVerdictId ? "verification-drift" : "verification-failed")
+      }, replayedVerdict.supersedesVerdictId ? "verification-drift" : "verification-failed", recordAbort)
       : null;
     if (!replayedRecovery && !isTaskAttemptAwaitingVerification(latestAttempt)) {
       throw new Error("Host verification requires the latest succeeded canonical Attempt at the verify stage");
@@ -743,7 +758,7 @@ export async function runPostUnitVerification(
         verdictId: invalidated.verdictId,
         evidenceId: invalidated.evidenceId,
         verdict: "inconclusive",
-      }, "verification-drift");
+      }, "verification-drift", recordAbort);
       if (recovery === "abort") return "abort";
       return recordDurableVerificationRetry(s, retryKey, failureContext);
     }
@@ -1106,7 +1121,7 @@ export async function runPostUnitVerification(
           verdictId: recordedVerdict.verdictId,
           evidenceId: recordedVerdict.evidenceId,
           verdict: hostTechnicalVerdict,
-        });
+        }, "verification-failed", recordAbort);
       }
 
       try {
@@ -1243,7 +1258,7 @@ export async function runPostUnitVerification(
         verdictId: storedVerdict.verdictId,
         evidenceId: storedVerdict.evidenceId,
         verdict: storedVerdict.verdict,
-      }, storedVerdict.supersedesVerdictId ? "verification-drift" : "verification-failed");
+      }, storedVerdict.supersedesVerdictId ? "verification-drift" : "verification-failed", recordAbort);
       if (recovery === "abort") return "abort";
       const retryKey = verificationRetryKey(s.currentUnit.type, s.currentUnit.id);
       return recordDurableVerificationRetry(s, retryKey, message);
@@ -1271,7 +1286,7 @@ export async function runPostUnitVerification(
       verdictId: recorded.verdictId,
       evidenceId: recorded.evidenceId,
       verdict: "inconclusive",
-    });
+    }, "verification-failed", recordAbort);
     if (recovery === "abort") return "abort";
     const retryKey = verificationRetryKey(s.currentUnit.type, s.currentUnit.id);
     return recordDurableVerificationRetry(s, retryKey, message);
--- a/src/resources/extensions/gsd/auto/finalize.ts
+++ b/src/resources/extensions/gsd/auto/finalize.ts
@@ -248,9 +248,17 @@ export async function runFinalize(
     );
 
     if (verificationResult === "abort") {
-      debugLog("autoLoop", { phase: "exit", reason: "verification-abort" });
+      // The abort is terminal but resumable via `gsd_task_recovery_resume`, which needs
+      // the exact recoveryActionId. Carry it in the break reason so it reaches the
+      // journal, the dispatch ledger, and the operator rather than being discarded —
+      // otherwise every later dispatch loops on an opaque `task-recovery-abort` (#1593).
+      const abortId = s.lastTaskRecoveryAbortId;
+      const abortReason = abortId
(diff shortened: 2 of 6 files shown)


jeremymcs:
🔎 **Verification — 0 of 1 addressed**

Re-checked PR #1597 against [#1593 task-recovery-abort loop after verify timeout: gsd_task_recovery_resume requires a recoveryActionId that is never surfaced anywhere](https://github.com/open-gsd/gsd-pi/issues/1593).

## Subtasks
- [ ] **task-recovery-abort loop after verify timeout: gsd_task_recovery_resume requires a recoveryActionId that is never surfac** _(pending: Diff only appends recoveryActionId to the cutover break reason string. Never surfaces the ID on the finalize/verification-abort path, journal event, by-task recovery reader, status/doctor, or prompt; )_

jeremymcs:
**Needs attention** — automatic fix failed: Command timed out after 900000ms.
**Resolved** — addressed in commit `fc59726`.

jeremymcs:

jeremymcs:
## Code-owner review of the outstanding feedback

Commit `fc597264` pushed. No inline review threads exist on this PR (Copilot reviewed both files and generated no comments), so replies are posted here.

### 1. Verification comment — *"Diff only appends recoveryActionId to the cutover break reason string"* — **partially accepted, fixed**

Valid on its main point. The previous head only patched `task-execution-cutover.ts`, which is the break emitted on **later** dispatches. The abort is minted one step earlier, on the finalize path, and that first break — the one an operator actually sees when the session wedges — still printed a bare `verification-abort` with no way to obtain the id.

Fixed in `fc597264`:

- `auto-verification.ts` — `routeHostTechnicalFailure` now hands `recovery.recoveryActionId` back via a `recordAbort` callback for a **terminal** (non-replayed, non-authorized) abort. Wired at all 5 call sites inside `runPostUnitVerification`.
- `auto/session.ts` — new `AutoSession.lastTaskRecoveryAbortId`, reset in `resetSessionState()` and at the top of each `execute-task` verification pass so a stale id from an earlier unit can never be reported.
- `auto/finalize.ts` — the abort break reason becomes `verification-abort (recoveryActionId: <uuid>; resume with gsd_task_recovery_resume)`.

**On the journal/ledger sub-claim:** this is now covered by construction on both paths. `auto/loop.ts:1602-1616` takes `unitPhaseResult.reason` and feeds it to `settleDispatchFailed(dispatchId, breakReason, …)`, `finishIncompleteIteration({reason: breakReason, …})`, and `finishTurn("stopped", "execution", breakReason)` — so the reason string *is* the journal event and the dispatch-ledger record. There is no separate field to populate.

**Rejected sub-items, with reasons:**

- *"by-task recovery reader"* (`task-recovery-domain-operation.ts:161-163` returning `null` for un-resumed aborts) and *"prompt"* (`auto-prompts.ts:201`) — changing the reader does not help. The dispatch-prompt block is only reachable after `runWithTaskExecutionAttempt` claims an attempt, and the wedge breaks at `task-execution-cutover.ts` *before* `claimTaskAttempt`, so that section never renders in exactly this state. Rendering it would also require widening the prompt's action union (`retry | repair | remediate | replan | resume`) and its `executionContract` switch to admit `abort`, which is a behavioral change to the recovery contract, not a surfacing fix. Out of scope here.
- *"status/doctor"* — a genuine gap, but adding a new doctor check is a separate feature, not a fix to this break path. The id now reaches the operator on both the mint break and every subsequent dispatch break, which unblocks the reported deadlock.

### 2. PR Risk Report (🔴 Auto Engine — *"validate auto-mode trigger conditions and loop termination"*) — **verified, no change needed**

Checked as requested. Loop termination is unchanged by this PR:

- Both the finalize abort branch and the cutover abort branch still return `{ action: "break", … }`; only the `reason` **string content** changed. `auto/loop.ts:1602` branches on `unitPhaseResult.action === "break"`, never on the reason text.
- I grepped for literal comparisons against `"task-recovery-abort"` / `"verification-abort"` across `src/`: the only matches outside the emitters are two test fixtures in `auto-loop.test.ts` that *produce* the string, none that consume it. No production code pattern-matches these reasons, so the added suffix cannot alter routing.
- The one new piece of state, `lastTaskRecoveryAbortId`, is write-only outside `runFinalize` and is cleared both in `resetSessionState()` and at the start of every `execute-task` verification pass — it cannot influence trigger conditions.

### Verification run

- `pnpm run typecheck:extensions` — clean (after `build:contracts`, `build:pi`, `build:mcp-server`; the pre-existing `TS2307` noise was unbuilt workspace packages, not this change).
- 348 tests pass across `auto-verification`, `auto-task-execution-cutover`, `auto-loop`, `loop`, `post-unit-task-recovery-authority`, `auto-orchestrator`, `auto-phases-lifecycle`, `recovery-finalize-logs`, `finalize-timeout-guard`, `finalize-survivor-branch`, `artifact-retry-cap`, `post-exec-retry-bypass`, `journal-integration`, `validate-milestone-stuck-guard`, `auto-paused-ui-cleanup`.
- Two new focused tests in `auto-verification.test.ts`: a terminal abort surfaces its `recoveryActionId`; a replayed authorized abort routes to `retry` and surfaces nothing.

jeremymcs:
Addressed in commit `fc59726`.

> Responding to [comment by @jeremymcs](https://github.com/open-gsd/gsd-pi/pull/1597#issuecomment-5174022588):
> 🔎 **Verification — 0 of 1 addressed**

PR #1620, opened 2026-08-07
fix(issue): roadmap-divergence reconciliation never converges when a slice is skipped (detector includes skipped slices, renderer omits them)

### LINKED ISSUE #1619: roadmap-divergence reconciliation never converges when a slice is skipped (detector includes skipped slices, renderer omits them)
## Problem

Auto-mode is permanently stuck on a milestone whose plan contains a **skipped slice**. `reconcileBeforeDispatch` throws `Reconciliation drift persisted after cap=2 passes: roadmap-divergence` before any unit can run, because the ROADMAP renderer and the roadmap-divergence detector disagree about whether a `skipped` slice belongs in the projection. The drift can never be repaired, so every dispatch re-throws.

## Root Cause

The ROADMAP projection **excludes** skipped slices, but the divergence detector **includes** them and treats them as "ready", so the detector permanently reports a divergence that the repair (a re-render) can never resolve.

- Renderer omits skipped slices — `markdown-renderer.js:467`:
  ```js
  const slices = getMilestoneSlices(milestoneId).filter((slice) => slice.status !== "skipped");
  ```
  So the rendered/parsed roadmap contains only the non-skipped slices.

- Detector reads the **unfiltered** DB slice list — `state-reconciliation/drift/roadmap.js:41`:
  ```js
  const dbSlices = getMilestoneSlices(milestoneId);
  ```
  and marks a slice "ready" when `isClosedStatus(slice.status)` is true — `roadmap.js:24`:
  ```js
  if (isClosedStatus(slice.status) || getSliceTasks(milestoneId, slice.id).length > 0) {
      ready.add(slice.id);
  }
  ```
  Because `skipped` ∈ `RAW_CLOSED_STATUSES` (`status-guards.js:34`), `isClosedStatus("skipped") === true`, so the skipped slice is "ready".

- The absence check then fires permanently — `roadmap.js:64-69`:
  ```js
  for (const dbSlice of dbSlices) {
      if (!readySliceIds.has(dbSlice.id)) continue;      // skipped slice IS ready
      if (!roadmapSliceIds.has(dbSlice.id)) return true; // ...but renderer omitted it → divergence
  }
  ```
  (When the skipped slice is not last, the order check at `roadmap.js:57` also fires, since `dbSliceOrder` counts the skipped row but the parsed roadmap does not — same permanent-divergence class.)

`repairRoadmapDivergence` (`roadmap.js:96`) just calls `renderRoadmapFromDb` again, which omits the skipped slice again, so the next detect pass returns the same drift. Two passes (`state-reconciliation/index.js`, `MAX_PASSES=2`) plus the final verify never converge → `ReconciliationFailedError({ persistentDrift })` → `state-reconciliation/errors.js:38`.

## Expected Behavior

The detector's DB slice view must match what the renderer actually projects. Mirror the renderer's filter at `roadmap.js:41`:

```js
// roadmap.js:41 — exclude skipped slices, matching renderRoadmapFromDb (markdown-renderer.js:467)
const dbSlices = getMilestoneSlices(milestoneId).filter((s) => s.status !== "skipped");
```

This aligns `dbSliceOrder`, `readySliceIds`, and the absence check with the projection. For the reproducing milestone it collapses to: dbSlices = the non-skipped slices, all pending with no tasks → `readySliceIds` empty → early return `false` at `roadmap.js:45` → no false divergence.

## Environment
- GSD version: 1.12.0
- Model: claude-opus-4-8 (plan-milestone)
- Unit: plan-milestone/M017

## Reproduction Context

During `plan-milestone` for the active milestone M017, the planner created a blocker slice `S00-blocker` and skipped it. Live DB state:

```
M017 (queued): S00-blocker=skipped, S01..S05=pending
```

From that point, `reconcileBeforeDispatch` (runs before every dispatch decision) throws on the roadmap-divergence detector, blocking all forward progress on the milestone.

## Forensic Evidence

- Top-line error: `Reconciliation drift persisted after cap=2 passes: roadmap-divergence`
- stuck-loop: `plan-milestone/M017` dispatched 3 additional times at $0.00 each (unit re-returned after the reconciler gate throws).
- error-trace: `8 error(s) in plan/milestone-M017`; milestone lease kept renewing while the loop retried; recent reasoning traces show repeated "wait for the lease to expire and retry".
- Downstream: 338 worktree orphans / unmerged M017 branch — a consequence of auto-mode never clearing this gate, not a separate defect.

### Suggested recovery (for affected users)
- Un-skip / re-plan so the blocker slice no longer sits in `skipped` (e.g. `gsd_slice_reopen`, or replan so the blocker slice is removed rather than skipped). Removing the skipped row makes render and detect agree and the reconciler passes.

---

jeremymcs:
## Summary
- Filtered skipped slices out of the roadmap-divergence detector to match the renderer's projection, with a regression test verified to fail without the fix.

## Verification

## Related Issue
- Closes #1619
- [#1619 roadmap-divergence reconciliation never converges when a slice is skipped (detector includes skipped slices, renderer omits them)](https://github.com/open-gsd/gsd-pi/issues/1619)

## User
- @afkayla

## Repo
- `open-gsd/gsd-pi`

## Branch
- `issue/1619-roadmap-divergence-reconciliation-never--1786063701`

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-pi/pr/1620"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a>

### DIFF
--- a/src/resources/extensions/gsd/state-reconciliation/drift/roadmap.ts
+++ b/src/resources/extensions/gsd/state-reconciliation/drift/roadmap.ts
@@ -58,7 +58,13 @@ function milestoneHasDivergence(
     return false;
   }
 
-  const dbSlices = getMilestoneSlices(milestoneId);
+  // Mirror renderRoadmapFromDb's filter (#1619): the ROADMAP projection omits
+  // skipped slices, so the detector must compare against the same view.
+  // Without this, a skipped slice counts as "ready" (skipped ∈
+  // RAW_CLOSED_STATUSES) but never appears in the rendered markdown, so the
+  // absence/order checks below report a divergence that re-rendering can never
+  // repair — reconcileBeforeDispatch then throws on every dispatch.
+  const dbSlices = getMilestoneSlices(milestoneId).filter((s) => s.status !== "skipped");
   const dbSliceMap = new Map(dbSlices.map((s) => [s.id, s]));
   const dbSliceOrder = new Map(dbSlices.map((s, index) => [s.id, index]));
   const readySliceIds = getSlicesReadyForDivergenceCheck(milestoneId, dbSlices);
--- a/src/resources/extensions/gsd/tests/state-reconciliation-drift.test.ts
+++ b/src/resources/extensions/gsd/tests/state-reconciliation-drift.test.ts
@@ -1384,6 +1384,61 @@ test("ADR-017 (#870): roadmap-divergence accepts recovered S00 blocker sequence"
   assert.equal(readFileSync(roadmapPath, "utf-8"), originalRoadmap);
 });
 
+test("ADR-017 (#1619): skipped slices are excluded from the divergence view", async (t) => {
+  const base = mkdtempSync(join(tmpdir(), "gsd-adr017-roadmap-skipped-"));
+  const milestoneId = "M017";
+  const milestoneDir = join(base, ".gsd", "milestones", milestoneId);
+  const roadmapPath = join(milestoneDir, `${milestoneId}-ROADMAP.md`);
+  mkdirSync(milestoneDir, { recursive: true });
+  // The projection omits the skipped S00-blocker, so S01 renders first.
+  const originalRoadmap = [
+    `# ${milestoneId}: Blocker Skipped Milestone`,
+    "",
+    "**Vision:** Skipped blocker slice must not wedge reconciliation.",
+    "",
+    "## Slices",
+    "",
+    "- [ ] **S01: Source of truth contract** `risk:medium` `depends:[]`",
+    "  > After this: ",
+    "",
+    "- [ ] **S02: Policy implementation** `risk:medium` `depends:[]`",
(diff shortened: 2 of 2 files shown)


jeremymcs:

Re-checked PR #1620 against [#1619 roadmap-divergence reconciliation never converges when a slice is skipped (detector includes skipped slices, renderer omits them)](https://github.com/open-gsd/gsd-pi/issues/1619).

## Subtasks
- [x] **roadmap-divergence reconciliation never converges when a slice is skipped (detector includes skipped slices, renderer om**

PR #1625, opened 2026-08-07
fix(issue): [Bug]: [drift-check] /gsd auto enters a two-state drift loop: disk-slice-id-divergence fix triggers roadmap-divergence

### LINKED ISSUE #1623: [Bug]: [drift-check] /gsd auto enters a two-state drift loop: disk-slice-id-divergence fix triggers roadmap-divergence
### Bug type

Crash or hang

## Summary

`/gsd auto` can enter a **two-state drift loop** between two reconciliation checks that are in tension with each other. Fixing the first drift (`disk-slice-id-divergence`) by adding skipped placeholder
 slices *introduces* the second drift (`roadmap-divergence`), which the auto-repair path cannot reconcile by design. The result is an unrecoverable pause loop that cannot be escaped via the normal `/gsd auto`
 workflow.

## Where

   Affected module: `state-reconciliation` drift checks, run during the reconciliation phase of `/gsd auto`.

   Relevant source files (gsd-pi):
   - `src/resources/extensions/gsd/state-reconciliation/drift/artifact-db.ts` — `detectDiskSliceIdDivergenceForMilestone`
   - `src/resources/extensions/gsd/state-reconciliation/drift/roadmap.ts` — `milestoneHasDivergence`, `getSlicesReadyForDivergenceCheck`
   - `src/resources/extensions/gsd/markdown-renderer.ts` — `renderRoadmapFromDb`
   - `src/resources/extensions/gsd/status-guards.ts` — `RAW_CLOSED_STATUSES`

### Expected behavior

normal work

### What happens
   The drift check `detectDiskSliceIdDivergenceForMilestone` scans a milestone's phase directory for files matching the regex:

 ```
 ^\d+-(\d+)-PLAN.md$
 ```

   For each match it derives a **bare** slice id, e.g. from `NN-01-PLAN.md` it extracts group `01` and builds `sliceId = "S01"` (zero-padded, no suffix). It then checks membership against the set of slice ids
 known to the DB for that milestone:

 ```
 knownSliceIds = new Set( milestoneIdAliases(...).flatMap(id => getMilestoneSlices(id).map(s => s.id)) )
 ```

   where `getMilestoneSlices` runs `SELECT id FROM slices WHERE milestone_id = :mid`.

   ### Why it drifts
   If the milestone contains a PLAN file whose derived bare slice id (`S01`) is **not** an exact id present in the DB (the DB may hold `S01-replan`, `S01-manual-review`, etc., but never the bare `S01`), the
 check flags `disk-slice-id-divergence` and `/gsd auto` pauses.

   The comparison is **exact id match, not prefix match**: a slice named `S01-manual-review` does **not** satisfy the lookup for `S01`.

   ### Workaround that was required
   Create the bare slice id (`S01`) in the DB via `gsd_reassess_roadmap` (the only planning entry point that accepts a bare slice id; `gsd_plan_slice` rejects it with "missing parent slice"), then mark it
 `skipped` so it requires no execution. Only then does `knownSliceIds.has("S01")` return true and the check passes.

   This workaround is fragile and non-obvious: there is no surfaced guidance that the drift check expects an **exact** bare id derived from the PLAN filename.

## Problem 2 — `roadmap-divergence` (side-effect of the Problem 1 workaround)

   The Problem 1 workaround (adding skipped placeholder slices) reliably *triggers a second, unrecoverable drift*.

   ### What happens
   `milestoneHasDivergence` compares the rendered `ROADMAP.md` against the DB. A slice is considered "ready" for the comparison if:

 ```
 isClosedStatus(slice.status)  OR  getSliceTasks(slice.id).length > 0
 ```

   The `status === "skipped"` value is listed in `RAW_CLOSED_STATUSES = ["complete","done","skipped","closed"]`, so `isClosedStatus("skipped") === true` → **skipped slices are "ready"** and participate in the
 comparison.

   However, `renderRoadmapFromDb` **explicitly filters out** slices with `status === "skipped"` — they are never rendered into `ROADMAP.md`.

   ### The loop
   For any skipped slice that is "ready":
   - it is present in the DB → comparison expects it in `ROADMAP.md`
   - it is filtered out of `ROADMAP.md` by design → it is absent from the roadmap
   - `milestoneHasDivergence` returns `true` (ready slice missing from roadmap)
   - auto-repair re-renders the roadmap via `renderRoadmapFromDb` → still filters the skipped slice → still missing → drift persists
   - after the cap of reconciliation passes, `/gsd auto` pauses with `Reconciliation drift persisted after cap=2 passes: roadmap-divergence`

   The skipped-by-design filter and the skipped-is-ready semantics are **mutually incompatible**, so the auto-repair path can never converge.

   ### Workaround that was required
   Reopen the skipped placeholder slices (`gsd_slice_reopen`) so their status becomes `in_progress`. Then:
   - `isClosedStatus("in_progress") === false` and `getSliceTasks` returns 0 → the slice is **not "ready"** → it is excluded from the comparison
   - `renderRoadmapFromDb` does **not** filter `in_progress` slices → they render into `ROADMAP.md`

   The two-passage loop is broken only by leaving placeholder slices in a non-closed, task-less state.

## Root cause synthesis

   The two checks encode contradictory assumptions about the same slice lifecycle state (`skipped`):

   | Component | Assumption about `skipped` |
   |---|---|
   | `renderRoadmapFromDb` (`markdown-renderer.ts`) | skipped slices should **not appear** in the roadmap |
   | `getSlicesReadyForDivergenceCheck` / `milestoneHasDivergence` (`roadmap.ts`) | skipped slices **are ready** and **must appear** in the roadmap |
   | `detectDiskSliceIdDivergenceForMilestone` (`artifact-db.ts`) | expects an **exact** bare slice id derived from the PLAN filename to exist in the DB |

   A milestone that legitimately needs a placeholder slice (e.g. to satisfy the filename→id derivation in Problem 1) cannot keep that slice `skipped` without falling into the Problem 2 loop, and cannot omit
 it without falling into the Problem 1 drift. There is no slice status that satisfies both checks simultaneously for a placeholder slice.

## Suggested fixes (not mutually exclusive)

   Pick the smallest one that removes the contradiction; the goal is to make `skipped` slices consistent across both checks, or to make the filename→id derivation not require a bare placeholder in the first
 place.

   1. **Align the "ready" definition with the render filter** — in `getSlicesReadyForDivergenceCheck`, treat `skipped` as **not ready** (exclude it from the comparison), mirroring `renderRoadmapFromDb`
 filtering them out. This is the most localized fix and directly removes the loop.

   2. **Render skipped slices into the roadmap** — stop filtering `status === "skipped"` in `renderRoadmapFromDb`, and render them in a clearly descopeed state (e.g. a `-` or a `~~struck~~` checkbox).
 filtering them out. This is the most localized fix and directly removes the loop.

   2. **Render skipped slices into the roadmap** — stop filtering `status === "skipped"` in `renderRoadmapFromDb`, and render them in a clearly descopeed state (e.g. a `-` or a `~~struck~~` checkbox).
 Restores symmetry but changes the roadmap's visible surface.

   3. **Relax the filename→id derivation** — in `detectDiskSliceIdDivergenceForMilestone`, when the derived bare id (`S01`) is absent from the DB, fall back to a prefix/alias match against existing slice ids
 (e.g. `S01-replan`), instead of requiring an exact bare id. Removes the need for a bare placeholder slice entirely, which also removes the Problem 2 trigger.

   4. **Document the contract** — regardless of the chosen fix, surface (in the drift error message) that the check requires an **exact** bare slice id derived from the PLAN filename, and that `skipped`
 slices are excluded from the roadmap but included in the divergence comparison. The current behavior is non-obvious and forces a workaround that itself triggers the next drift.

## Reproduction / notes

   - Trigger conditions: a milestone phase directory contains a PLAN file `NN-MM-PLAN.md` whose derived bare slice id `S{MM}` is not present as an exact id in the DB.
   - Observed resolution path: (a) `/gsd auto` pauses on `disk-slice-id-divergence`; (b) a bare `S{MM}` slice is added and marked `skipped`; (c) `/gsd auto` resumes, then pauses on `roadmap-divergence
 persisted after cap=2 passes`; (d) the skipped placeholder is reopened to `in_progress` (task-less); (e) both checks pass.
   - The loop is in the reconciliation layer only; deliverables and tests are unaffected. The blocker is purely workflow/state, not execution state.
   - Severity: blocks `/gsd auto` entirely until a manual workaround is applied; the workaround is not discoverable from the error message.

### GSD version

1.12.0-dev.f884717

### Node.js version

v24.18.0

### Operating system

Debian 13.6

### Install method

npm global install

### Affected area

Auto-mode / dispatch loop

### Model and provider

_No response_

```shell

```

### Impact

_No response_

### Additional information

This issue was investigated by gsd-pi itself, which resolved (temporally???) the problem in the current project. Obviously, the actual cause of the slice names not being in SNN format, but slices with names like:
S00-blocker, S01-replan, S02-db-repair, S03-execute, S04-validate, S05-trigger-resolver, S06-hooks-planning, S07-tui-wizard, S08-e2e-test-harness, S09-docs-and-req, needs to be analyzed.
I don't know if the error is in the match/search rules or the creation rules.

jeremymcs:
## Summary
- Fixed the two-state drift loop by prefix-matching suffixed slice ids against bare PLAN-filename ids and by sharing one skipped-slice predicate between the roadmap renderer and the divergence check; verified with 4 new regression tests plus the drift, renderer, and status-guard suites and verify:fast.

## Bugs Addressed
- [x] **disk-slice-id-divergence requires exact bare slice id from PLAN filename**
- [x] **skipped slices are 'ready' for roadmap divergence but filtered from rendered ROADMAP.md**

## Verification

## Related Issue
- Closes #1623
- [#1623 [Bug]: [drift-check] /gsd auto enters a two-state drift loop: disk-slice-id-divergence fix triggers roadmap-divergence](https://github.com/open-gsd/gsd-pi/issues/1623)

## User
- @efrembaraldo

## Repo
- `open-gsd/gsd-pi`

## Branch
- `issue/1623-bug-drift-check-gsd-auto-enters-a-two-st-1786113581`

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-pi/pr/1625"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a>

### DIFF
--- a/src/resources/extensions/gsd/markdown-renderer.ts
+++ b/src/resources/extensions/gsd/markdown-renderer.ts
@@ -12,7 +12,7 @@
 import { readFileSync, existsSync, mkdirSync, statSync } from "node:fs";
 import { createProjectionDirectorySync, removeProjectionFileSync } from "./atomic-write.js";
 import { logWarning } from "./workflow-logger.js";
-import { isClosedStatus, toStatus } from "./status-guards.js";
+import { isClosedStatus, isHiddenFromRoadmap, toStatus } from "./status-guards.js";
 import { dirname, join } from "node:path";
 import {
   getAllMilestones,
@@ -568,7 +568,11 @@ export async function renderRoadmapFromDb(
     throw new Error(`milestone ${milestoneId} not found`);
   }
 
-  const slices = getMilestoneSlices(milestoneId).filter((slice) => slice.status !== "skipped");
+  // Shared with roadmap-divergence detection (#1623) so the projection and the
+  // drift check agree on which slices ROADMAP.md is expected to contain.
+  const slices = getMilestoneSlices(milestoneId).filter(
+    (slice) => !isHiddenFromRoadmap(slice.status),
+  );
 
   // Refuse to render a stub ROADMAP for an unplanned milestone (#852).
   // A milestone row created by gsd_milestone_generate_id / ensureMilestoneDbRow
--- a/src/resources/extensions/gsd/state-reconciliation/drift/artifact-db.ts
+++ b/src/resources/extensions/gsd/state-reconciliation/drift/artifact-db.ts
@@ -320,6 +320,26 @@ function milestoneIdAliases(
   return aliases;
 }
 
+/**
+ * Slice ids that a bare, PLAN-filename-derived id may legitimately resolve to.
+ *
+ * #1623: a flat-phase plan file only encodes a numeric index (`NN-MM-PLAN.md`),
+ * so the derived id is always bare (`S01`). DB rows, however, routinely carry a
+ * suffix — `S01-replan`, `S02-db-repair`, `S00-blocker` — and exact-equality
+ * membership flagged every one of them as `disk-slice-id-divergence`, pausing
+ * `/gsd auto` with no repair path short of hand-creating a bare placeholder
+ * slice. Accept the bare id itself or any id that extends it with a `-<suffix>`;
+ * the `-` separator keeps `S1` from matching `S10`.
+ */
+function knownSliceIdMatches(knownSliceIds: ReadonlySet<string>, sliceId: string): boolean {
+  if (knownSliceIds.has(sliceId)) return true;
+  const prefix = `${sliceId}-`;
+  for (const known of knownSliceIds) {
+    if (known.startsWith(prefix)) return true;
+  }
+  return false;
+}
+
 function detectDiskSliceIdDivergenceForMilestone(
   milestoneId: string,
   filesystemIds: string[],
@@ -343,7 +363,7 @@ function detectDiskSliceIdDivergenceForMilestone(
       if (!planMatch) continue;
       const planNum = parseInt(planMatch[1]!, 10);
       const sliceId = `S${String(planNum).padStart(2, "0")}`;
-      if (knownSliceIds.has(sliceId)) continue;
+      if (knownSliceIdMatches(knownSliceIds, sliceId)) continue;
       drifts.push({
         kind: "disk-slice-id-divergence",
         milestoneId,
--- a/src/resources/extensions/gsd/state-reconciliation/drift/roadmap.ts
+++ b/src/resources/extensions/gsd/state-reconciliation/drift/roadmap.ts
@@ -16,7 +16,11 @@ import { renderRoadmapFromDb } from "../../markdown-renderer.js";
 import { findMilestoneIds } from "../../milestone-ids.js";
 import { parseRoadmap } from "../../parsers-legacy.js";
 import { resolveMilestoneFile } from "../../paths.js";
-import { isClosedStatus, isSkippedForDispatch } from "../../status-guards.js";
+import {
+  isClosedStatus,
+  isHiddenFromRoadmap,
+  isSkippedForDispatch,
+} from "../../status-guards.js";
 import type { GSDState } from "../../types.js";
 import type { DriftContext, DriftHandler, DriftRecord } from "../types.js";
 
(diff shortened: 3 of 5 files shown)


jeremymcs:

Re-checked PR #1625 against [#1623 [Bug]: [drift-check] /gsd auto enters a two-state drift loop: disk-slice-id-divergence fix triggers roadmap-divergence](https://github.com/open-gsd/gsd-pi/issues/1623).

## Subtasks
- [x] **disk-slice-id-divergence requires exact bare slice id from PLAN filename**
- [x] **skipped slices are 'ready' for roadmap divergence but filtered from rendered ROADMAP.md**

PR #1731, opened 2026-08-13
fix: isolate canonical read tools from global DB handle

### LINKED ISSUE #1727: fix(gsd): isolate canonical reads from workflow database state
## Summary
Canonical read tools currently reach the database through pathways that can alter or depend on the global workflow DB handle. We need isolated, read-only database access for canonical reads.

Follow-up to #1608  
Related: #1613, #1682  
Potentially affects released behavior in v1.15.0.

## Contract questions (maintainer confirmation needed before implementation)
1. Should `gsd_requirement_list` also support `includeSuperseded` (as earlier discussed in #1608), or keep the current surface?
2. Is the hard list limit final at 500, or should it be 1000?
3. For decision ordering, should canonical list order be `seq ASC` or `seq DESC`?
4. For MCP, should `query_error` always be returned as structured tool output, or are protocol-level errors acceptable in specific cases?

## Problem
Read tools must not:
- mutate global DB state,
- switch global project context,
- or create missing DB files as a side effect of read operations.

## Target behavior
- Introduce a single read boundary (e.g. `withCanonicalReadAdapter()`).
- Resolve path via `resolveProjectRootDbPath()`.
- Open only existing DB in read-only isolated mode (e.g. via `openWorkflowDatabaseIsolated()`).
- Pass adapter explicitly to canonical query helpers (no hidden global `_getAdapter()` dependency for this path).
- Ensure adapter closure in `finally` on success and failure.
- Align both native and MCP canonical read tools with this boundary.

## Required regression tests
- Missing DB returns `db_unavailable`.
- No new `gsd.db` gets created as a side effect.
- Global project A remains active after isolated read from project B.
- Isolated adapter always closes, including on query failure.
- Native and MCP both read the same temp DB fixture behaviorally.

## Acceptance
- Behavioral tests reproduce previous failure and pass after fix.
- No global-handle mutation from canonical read operations.
- No side-effect DB creation from read operations.

pimmink:
## Summary
- route canonical read tools through an isolated per-project DB adapter instead of `ensureDbOpen` global-handle switching
- apply the same isolated adapter strategy to MCP canonical read handlers for parity with native tooling
- add canonical read regression tests that validate no implicit DB creation, no global handle mutation, structured `query_error` handling, and native/MCP parity

## Changes
- add `runSerializedCanonicalReadOperation` in MCP workflow tools and migrate canonical decision/requirement read handlers
- add `withCanonicalReadAdapter` execution path in native db tools and move canonical read handlers to strict read functions with adapter injection
- allow strict context-store read functions to accept an explicit adapter for isolated reads
- replace prior lightweight canonical-read tests with temp-DB behavioral regressions

## Validation
- `pnpm run test:compile`
- `node --import ./scripts/dist-test-resolve.mjs --experimental-test-isolation=process --test dist-test/src/resources/extensions/gsd/tests/canonical-read-tools.test.js`
- `pnpm run verify:fast` *(known local path-encoding failure in `validate-pack` test under space-containing workspace path; unrelated to canonical read logic)*

Closes #1727

### DIFF
--- a/packages/mcp-server/src/workflow-tools.ts
+++ b/packages/mcp-server/src/workflow-tools.ts
@@ -1287,6 +1287,30 @@ async function runSerializedWorkflowDbOperation<T>(
   });
 }
 
+async function runSerializedCanonicalReadOperation<T>(
+  projectDir: string,
+  fn: (adapter: { close(): void }) => Promise<T>,
+): Promise<T> {
+  return runSerializedWorkflowOperation(async () => {
+    const { resolveProjectRootDbPath, openWorkflowDatabaseIsolated } = await importWorkflowRuntimeModule<any>(
+      "../../../src/resources/extensions/gsd/db-workspace.js",
+    );
+    const dbPath = resolveProjectRootDbPath(projectDir);
+    if (!existsSync(dbPath)) {
+      throw new Error("Database adapter not available (db_unavailable)");
+    }
+    const adapter = openWorkflowDatabaseIsolated(dbPath);
+    if (!adapter) {
+      throw new Error("Database adapter not available (db_unavailable)");
+    }
+    try {
+      return await fn(adapter);
+    } finally {
+      adapter.close();
+    }
+  });
+}
+
 async function enforceWorkflowWriteGate(
   toolName: string,
   projectDir: string,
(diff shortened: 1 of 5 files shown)


pimmink:
Implementation is complete and branch-local validation is green. CI/build is still pending; once it passes this PR is ready to merge.

PR #1750, opened 2026-08-14
fix(issue): Orphaned Task Attempt from a killed auto-mode process can never be settled (milestone_leases keeps no history)

### LINKED ISSUE #1748: Orphaned Task Attempt from a killed auto-mode process can never be settled (milestone_leases keeps no history)
## Summary

If the process running `gsd auto` is killed (SIGTERM/SIGKILL — crash, terminal closed, OOM, etc.) while a Task Attempt is `running`, that attempt can be **permanently un-settleable** through any normal path, because `milestone_leases` retains only the single most recent lease row per milestone (no history). Once a newer lease has been issued, there is no way to re-establish the fencing token the orphaned attempt was claimed under, and every settle path that requires "a currently held lease matching the attempt's worker/token" fails forever.

## Environment

- `@opengsd/gsd-pi` v1.15.0
- macOS, local SQLite-backed project

## Reproduction evidence

Auto-mode's own audit log recorded the kill directly:

```
"Auto-mode process received a termination signal"  (2026-08-13T19:27:56Z, approx.)
```

This happened mid-execution of `execute-task M008/S01/T01`, attempt #5 (`c57966a5-dfa6-4bc5-8afc-a8646ce01bba`), claimed at `2026-08-13T19:27:54.775Z` under worker `auto-Dvaraka2601.local-40183-31796528`, lease token `14`.

That attempt row was left at `attempt_state = 'running'` with no `ended_at`, forever, because:

- `milestone_leases` is keyed by `milestone_id` alone (one row per milestone) — the token-14 lease row was long since overwritten by later takeovers (tokens 16, 17, 19, 21, ... by the time we investigated).
- `trg_workflow_attempt_transition_fencing` (the SQLite trigger gating any `UPDATE` to `workflow_execution_attempts.attempt_state`) requires either (a) a *currently held, non-expired* lease matching the attempt's original `worker_id`/`milestone_lease_token`, or (b) the interrupt-with-recovery path (`recovery_worker_id`/`recovery_milestone_lease_token`, backed by an `attempt.interrupt` operation, itself requiring *its own* currently-held lease).
- Since (a) is permanently impossible (that specific token is gone from `milestone_leases`), and nothing in `gsd_doctor`, `/gsd auto`, `gsd_task_recovery_resume`, or `gsd_task_reopen` attempts path (b) automatically, the attempt just sits `running` indefinitely.

Concretely, every subsequent `/gsd auto` run against this project hit one of:

```
Error: workflow attempt requires the current held lease or current replacement lease
Error: Task recovery resume requires the current agent-owned abort
```

...neither of which is the actual problem (the actual problem is the orphaned attempt, not the recovery-action state).

Once we understood this, resolving it required calling gsd-pi's own `settleTaskAttempt({ outcome: 'interrupted', recovery: { workerId, milestoneLeaseToken } })` directly (after first registering a fresh worker + claiming a fresh lease to satisfy path (b) above) — there is no MCP tool, CLI subcommand, or `gsd doctor` check that does this.

## Impact

Any process-level kill of a running `gsd auto` session (which is not a rare event — a laptop sleep, an OOM, a user Ctrl-C at the wrong moment, a supervisor timeout) can permanently strand that task's lifecycle in a way that:

- `gsd_doctor` doesn't detect (checked both scoped to the milestone and unscoped — zero issues reported either way).
- `gsd_task_reopen` explicitly refuses to touch (`"Task reopen cannot target a running Attempt"`).
- `gsd_task_recovery_resume` doesn't apply to (it only resolves `abort`-type recovery actions, and an orphaned-but-never-failed attempt never generates one).
- Blocks not just that task, but eventually `gsd_slice_complete` for the whole slice (`"Task T01 has a running Attempt descendant"`), since slice completion unconditionally checks every task's lifecycle for a running attempt.

## Suggested direction

A `gsd doctor` check (or a startup reconciliation pass in `auto/loop.ts`) that detects attempts stuck `running` whose owning worker is no longer in `workers` with `status='active'` (or has expired its heartbeat), and offers to settle them as `interrupted` via the existing `attempt.interrupt` recovery path — automatically claiming a fresh lease to do so, since the worker/lease that originally claimed them is unrecoverable by construction.

jeremymcs:
## Summary
- Reconciled orphaned Task Attempts through replacement leases, with 148 focused tests and fast gates passing.

## Verification

## Related Issue
- Closes #1748
- [#1748 Orphaned Task Attempt from a killed auto-mode process can never be settled (milestone_leases keeps no history)](https://github.com/open-gsd/gsd-pi/issues/1748)

## User
- @rsieb

## Repo
- `open-gsd/gsd-pi`

## Branch
- `issue/1748-orphaned-task-attempt-from-a-killed-auto-1786670157`

### DIFF
--- a/src/resources/extensions/gsd/auto/loop.ts
+++ b/src/resources/extensions/gsd/auto/loop.ts
@@ -134,6 +134,7 @@ import {
 } from "./custom-task-host-verification.js";
 import {
   claimTaskAttempt,
+  interruptOrphanedTaskAttempts,
   readLatestTaskAttempt,
   readTaskAttempt,
   settleTaskAttempt,
@@ -465,6 +466,21 @@ export async function autoLoop(
 ): Promise<void> {
   debugLog("autoLoop", { phase: "enter" });
   resetSessionTimeoutState();
+  if (s.workerId && s.currentMilestoneId) {
+    const repaired = interruptOrphanedTaskAttempts({
+      workerId: s.workerId,
+      milestoneId: s.currentMilestoneId,
+    });
+    if (repaired.milestoneLeaseToken !== null) {
+      s.milestoneLeaseToken = repaired.milestoneLeaseToken;
+    }
+    if (repaired.attemptIds.length > 0) {
+      debugLog("autoLoop", {
+        phase: "orphaned-task-attempts-interrupted",
+        attemptIds: repaired.attemptIds,
+      });
+    }
+  }
   let iteration = 0;
   const dispatchContract = options?.dispatchContract ?? "legacy-direct";
   const unitDispatchDeps = createExecutionGraphUnitDispatchDeps();
--- a/src/resources/extensions/gsd/task-execution-domain-operation.ts
+++ b/src/resources/extensions/gsd/task-execution-domain-operation.ts
@@ -23,7 +23,16 @@ import {
   type StagedTaskCompletionWriteInput,
   writeStagedTaskCompletion,
 } from "./db/writers/task-execution.js";
-import type { ExecutionInvocation } from "./execution-invocation.js";
+import {
+  claimMilestoneLease,
+  getMilestoneLease,
+} from "./db/milestone-leases.js";
+import { autoWorkerHeartbeatTtlSeconds } from "./db/auto-workers.js";
+import {
+  internalExecutionInvocation,
+  type ExecutionInvocation,
+} from "./execution-invocation.js";
+import { isDbAvailable } from "./gsd-db.js";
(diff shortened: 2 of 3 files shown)


jeremymcs:

Re-checked PR #1750 against [#1748 Orphaned Task Attempt from a killed auto-mode process can never be settled (milestone_leases keeps no history)](https://github.com/open-gsd/gsd-pi/issues/1748).

## Subtasks
- [x] **Orphaned Task Attempt from a killed auto-mode process can never be settled (milestone_leases keeps no history)**

PR #1820, opened 2026-08-17
fix(issue): Auto-mode: complete-slice tool error is finalized as "next" (no-artifact never converted to a retry → deterministic liveness wedge)

### LINKED ISSUE #1794: Auto-mode: complete-slice tool error is finalized as "next" (no-artifact never converted to a retry → deterministic liveness wedge)
## Summary

In auto-mode, a `complete-slice` unit whose `gsd_slice_complete` call **returns a tool-level error** (so no SUMMARY/UAT artifact is written and no rows are committed) is still finalized with `action: "next"`. The failure is recorded only as a journal `unit-end` event with `status: "no-artifact"`; it is never converted into a `retry`, `blocker`, or `error` action. As a result the tool error is never surfaced back to the agent so it can fix it. The orchestrator instead sees "completed → next", notices the state did not advance, re-dispatches the identical unit, and only the ADR-047 liveness backstop eventually catches it — after a second full (multi-minute) agent turn — and hard-wedges the run.

The net effect: a completely recoverable, self-describing tool error (the tool's own error string says exactly what to fix) is turned into a hard auto-mode wedge that requires manual `--resume-wedge` intervention, and burns two long agent dispatches doing it.

## Version

`@opengsd/gsd-pi@1.15.0`

## Root cause

`dist/resources/extensions/gsd/auto/unit-phase.js`:

The `no-artifact` condition is derived **for the journal event only** and is not fed back into the returned action.

```js
// ~line 700
const artifactVerified = skipArtifactVerification ||
    isTaskExecutionReadyForHostVerification(unitType, unitId) ||
    verifyExpectedArtifact(unitType, unitId, s.basePath);
...
// ~line 730 — status is downgraded ONLY for the journal event
const unitEndStatus = !artifactVerified && unitResult.status === "completed"
    ? "no-artifact"
    : unitResult.status;
deps.emitJournalEvent({ ... status: unitEndStatus, artifactVerified, ... });
...
// ~line 753 — action is returned UNCONDITIONALLY, artifactVerified is never consulted
return { action: "next", data: { ... } };
```

`unitResult.status` is `"completed"` because, from the dispatched agent's perspective, its turn ended normally (it called the tool, received an error object, and stopped). Artifact verification then fails (`verifyExpectedArtifact` → false, because the tool wrote nothing). That `!artifactVerified && completed` state is labelled `no-artifact` in the journal — but the function still falls through to `return { action: "next" }`. Nothing in the finalize/advance path distinguishes a real artifact-backed completion from a `no-artifact` one; only the liveness backstop does, and only after a 2× recurrence.

## Reproduce

Any unit where the completion tool can return an error for a *correctable* reason will trigger this. Minimal repro with a backend-only slice (no UI):

1. A slice whose tasks are all `done` and green.
2. The `complete-slice` agent authors a `## UAT Type` section that does **not** contain the exact bullet `- UAT mode: runtime-executable` (e.g. it omits the mode line, or the demo/UAT text trips the browser-requirement classifier — see "contributing trigger" below).
3. `gsd_slice_complete` returns `{ error: "UAT requires browser verification ... Re-author the UAT Type section and complete the slice again." }` and writes no artifact.

Observed journal sequence:

```
unit-end            status: "no-artifact", artifactVerified: false     ← tool returned an error, 0 rows committed
post-unit-finalize  status: "completed", action: "next"                ← finalized as success anyway
orchestrator        advance-skipped: "state did not advance ..."        ← nothing changed
(repeat once)
orchestrator        advance-blocked: "liveness backstop tripped: completed-no-advance recurred 2x ..."
```

Expected: on `no-artifact`, the tool's error string is surfaced to the dispatched agent as a `retry`/blocker on the *same* unit so it can re-author and re-call, rather than reporting `completed → next` and deferring to the backstop.

## Why it is deterministic ("recurred 2x with unchanged inputs")

The completion agent re-derives the UAT from the fixed slice-plan `Demo`/verification text on every dispatch. Fixed input text → same UAT authored → same tool rejection → same `no-artifact` → same `action: next`. The two dispatches are byte-identical in effect, which is exactly what the backstop's "unchanged inputs" guard keys on. So the loop can never self-correct; it can only wedge.

## Contributing trigger (secondary)

The reason the tool rejects in the first place is the browser-UAT classification gate in `dist/resources/extensions/gsd/tools/complete-slice.js` (~line 265) combined with `BROWSER_REQUIREMENT_RE` in `dist/resources/extensions/gsd/browser-evidence.js`. For a **backend-only** slice (proven by a test command such as `dotnet test` / `npx vitest`, no UI at all), ordinary domain verbs sitting near words like `page` can push the UAT over the browser-requirement threshold, and if the agent has not written the exact `- UAT mode: runtime-executable` bullet, the content defaults to `artifact-driven` and is rejected. This part is arguably working-as-designed as a guardrail — but because of the primary bug above, its rejection escalates into a hard wedge instead of a recoverable in-turn retry.

Two things would each independently prevent the wedge:

1. **(primary)** Treat `no-artifact` as a non-`next` outcome: return the tool's error to the dispatched agent as a same-unit retry/blocker instead of `action: "next"`, so a self-describing, correctable tool error is fixed in-turn rather than deferred to the liveness backstop.
2. **(secondary, optional)** Let a slice plan carry a persisted UAT-mode declaration so the completion agent does not have to re-derive it from prose each dispatch, removing the non-determinism at the source.

## Impact

- A recoverable tool error becomes a hard auto-mode stop requiring manual `--resume-wedge`.
- Two long agent dispatches are spent before the backstop fires.
- Most likely to bite backend-only slices whose completion legitimately needs no browser verification.

jeremymcs:
## Summary
- Fixed complete-slice tool errors to retry with preserved error context; focused tests and typechecking pass.

## Verification

## Related Issue
- Closes #1794
- [#1794 Auto-mode: complete-slice tool error is finalized as "next" (no-artifact never converted to a retry → deterministic liveness wedge)](https://github.com/open-gsd/gsd-pi/issues/1794)

## User
- @sandersietses

## Repo
- `open-gsd/gsd-pi`

## Branch
- `issue/1794-auto-mode-complete-slice-tool-error-is-f-1786936192`

### DIFF
--- a/src/resources/extensions/gsd/auto/unit-phase.ts
+++ b/src/resources/extensions/gsd/auto/unit-phase.ts
@@ -42,6 +42,7 @@ import { classifyError, isTransient } from "../error-classifier.js";
 import { setCurrentPhase, clearCurrentPhase } from "../../shared/gsd-phase-state.js";
 import { setAutoActiveStatus } from "../auto-dashboard.js";
 import { runUnit } from "./run-unit.js";
+import { verificationRetryKey } from "./verification-retry-policy.js";
 import { validateSourceWriteWorktreeSafety } from "./worktree-safety-phase.js";
 import { isTaskExecutionReadyForHostVerification } from "./task-execution-cutover.js";
 import {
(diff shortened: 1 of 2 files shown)


jeremymcs:

Re-checked PR #1820 against [#1794 Auto-mode: complete-slice tool error is finalized as "next" (no-artifact never converted to a retry → deterministic liveness wedge)](https://github.com/open-gsd/gsd-pi/issues/1794).

## Subtasks
- [x] **Auto-mode: complete-slice tool error is finalized as "next" (no-artifact never converted to a retry → deterministic live**

PR #1858, opened 2026-08-19
fix(issue): [Bug]: v1.16.0 TUI shows only "Working..." — PR #1736 removed `partial` from delta events but agent-loop.ts never implemented the promised local accumulator

### LINKED ISSUE #1857: [Bug]: v1.16.0 TUI shows only "Working..." — PR #1736 removed `partial` from delta events but agent-loop.ts never implemented the promised local accumulator
## Problem

Since v1.16.0 the interactive TUI shows only `⠇ Working...` for the entire duration of a turn. No streaming text, no tool calls, no subagent prompts — no incremental progress of any kind. The agent is genuinely working (tools execute, artifacts are written, the final message eventually paints), but the user has zero visibility while it runs.

Regression introduced by PR #1736 (merged 2026-08-16), which shipped in v1.16.0 (release backfill #1824, 2026-08-17). v1.15.0 is unaffected.

## Root Cause

`packages/pi-agent-core/src/agent-loop.ts:417-456`

PR #1736 removed the `partial: AssistantMessage` field from all delta event variants to fix O(n²) payload overhead (#1735). Its stated contract was that *"consumers maintain their own local `partialMessage` accumulator."* For the agent-loop consumer, that accumulator was never implemented.

`partialMessage` is assigned exactly once — line 432, from `event.partial` on the `start` event — and thereafter only read:

```ts
let partialMessage: AssistantMessage | null = null;   // :417

case "start":
    partialMessage = event.partial;                    // :432  ← the only assignment
    ...

case "text_start": case "text_delta": case "text_end":
case "thinking_start": case "thinking_delta": case "thinking_end":
case "toolcall_start": case "toolcall_delta": case "toolcall_end":
    if (partialMessage) {
        // partialMessage is already updated locally by individual delta events
        // (text_delta appends to text, thinking_delta to thinking, etc.)
        //   ^^^ this comment describes logic that does not exist anywhere in the file
        context.messages[context.messages.length - 1] = partialMessage;
        await emit({
            type: "message_update",
            assistantMessageEvent: event,
            message: partialMessage,
        });
    }
    break;
```

The exact regression — `git diff v1.15.0 v1.16.0` on that block:

```diff
-  partialMessage = event.partial;
+  // partialMessage is already updated locally by individual delta events
+  // (text_delta appends to text, thinking_delta to thinking, etc.)
 ...
-  message: { ...partialMessage },
+  message: partialMessage,
```

Two independent defects result:

**1. Stale (empty) content.** Providers emit `start` with a freshly-initialized message whose `content` is `[]` — see `packages/pi-ai/src/providers/anthropic.ts:463-465` (`const output: AssistantMessage = { role: "assistant", content: [], ... }`) pushed at `:528` (`stream.push({ type: "start", partial: output })`). Delta events now carry only the incremental string (`:624-626`: `stream.push({ type: "text_delta", contentIndex: index, delta: event.delta.text })`). Because nothing repopulates `partialMessage`, the loop forwards that empty-content snapshot for the whole turn.

**2. Lost object identity.** Changing `message: { ...partialMessage }` to `message: partialMessage` means every `message_update` carries the *same object reference*. The TUI consumer at `packages/gsd-agent-modes/src/modes/interactive/controllers/chat-controller.ts:201` does `host.streamingMessage = event.message` — assigning an unchanged reference, so downstream change detection sees nothing new to paint.

The `pi-tui` / `gsd-agent-modes` renderer was never updated to compensate: PR #1736's own Scope checklist leaves `pi-tui` unchecked, and its 26-file diff touches no renderer file.

Why the failure is partial rather than total: the Anthropic provider mutates its content blocks by reference (`:622`, `block.text += event.delta.text`), so `context.messages` does eventually hold real content and `message_end` paints the final answer. `tui-transcript-tracker.ts:44-51` also reads `inner.delta` directly. What is uniformly lost is *incremental* progress — tool cards, subagent prompts, live streaming text.

## Expected Behavior

Implement the local accumulator that #1736 promised, and restore per-event object identity so the renderer detects changes.

In `agent-loop.ts`, apply each delta to a locally-maintained message before emitting, and emit a fresh object:

```ts
case "text_delta":
    if (partialMessage) {
        const block = partialMessage.content[event.contentIndex];
        if (block?.type === "text") block.text += event.delta;
        // ... analogous handling for thinking_delta / toolcall_delta,
        //     and *_start events that append new content blocks
        context.messages[context.messages.length - 1] = partialMessage;
        await emit({
            type: "message_update",
            assistantMessageEvent: event,
            message: { ...partialMessage },   // fresh identity per update
        });
    }
    break;
```

Concretely:
- `text_start` / `thinking_start` / `toolcall_start` must append the new content block to `partialMessage.content` at `contentIndex`.
- `text_delta` / `thinking_delta` / `toolcall_delta` must append their `delta` to the block at `contentIndex`.
- Restore the `{ ...partialMessage }` spread on the emitted `message` so consumers relying on reference change still update.

This preserves #1736's perf win (delta events stay small; the accumulator is local) while restoring the streaming UI. A regression test asserting that `message_update` payloads contain monotonically growing content during a multi-delta stream would catch this class of break.

## Environment

- GSD version: 1.16.0 (installed via npm, global)
- Previous working version: 1.15.0 (local dev build via `scripts/dev-cli.js`)
- Platform: Linux
- Provider: Anthropic (mutate-by-reference masks the failure into a partial one; providers that rebuild content per delta would show a total blackout)
- Unit: not unit-specific — affects every interactive turn

## Reproduction Context

1. Install v1.16.0 and start an interactive session.
2. Issue any prompt that involves tool calls or a multi-token response.
3. Observe `⠇ Working...` for the whole turn — no tool cards, no subagent prompts, no streaming text.
4. The final assistant message appears only at `message_end`.

Downgrading to 1.15.0 restores normal streaming output, confirming the bisect.

## Forensic Evidence

- Automated forensics reported **0 anomalies** and no active milestone/slice — consistent with a pure presentation-layer regression, not a state or workflow defect.
- Single-assignment proof — every `partialMessage` occurrence in v1.16.0 `agent-loop.ts`:
  - `:417` declaration (`= null`)
  - `:432` assignment from `event.partial` (**only** write)
  - `:433`, `:435`, `:447`, `:450`, `:454` — all reads
- `packages/pi-ai/src/types.ts:410` — `partial` now survives only on `{ type: "start"; partial: AssistantMessage }`, confirming delta events no longer carry it.
- The identical defect is present in the shipped build (`packages/pi-agent-core/dist/agent-loop.js:286-321`), including the misleading comment at `:315` — so this is not a source/build skew.
- Duplicate check performed: #1735 is the *originating* perf issue (closed by #1736), not a report of this regression. #1684 / #1008 / #900 are unrelated render-perf issues. No open or merged PR addresses the missing accumulator.

## Workaround

```bash
npm i -g @opengsd/gsd-pi@1.15.0
```

Work performed under v1.16.0 is not lost or corrupted — only invisible. Progress can be followed live via `.gsd/activity/*.jsonl`.

---

jeremymcs:
## Summary
- Restored incremental agent streaming and fresh update identities with focused regression coverage and passing verification.

## Bugs Addressed
- [x] **Agent loop does not accumulate streamed delta content**
- [x] **Message updates reuse one object reference**

## Verification

## Related Issue
- Closes #1857
- [#1857 [Bug]: v1.16.0 TUI shows only "Working..." — PR #1736 removed `partial` from delta events but agent-loop.ts never implemented the promised local accumulator](https://github.com/open-gsd/gsd-pi/issues/1857)

## User
- @splichy

## Repo
- `open-gsd/gsd-pi`

## Branch
- `issue/1857-bug-v1-16-0-tui-shows-only-working-pr-17-1787152477`

### DIFF
--- a/packages/pi-agent-core/src/agent-loop-streaming.test.ts
+++ b/packages/pi-agent-core/src/agent-loop-streaming.test.ts
@@ -1,0 +1,132 @@
+import assert from "node:assert/strict";
+import { test } from "node:test";
+import {
+	type AssistantMessage,
+	type AssistantMessageEvent,
+	EventStream,
+	type Message,
+	type Model,
+	type UserMessage,
+} from "@gsd/pi-ai";
+import { agentLoop } from "./agent-loop.js";
+import type { AgentContext, AgentLoopConfig, AgentMessage } from "./types.js";
+
+class MockAssistantStream extends EventStream<AssistantMessageEvent, AssistantMessage> {
+	constructor() {
+		super(
+			(event) => event.type === "done" || event.type === "error",
+			(event) => {
+				if (event.type === "done") return event.message;
+				if (event.type === "error") return event.error;
+				throw new Error("Unexpected event type");
+			},
+		);
+	}
+}
+
+function createAssistantMessage(content: AssistantMessage["content"]): AssistantMessage {
+	return {
+		role: "assistant",
+		content,
+		api: "openai-responses",
+		provider: "openai",
+		model: "mock",
+		usage: {
+			input: 0,
+			output: 0,
+			cacheRead: 0,
+			cacheWrite: 0,
+			totalTokens: 0,
+			cost: { input: 0, output: 0, cacheRead: 0, cacheWrite: 0, total: 0 },
+		},
+		stopReason: "stop",
+		timestamp: Date.now(),
+	};
+}
+
+function createUserMessage(text: string): UserMessage {
+	return { role: "user", content: text, timestamp: Date.now() };
+}
+
+function createModel(): Model<"openai-responses"> {
+	return {
+		id: "mock",
+		name: "mock",
+		api: "openai-responses",
+		provider: "openai",
+		baseUrl: "https://example.invalid",
+		reasoning: false,
+		input: ["text"],
+		cost: { input: 0, output: 0, cacheRead: 0, cacheWrite: 0 },
+		contextWindow: 8192,
+		maxTokens: 2048,
+	};
+}
+
+function identityConverter(messages: AgentMessage[]): Message[] {
+	return messages.filter((message) =>
+		message.role === "user" || message.role === "assistant" || message.role === "toolResult"
+	) as Message[];
+}
+
+test("agent loop accumulates streamed content and emits fresh message objects", async () => {
+	const context: AgentContext = { systemPrompt: "", messages: [], tools: [] };
+	const config: AgentLoopConfig = { model: createModel(), convertToLlm: identityConverter };
+	const toolCall = { type: "toolCall" as const, id: "tool-1", name: "echo", arguments: { value: "hello" } };
+	const providerPartial = createAssistantMessage([]);
+	const streamFn = () => {
+		const stream = new MockAssistantStream();
+		queueMicrotask(() => {
+			stream.push({ type: "start", partial: providerPartial });
+		});
+		setImmediate(() => {
+			stream.push({ type: "thinking_start", contentIndex: 0 });
+			stream.push({ type: "thinking_delta", contentIndex: 0, delta: "think" });
+			stream.push({ type: "thinking_delta", contentIndex: 0, delta: "ing" });
+			stream.push({ type: "thinking_end", contentIndex: 0, content: "thinking" });
+			stream.push({ type: "text_start", contentIndex: 1 });
+			stream.push({ type: "text_delta", contentIndex: 1, delta: "hel" });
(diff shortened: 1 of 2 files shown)


jeremymcs:

Re-checked PR #1858 against [#1857 [Bug]: v1.16.0 TUI shows only "Working..." — PR #1736 removed `partial` from delta events but agent-loop.ts never implemented the promised local accumulator](https://github.com/open-gsd/gsd-pi/issues/1857).

## Subtasks
- [x] **Agent loop does not accumulate streaming delta content**
- [x] **Streaming updates reuse the same message object reference**

PR #1862, opened 2026-08-19
fix(issue): [Bug]: slice-prefixed task ids make parseProjectionPlan return zero tasks — startup auto-rebuild never converges (1.16.0)

### LINKED ISSUE #1859: [Bug]: slice-prefixed task ids make parseProjectionPlan return zero tasks — startup auto-rebuild never converges (1.16.0)
### Summary

When a slice's DB task ids carry a redundant slice prefix (`S05-T01` instead of bare `T01`), the PLAN renderer emits headings that `parseProjectionPlan` cannot read. The slice's PLAN then parses to **zero** tasks, so those tasks are permanently absent from `scanMarkdownHierarchy`. `checkMarkdownHierarchyAgainstDb` therefore returns `recovery-required` forever, and because `guided-flow.js` auto-runs `rebuildMarkdownProjectionsFromDb` on that verdict, **every launch re-projects the whole tree and never converges**.

On this project that is ~9 minutes and ~3.3 GB of writes per start, with the node main thread parked in `fsync` (uninterruptible `D` state on btrfs, where each fsync is a log commit + device barrier). The TUI drops all keystrokes for the duration, which reads to a user as a hang.

### Version / environment

- GSD v1.16.0, node v24.14.0, Fedora (kernel 7.1.8), `/home` on **btrfs**
- Flat-phase layout (`.gsd/phases/NN-slug/`)

### Reproduction

Milestone M016, slices S05/S06/S07 store task ids with the slice prefix:

```
sqlite> select id, slice_id from tasks where milestone_id like 'M016%';
S05-T01 | S05     <- prefixed
S06-T01 | S06     <- prefixed
T01     | S01     <- normal, same milestone
```

The renderer writes them straight into the heading:

```
16-04-PLAN.md:  - [x] **T01**: Data-quality alert strip component _(medium)_
16-05-PLAN.md:  - [x] **S05-T01**: Account State Panel with §6.7 labels _(medium)_
```

Measured against `schemas/parsers.js`:

```js
parseProjectionPlan(read("16-04-PLAN.md")).tasks   // 4  -> T01,T02,T03,T04
parseProjectionPlan(read("16-05-PLAN.md")).tasks   // 0
parseProjectionPlan(read("16-06-PLAN.md")).tasks   // 0
parseProjectionPlan(read("16-07-PLAN.md")).tasks   // 0
```

Same file format, same `<tasks>` block, same milestone — the only difference is the id shape. 11 tasks vanish from the scan.

The same concatenation also produces doubled artifact filenames, e.g. `S05-S05-T01-SUMMARY.md`, since the path is built as `<slice_id>-<task_id>-SUMMARY.md`.

### Why it can't be worked around

- `/gsd rebuild markdown` re-renders with the same ids and reports success — "Rendered: 610, Skipped: 118, Quarantined: 0" — while the counts stay byte-identical.
- Renaming only the PLAN headings to `**T01**` fixes `sameCounts` but breaks `scanIdentitiesMatch`, since the DB identity is still `M016/S05/S05-T01`. The in-sync test needs both.
- Normalizing the ids in the DB is blocked by design: `workflow_item_lifecycles` has a composite FK onto `tasks(milestone_id, slice_id, id)` plus `trg_workflow_lifecycle_identity_immutable` (task_id may never change), `trg_workflow_lifecycle_transition` (aborts any update where `lifecycle_status` is unchanged), and `trg_workflow_lifecycle_delete`.

- Hand-editing the PLAN files is reverted too: projections are content-addressed, so an externally modified file is quarantined and re-rendered from the DB on the next launch (`[reconcile] external-markdown-edit drift ... expectedSha=...`, related #1675).

So an affected project has no user-side repair path — the loop runs on every launch until this is fixed upstream.

### Suggested fixes

1. Have `parseProjectionPlan` accept an optionally slice-prefixed task id (`(?:S\d+-)?T\d+`), normalizing to the bare form — this alone restores convergence for existing projects.
2. Strip the redundant prefix at render time so ids round-trip, and normalize on write so new rows can't acquire it.
3. Independently: don't re-run the full projection unconditionally on `recovery-required`. If a rebuild leaves the verdict unchanged, back off and warn once instead of repeating a multi-GB fsync-bound pass on every launch.

Point 3 matters on its own — any non-convergent drift currently costs a full re-projection per start.

jeremymcs:
## Summary
- Fixed prefixed task projection convergence and persistent startup rebuild loops with focused regression coverage.

## Bugs Addressed
- [x] **Slice-prefixed task IDs break PLAN parsing and artifact paths**
- [x] **Persistent recovery-required verdict triggers rebuild on every launch**

## Verification

## Related Issue
- Closes #1859
- [#1859 [Bug]: slice-prefixed task ids make parseProjectionPlan return zero tasks — startup auto-rebuild never converges (1.16.0)](https://github.com/open-gsd/gsd-pi/issues/1859)

## User
- @jg-main

## Repo
- `open-gsd/gsd-pi`

## Branch
- `issue/1859-bug-slice-prefixed-task-ids-make-parsepr-1787158716`

### DIFF
--- a/src/resources/extensions/gsd/guided-flow.ts
+++ b/src/resources/extensions/gsd/guided-flow.ts
@@ -148,6 +148,11 @@ export {
 import { logWarning } from "./workflow-logger.js";
 import { deleteRuntimeKv } from "./db/runtime-kv.js";
 import { PAUSED_SESSION_KV_KEY } from "./interrupted-session.js";
+import {
+  clearMarkdownAutoRebuildBackoff,
+  recordMarkdownAutoRebuildFailure,
+  shouldAttemptMarkdownAutoRebuild,
+} from "./markdown-auto-rebuild-backoff.js";
 import { buildWorkflowDispatchContent } from "./workflow-protocol.js";
 import { isFullGsdToolSurfaceRequested, restoreGsdWorkflowTools, scopeGsdWorkflowToolsForDispatch } from "./bootstrap/register-hooks.js";
 import {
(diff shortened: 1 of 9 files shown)


jeremymcs:

Re-checked PR #1862 against [#1859 [Bug]: slice-prefixed task ids make parseProjectionPlan return zero tasks — startup auto-rebuild never converges (1.16.0)](https://github.com/open-gsd/gsd-pi/issues/1859).

## Subtasks
- [x] **Slice-prefixed task IDs break PLAN projection round-tripping**
- [x] **Non-convergent recovery triggers a full rebuild on every launch**

PR #1888, opened 2026-08-20
fix(issue): phase-dir slug drift and Windows lock rmSync no-op (#1526)

### LINKED ISSUE #1526: [Bug] Milestone phase-dir slug drift + OneDrive rmSync no-op strand the session lock and crash /gsd to terminal on Windows
## Problem

On Windows with the project path containing **non-ASCII (Cyrillic) characters**, `/gsd` crashes the whole process back to the terminal, and the auto-mode session lock becomes permanently stranded. The user is told to run a shell-removal command against `.gsd.lock` and `.gsd/auto.lock`; doing so clears them from a shell, but the next `/gsd` crashes again and re-strands the lock, so the loop never ends. Two confirmed defects combine to cause this. (Project-specific names redacted.)

## Root Cause #1 (trigger): phase-directory slug drift when a milestone is renamed

`layout-policy.js` derives the on-disk phase directory name from the **mutable** milestone title:

```js
// layout-policy.js:69
export function derivePhaseSlug(title) {
  return title.toLowerCase().replace(/[^a-z0-9]+/g,"-").replace(/^-+|-+$/g,"").slice(0,40)…;
}
// layout-policy.js:24  -> phaseDirName(phaseNum, slug) => "NN-slug"
```

The slug is deterministic from the title, but **there is no code anywhere that renames/moves the phase directory when the milestone title changes** (no `renameSync`/move of the phase dir on title update; grep found only atomic-write/extensions/maintenance renames). The file's own comment admits this is load-bearing: *"the renderer must produce the same slug for the same title on every run, or the directory churns on every projection."*

Reproduction of the drift (confirmed on a real project; titles redacted, slugs shown schematically):

```
derivePhaseSlug("<default placeholder title>") -> 01-<default-slug>   // dir created by /gsd discuss under the default title
derivePhaseSlug("<user's real title>")        -> 01-<real-slug>       // path the DB projects to after plan-milestone sets the real title
```

Observed on disk:

- Discuss phase created `phases/01-<default-slug>/{CONTEXT,DISCUSSION,ROADMAP}.md` under the default title.
- `plan-milestone` set the real title; the DB/flat-phase projection then pointed at `phases/01-<real-slug>/`, **which does not exist** — the files were never moved.
- Doctor reports `artifact_file_missing` for `phases/01-<real-slug>/01-CONTEXT-DRAFT.md` (`fixable: no`) and even "pruned" a stale `phases/01-<real-slug>/01-CONTEXT.md` row.

Result: the active milestone's artifacts are unreachable at the projected path -> inconsistent milestone state -> `/gsd` bootstrap throws -> process exits to terminal.

## Root Cause #2 (compounding): `fs.rmSync({recursive:true})` silently no-ops on directories whose path contains non-ASCII characters (Windows, Node 24)

NOTE: An earlier draft of this issue attributed the cause to the OneDrive cloud-sync driver. That was wrong. Empirical isolation below shows the sole determining factor is **non-ASCII (Cyrillic) characters in the path**, not the cloud driver.

`session-lock.js` removes the proper-lockfile lock directory (`.gsd.lock/`) with `fs.rmSync(lockDir, { recursive: true, force: true })`. On Windows + Node v24.11.1, when the lock directory's absolute path contains non-ASCII characters (e.g. a localized user folder), **`fs.rmSync` returns without throwing but the directory persists**. `fs.rmdirSync` (non-recursive) removes the same directory successfully, so this is specific to the recursive `rm` path.

Isolation matrix (all on the same machine, Node v24.11.1, empty directories):

| Directory path | Non-ASCII in path? | In OneDrive? | `fs.rmSync({recursive:true})` result |
|----------------|--------------------|--------------|--------------------------------------|
| `%TEMP%\gsd-rmtest-...` | no | no | removed |
| `OneDrive\gsd-rmtest-ascii` | no | yes | removed |
| `OneDrive\<localized>\…\testdir-ascii` | **yes** | yes | **NOT removed (no error)** |
| `OneDrive\<localized>\…\.gsd.lock` | **yes** | yes | **NOT removed; `rmdirSync` removes it** |

Conclusions: (a) OneDrive presence is irrelevant; (b) ASCII directory names under a non-ASCII parent path still fail; (c) the determining factor is non-ASCII characters anywhere in the path.

Minimal reproduction:

```
node -e 'const fs=require("fs"); fs.rmSync("<dir-under-non-ascii-path>",{recursive:true,force:true}); console.log("exists:", fs.existsSync("<dir-under-non-ascii-path>"))'
// -> exists: true   (no error thrown)
```

Every removal site in `session-lock.js` is wrapped in a bare `try { } catch { }` and **never verifies removal with `existsSync` afterward**, so the no-op is invisible. The sites driving the unrecoverable loop:

- `session-lock.js:267` — `acquireSessionLock` #3218 preflight orphan-cleanup: detects orphan, calls rmSync -> no-op -> `.gsd.lock/` persists.
- `session-lock.js:310` — catch-block retry: rmSync again -> no-op -> `lockfile.lockSync` retry throws ELOCKED -> caught -> falls through.
- `session-lock.js:334` — falls through to the hardcoded `Another auto-mode session lock is stuck...` message (the exact error seen by the user).
- `session-lock.js:484` / `:512` / `:153` — `releaseSessionLock()` and the `process.once("exit")` handler: delete `auto.lock` (file, works) but `.gsd.lock/` (directory, no-ops) -> stranded after any abnormal exit.

Net effect: any bootstrap crash (triggered by Root Cause #1 or anything else) strands `.gsd.lock/` in a state GSD can **never recover from programmatically**, because every recovery path re-runs the same no-op `rmSync`. The shell-removal command uses different Win32 calls and clears it, but the next crash re-strands it — hence the permanent loop.

## Expected Behavior / Fix

1. **Slug stability:** When a milestone title changes (or `plan-milestone` assigns the first real title after files were created under a default slug), migrate the phase directory: rename `phases/NN-<old-slug>/` -> `phases/NN-<new-slug>/` (or, better, derive the directory name from the immutable milestone id instead of the mutable title, and stop churn entirely).
2. **Verify lock removal:** After each `rmSync(lockDir)`, check `if (existsSync(lockDir))` and escalate. Concretely, fall back to `fs.rmdirSync` (which works on non-ASCII paths where `rmSync({recursive:true})` silently no-ops), retry with backoff, or shell out — instead of silently swallowing the no-op. If the directory truly cannot be removed, force a stale takeover via `proper-lockfile` rather than dead-ending on a hardcoded message.
3. **Non-ASCII path awareness:** Detect a `rmSync`+`existsSync` mismatch (surrogate for the non-ASCII / Windows recursive-rm defect) and surface an accurate, actionable message (e.g. recommend relocating the project to an ASCII-only path) instead of a shell command the engine itself cannot execute.

## Environment

- GSD version: 1.11.0
- OS: Windows
- Node.js: v24.11.1
- Project path contains non-ASCII (Cyrillic) characters (localized user folder)
- proper-lockfile: loaded (lockSync/unlockSync available) — OS-lock path active
- Unit: `/gsd` home -> `showGsdHome` (`deriveState`) and `bootstrapAutoSession` -> `acquireSessionLock`

## Reproduction Context

Fresh project: `/gsd discuss` created the milestone under the default placeholder title (dir `phases/01-<default-slug>/`); the title was then set to the real value and `plan-milestone` run (active milestone with 4 slices, 0 tasks). After that, `/gsd` crashes to terminal and strands `.gsd.lock/`. No `activity/`, `journal/`, or `debug/` artifacts were ever written — the crash is in bootstrap before the dispatch loop.

## Forensic Evidence

- Disk vs DB mismatch: milestone files exist under `phases/01-<default-slug>/` (CONTEXT.md, DISCUSSION.md, ROADMAP.md, all carrying the real title inside); the DB projects to `phases/01-<real-slug>/`, which does not exist.
- Doctor history: persistent `artifact_file_missing` for `01-CONTEXT-DRAFT.md`; `stranded_lock_directory` (fixable: yes) but its heal uses the same rmSync (`doctor-runtime-checks.js:117`) and cannot clear it.
- Lock asymmetry: `.gsd.lock/` present and empty (proper-lockfile dir marker), `.gsd/auto.lock` absent — file deletion works, directory deletion does not.
- Non-ASCII path isolation matrix above; minimal reproduction returns success while the directory persists.
- Timeline (`.gsd/notifications.jsonl`): milestone ready, then repeated identical "session loc
k is stuck" errors on each launch.

---
*Auto-generated by `/gsd forensics`; project-specific identifiers redacted. Root Cause #2 corrected: cause is non-ASCII path characters, not the OneDrive sync driver.*

jeremymcs:
## TL;DR

**What:** Rename the phase directory when a milestone title changes, and verify session-lock directory removal.
**Why:** Title changes left files under the old slug while DB/writers pointed at the new one; Windows recursive `rmSync` can silently no-op on non-ASCII paths and strand `.gsd.lock`.
**How:** Relocate the on-disk phase dir through the projection write boundary; fall back to walk+`rmdirSync` when recursive rm leaves the lock dir.

## What

- `relocatePhaseDirToCanonical` (atomic-write) renames `phases/NN-old-slug` to `phases/NN-new-title-slug` before ROADMAP render.
- `removeDirectoryVerified` tries `rmSync({recursive:true})`, then empties and `rmdirSync` if the directory still exists.
- Session lock acquire/release/stale cleanup and doctor stranded-lock heal use the verified remover.
- Stuck-lock error mentions the Windows non-ASCII recursive-rm failure and `rmdir /s /q`.

## Why

Closes #1526. Discuss created the phase dir under a placeholder title; `plan-milestone` updated the title and artifact paths without moving the directory. A later bootstrap crash then stranded `.gsd.lock` because `rmSync` returned success while the directory remained.

## How

Path builders stay read-only. The rename lives on the managed projection write boundary so title-derived writes land in the same directory as the files. Lock removal checks `existsSync` after delete instead of treating a silent no-op as success.

## Verification

- `node --import ./src/resources/extensions/gsd/tests/resolve-ts.mjs --experimental-strip-types --test src/resources/extensions/gsd/tests/verified-rmdir.test.ts src/resources/extensions/gsd/tests/phase-dir-title-rename.test.ts src/resources/extensions/gsd/tests/flat-phase-validation-path.test.ts src/resources/extensions/gsd/tests/stale-lockfile-recovery.test.ts` — 13 passed
- `bash scripts/ci-fast-gates.sh` — passed

### DIFF
--- a/src/resources/extensions/gsd/atomic-write.ts
+++ b/src/resources/extensions/gsd/atomic-write.ts
@@ -15,6 +15,15 @@ import {
 } from "./managed-projection-history.js";
 import { classifyGsdLogicalPath } from "./projection-path-policy.js";
 import { preserveManagedProjectionBeforeMutation } from "./projection-mutation-guard.js";
+import { LAYOUT_SEGMENTS } from "./layout-policy.js";
+import {
+  canonicalPhaseDirName,
+  clearPathCache,
+  gsdProjectionRoot,
+  isLegacyMilestonesLayout,
+  legacyMilestonesDir,
+  resolveMilestonePath,
+} from "./paths.js";
 export { removeLegacyProjectionTreeSync };
 
 const TRANSIENT_LOCK_ERROR_CODES = new Set(["EBUSY", "EPERM", "EACCES"]);
@@ -615,4 +624,38 @@ export function createProjectionDirectorySync(directoryPath: string): void {
       mkdirSync(directoryPath, { recursive: true });
     }
   });
+}
+
+/**
+ * Rename an existing same-milestone phase directory to the title-derived
+ * canonical name. Discuss/plan title changes otherwise leave files under the
+ * old slug while writers and the artifacts table point at the new one (#1526).
+ */
+export function relocatePhaseDirToCanonical(
+  basePath: string,
+  milestoneId: string,
+  title?: string,
+): string {
+  const dest = join(
+    gsdProjectionRoot(basePath),
+    LAYOUT_SEGMENTS.level1,
+    canonicalPhaseDirName(milestoneId, title),
+  );
+  const existing = resolveMilestonePath(basePath, milestoneId);
+  if (!existing) return dest;
+  const legacyBase = legacyMilestonesDir(basePath);
+  const isLegacy = existing.startsWith(legacyBase + "/") || existing.startsWith(legacyBase + "\\")
+    || isLegacyMilestonesLayout(basePath);
+  if (isLegacy) return existing;
+  const existingName = existing.split(/[/\\]/).pop()!;
+  const destName = dest.split(/[/\\]/).pop()!;
+  if (existingName === destName) return existing;
+  if (existsSync(dest)) return dest;
+  try {
+    renameSync(existing, dest);
+    clearPathCache();
+    return dest;
+  } catch {
+    return existing;
+  }
 }
--- a/src/resources/extensions/gsd/doctor-runtime-checks.ts
+++ b/src/resources/extensions/gsd/doctor-runtime-checks.ts
@@ -1,4 +1,5 @@
 import { existsSync, lstatSync, readdirSync, readFileSync, realpathSync, rmSync, statSync } from "node:fs";
+import { removeDirectoryVerified } from "./verified-rmdir.js";
 import { basename, dirname, join } from "node:path";
 
 import type { DoctorIssue, DoctorIssueCode } from "./doctor-types.js";
@@ -196,10 +197,9 @@ export async function checkRuntimeHealth(
             fixable: true,
           });
           if (shouldFix("stranded_lock_directory")) {
-            try {
-              rmSync(lockDir, { recursive: true, force: true });
+            if (removeDirectoryVerified(lockDir)) {
               fixesApplied.push(`removed stranded lock directory ${lockDir}`);
-            } catch {
+            } else {
               fixesApplied.push(`failed to remove stranded lock directory ${lockDir}`);
             }
           }
--- a/src/resources/extensions/gsd/markdown-renderer.ts
+++ b/src/resources/extensions/gsd/markdown-renderer.ts
@@ -10,7 +10,7 @@
 // parseRoadmap(), parsePlan(), parseSummary() in files.ts.
 
 import { readFileSync, existsSync, mkdirSync, statSync } from "node:fs";
-import { createProjectionDirectorySync, removeProjectionFileSync } from "./atomic-write.js";
+import { createProjectionDirectorySync, relocatePhaseDirToCanonical, removeProjectionFileSync } from "./atomic-write.js";
 import { logWarning } from "./workflow-logger.js";
 import { isClosedStatus, isHiddenFromRoadmap, toStatus } from "./status-guards.js";
 import { isCanonicalStagedTaskSummaryState } from "./task-summary-projection-policy.js";
@@ -719,6 +719,9 @@ export async function renderRoadmapFromDb(
     return { skipped: "unplanned-milestone" };
   }
 
+  if (milestone.title) {
+    relocatePhaseDirToCanonical(basePath, milestoneId, milestone.title);
+  }
   const absPath = targetMilestoneFile(basePath, milestoneId, "ROADMAP", milestone.title);
   const artifactPath = toArtifactPath(absPath, basePath);
   const content = renderRoadmapMarkdown(milestone, slices);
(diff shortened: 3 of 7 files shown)


jeremymcs:
Closing as duplicate of #1892 for the same issue (#1526).

#1888 relocates the phase directory during ROADMAP render (`relocatePhaseDirToCanonical` in the projection write path). That side effect failed 34 unit tests (plan-milestone/plan-slice/reopen/authority paths looking for files under the old slug).

#1892 keeps the same two contracts with a smaller seam:
- title-change rename at `upsertMilestonePlanning` (the plan-milestone repro)
- `removeLockDirectory`: `rmSync` then `existsSync` then `rmdirSync`, loud error with the path

Keep #1892. Do not merge from this branch.

PR #1913, opened 2026-08-21
fix(issue): [Bug]: Windows verification stuck-loop — forward-slash pnpm.cmd path breaks cmd.exe + 30s exec default timeout kills next build

### LINKED ISSUE #1910: [Bug]: Windows verification stuck-loop — forward-slash pnpm.cmd path breaks cmd.exe + 30s exec default timeout kills next build
## Summary

On Windows, GSD's autonomous verification/build step fails in two independent ways that compound into a stuck-loop: (1) the verify command is built with a **forward-slash executable path** (`app/pnpm.cmd`) that cmd.exe cannot parse, yielding `'app' is not recognized`; and (2) `gsd_exec` kills the verify chain at the **30 s default timeout** even though `next build` alone exceeds it. Because verification never produces a clean pass, the same unit is re-dispatched repeatedly and the run ends in `artifact_db_status_divergence` with no guided recovery.

## Environment

- Platform: Windows 11 (cmd.exe execution environment)
- Package manager: pnpm v11.16.0 (monorepo with app in `app/` subdir)

## Symptom

Repeated dispatch of the same units within one auto session (forensics classified as `stuck-loop`):

- `execute-task/M001/S01/T02` — dispatched 2×
- `execute-task/M001/S02/T01` — dispatched 3×
- plus S02/T02, S02/T03, S02/T04, S03/T01

Every stuck unit shows the same verify-time stderr:

```
### ❌ `app/pnpm.cmd --dir app test` (exit code 1)
'app' is not recognized as an internal or external command,
operable program or batch file.
```

And, separately, a killed-but-valid build:

```
gsd_exec: exit=timeout duration=30143ms
```

The captured exec stdout confirms the build actually **succeeded** before it was killed on a later run (`Done in 227ms using pnpm v11.16.0`, full Next.js route table rendered) — so the workload is legitimate, just longer than 30 s end-to-end.

### Bug A — forward-slash executable path breaks on cmd.exe

The verification command is emitted as `app/pnpm.cmd --dir app test`. On Windows cmd.exe a forward-slash path in command position is parsed as the token `app` followed by arg `/pnpm.cmd`, so the shell reports `'app' is not recognized`. An explicitly-relative native invocation (`.\app\pnpm.cmd`) resolves correctly. The command builder does not emit a Windows-safe executable path.

- `package-manager.js:63-73` — `buildScriptCommand()` returns bare `<pm> <script>` with no OS-aware executable-path handling; the `app/pnpm.cmd`-style prefix that reaches cmd.exe is never normalized to a backslash / `.\`-relative form on win32.

### Bug B — 30 s default timeout kills verification/build workloads

`gsd_exec` clamps unspecified timeouts to the 30 s default, which is far below a `node --test` + `eslint` + `next build` + smoke chain.

- `exec-sandbox.js:22` — `defaultTimeoutMs: 30_000` in `EXEC_DEFAULTS`.
- `exec-sandbox.js:52-59` — `clampTimeout()` falls back to `opts.default_timeout_ms` when no `timeout_ms` is supplied.
- `tools/exec-tool.js:31,36` — `buildExecOptions()` sets `default_timeout_ms` from `cfg?.exec_timeout_ms ?? EXEC_DEFAULTS.defaultTimeoutMs`; the verification-oriented `verification_timeout_ms` pref (`auto-verification.js:137`, `preferences.js:772`) is **not** plumbed into the exec default for build/verify workloads.
- `verification-gate.js:761` — a timeout is surfaced with the same failure shape as a non-zero exit ("timed out after {limitMs}ms. Raise verification_timeout_ms…"), so a killed-but-valid build is treated as a hard verification failure.

### Consequence — no guided recovery from the resulting divergence

Because verification never cleanly passes, artifacts exist but the DB status diverges, and the recovery path treats it as fatal/non-fixable:

- `auto-recovery.js:168-175` — when a terminal abort is present the stuck-recovery returns `fatal: true` (`execute-task-recovery-aborted`), requiring manual `gsd_task_recovery_resume` / `gsd rebuild markdown` instead of auto-reconciling an exec-timeout-with-artifacts-present case.

## Suggested fix

1. **Windows-safe exec path (Bug A):** In `buildScriptCommand()` / the verify-command builder, on `process.platform === 'win32'` emit the executable with backslashes and an explicit `.\` prefix (e.g. `.\app\pnpm.cmd`) rather than `app/pnpm.cmd`, so cmd.exe parses it as a path, not a command token.
2. **Verify-aware timeout (Bug B):** When `gsd_exec` runs a verification/build workload, default its timeout to `verification_timeout_ms` (or a build-appropriate floor such as 300 s) instead of the 30 s `defaultTimeoutMs`. Plumb `verification_timeout_ms` into `buildExecOptions()` for these calls.
3. **Distinguish timeout from failure:** Surface `timed_out` distinctly from a non-zero exit in `verification-gate.js` so the completion gate does not treat a killed-but-valid build as a hard verification failure.
4. **Guided recovery (optional):** Add a `fixable`/guided path for `artifact_db_status_divergence` when the sole blocker was an exec timeout with artifacts present (`auto-recovery.js`).

## Workaround

1. Run `gsd_doctor` to confirm the divergence.
2. If a terminal abort is present, resume with `gsd_task_recovery_resume` using the recovery action id.
3. Clear the stale crash lock and resume with `/gsd auto` (completed work is preserved).

jeremymcs:
## Summary
- Fixed Windows verification command paths and routed configured verification timeouts to build and verification exec workloads.

## Bugs Addressed
- [x] **Windows cmd.exe cannot execute forward-slash package-manager path**
- [x] **Default 30-second exec timeout kills valid verification builds**

## Verification

## Related Issue
- Closes #1910
- [#1910 [Bug]: Windows verification stuck-loop — forward-slash pnpm.cmd path breaks cmd.exe + 30s exec default timeout kills next build](https://github.com/open-gsd/gsd-pi/issues/1910)

## User
- @affligem2001

## Repo
- `open-gsd/gsd-pi`

## Branch
- `issue/1910-bug-windows-verification-stuck-loop-forw-1787318665`

### DIFF
--- a/docs/user-docs/configuration.md
+++ b/docs/user-docs/configuration.md
@@ -691,7 +691,7 @@ verification_commands:
   - npm run test
 verification_auto_fix: true       # auto-retry on failure (default: true)
 verification_max_retries: 2       # max retry attempts (default: 2)
-verification_timeout_ms: 120000   # per-command spawn timeout (default: 120000)
+verification_timeout_ms: 120000   # host verification and verification-oriented gsd_exec default (default: 120000)
 ```
 
 | Field | Type | Default | Description |
@@ -699,9 +699,11 @@ verification_timeout_ms: 120000   # per-command spawn timeout (default: 120000)
 | `verification_commands` | string[] | `[]` | Simple executable commands to run after task execution |
 | `verification_auto_fix` | boolean | `true` | Auto-retry when verification fails |
 | `verification_max_retries` | number | `2` | Maximum auto-fix retry attempts |
-| `verification_timeout_ms` | number | `120000` | Per-command host-verification timeout in milliseconds. Unset keeps the 120s default. A timeout is reported as `failureClass: timeout`, never as exit 127. |
+| `verification_timeout_ms` | number | `120000` | Per-command host-verification timeout in milliseconds and the default timeout for verification-oriented `gsd_exec` workloads. Unset keeps the 120s host-verification default. A timeout is reported as `failureClass: timeout`, never as exit 127. |
 
 Verification commands must be simple executable commands. Shell piping (`|`) is supported, but logical OR (`||`) is rejected. GSD also rejects redirects (`>` and `<`), semicolons, backticks, and command substitution (`$(...)`) because verification is run as a controlled command list, not as an arbitrary shell program.
+
+For `gsd_exec`, verification-oriented workloads include builds, tests, linting, type checks, and verification commands. An explicit `context_mode.exec_timeout_ms` takes precedence over `verification_timeout_ms`; unrelated `gsd_exec` workloads keep the sandbox's 30-second default when `context_mode.exec_timeout_ms` is unset.
 
 For task-level `verify` commands (`taskPlanVerify`), GSD splits checks on newlines. `&&` chains stay within a single shell invocation, so commands such as `cd path && npm test` preserve directory context.
 
--- a/gitbook/configuration/preferences.md
+++ b/gitbook/configuration/preferences.md
@@ -177,8 +177,10 @@ verification_commands:
   - npm run test
 verification_auto_fix: true       # auto-retry on failure (default)
 verification_max_retries: 2       # max attempts (default: 2)
-verification_timeout_ms: 120000   # per-command spawn timeout (default: 120000)
+verification_timeout_ms: 120000   # host verification and verification-oriented gsd_exec default (default: 120000)
 ```
+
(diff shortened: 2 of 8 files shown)


jeremymcs:

Re-checked PR #1913 against [#1910 [Bug]: Windows verification stuck-loop — forward-slash pnpm.cmd path breaks cmd.exe + 30s exec default timeout kills next build](https://github.com/open-gsd/gsd-pi/issues/1910).

## Subtasks
- [x] **Windows verify commands use invalid forward-slash executable paths**
- [x] **Default 30-second timeout kills valid verification builds**

jeremymcs:

PR #1915, opened 2026-08-21
fix(issue): gsd quick lacks the right-sizing flags gsd-core has (--research/--validate/--discuss/--full)

### LINKED ISSUE #1553: gsd quick lacks the right-sizing flags gsd-core has (--research/--validate/--discuss/--full)
### Problem

`gsd quick` is documented as *"Lightweight task execution with GSD guarantees (atomic commits, state tracking) but without the full milestone/slice ceremony"* — the right home for off-lifecycle work.

But it accepts only `--squash`. There is no way to dial a one-off task up when it needs more rigour.

### Comparison with gsd-core

gsd-core's equivalent exposes a right-sizing dial:

| Flag | Effect |
|---|---|
| `--research` | investigate approaches before planning |
| `--validate` | plan-checking + post-execution verification |
| `--discuss` | surface design choices before planning |
| `--full` | all three |

Its guidance is explicit: use the plain form for typos and config, add flags "when unknowns exist, critical paths are involved, or design choices need locking in before execution."

gsd-pi has the lightweight end and not the dial.

### Why it matters

Off-lifecycle work is not uniformly trivial. A one-line config fix and a bug fix on a critical path both land in `quick`, and today they get identical treatment. Without the dial the practical choice is:

- run everything through `quick` and under-verify the risky ones, or
- promote real work to a full milestone purely to obtain verification, which is the ceremony `quick` exists to avoid

Both are bad, and the second one pushes teams toward doing off-lifecycle work *outside* GSD entirely — which loses the atomic commits and state tracking that make `quick` worth using at all.

### Requested

Bring gsd-pi's `quick` to parity with gsd-core's: `--research`, `--validate`, `--discuss`, `--full`, composable, defaulting to today's behaviour when none are passed.

If the flags mean something different in the Pi lifecycle, parity of *intent* matters more than parity of names — the point is a documented way to say "this one-off needs verification" without escalating it to a milestone.

jeremymcs:
## Summary
- Added composable quick-task rigor flags with documented behavior and passing focused tests and fast gates.

## Verification

## Related Issue
- Closes #1553
- [#1553 gsd quick lacks the right-sizing flags gsd-core has (--research/--validate/--discuss/--full)](https://github.com/open-gsd/gsd-pi/issues/1553)

## User
- @proofoftom

## Repo
- `open-gsd/gsd-pi`

## Branch
- `issue/1553-gsd-quick-lacks-the-right-sizing-flags-g-1787319323`

### DIFF
--- a/docs/user-docs/commands.md
+++ b/docs/user-docs/commands.md
@@ -8,7 +8,7 @@
 | `/gsd next` | Explicit step mode (same as `/gsd`) |
 | `/gsd auto` | Autonomous mode — research, plan, execute, commit, repeat |
 | `/gsd auto --resume-wedge <id>` | Acknowledge the named liveness wedge and re-enter auto mode after applying its sanctioned recovery |
-| `/gsd quick` | Execute a quick task with GSD guarantees (atomic commits, state tracking) without full planning overhead |
+| `/gsd quick [--discuss] [--research] [--validate] [--full] <task>` | Execute a quick task with GSD guarantees; optionally add discussion, research, plan checking, and post-execution verification (`--full` enables all three stages) |
 | `/gsd do <text>` | Route freeform text to the right GSD command |
 | `/gsd stop` | Stop auto mode gracefully |
 | `/gsd pause` | Pause auto-mode (preserves state, `/gsd auto` to resume) |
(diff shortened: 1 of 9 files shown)


jeremymcs:

Re-checked PR #1915 against [#1553 gsd quick lacks the right-sizing flags gsd-core has (--research/--validate/--discuss/--full)](https://github.com/open-gsd/gsd-pi/issues/1553).

## Subtasks
- [x] **gsd quick lacks the right-sizing flags gsd-core has (--research/--validate/--discuss/--full)**

PR #1916, opened 2026-08-21
fix(issue): quick has no non-interactive entry point, and 'gsd quick <task>' silently launches the console instead of erroring

### LINKED ISSUE #1554: quick has no non-interactive entry point, and 'gsd quick <task>' silently launches the console instead of erroring
### Summary

`/gsd quick` works. Verified on v1.11.0 by typing `/gsd quick <task>` into the TUI composer: it dispatched on a single Enter, created its branch and artifact directory, made the change and merged back correctly. The capability is sound.

The problem is entirely how it can be reached.

### 1. `gsd quick "<task>"` silently discards its arguments

`quick` is not a CLI subcommand. `gsd --help` lists `auto`, `headless`, `graph`, `hermes install`. So `gsd quick "<text>"` parses as bare `gsd` plus stray arguments, which launches the interactive console and drops the task text on the floor.

Observed under a real TTY: the console reached `Project Initialized` and idled at `0% 0/1.0M` — zero tokens consumed, empty composer, no model ever invoked. Nothing indicated that a task had been requested and ignored.

Without a controlling terminal, the same invocation emits:

```
[gsd] Error: Interactive mode requires a terminal (TTY).
[gsd] Non-interactive alternatives:
[gsd]   gsd auto ...
```

which reads as "quick needs a TTY" and is misleading — `quick` was never dispatched in either case; the *console* was. We spent two investigation rounds on that wrong hypothesis before reading `--help` carefully.

**Expected:** dispatch quick, or exit non-zero with `unknown command: quick — did you mean the /gsd quick slash command?`

### 2. There is no non-interactive entry point at all

- `gsd headless --help` → `auto`, `next`, `status`, `new-milestone`, `query`. No `quick`.
- No MCP tool exposes it: `grep -rhoE 'gsd_[a-z_]*quick[a-z_]*'` over `dist` returns nothing.

### Why it matters

The dividing line is **a controlling terminal**, not automation:

| | Can reach `/gsd quick` |
|---|---|
| Interactive user session | yes |
| An agent driving a tmux/herdr pane | yes |
| systemd timer, cron | **no** |
| CI runner | **no** |
| `gsd headless`, MCP client, `claude -p` | **no** |
| Any subprocess spawned on pipes | **no** |

So quick is not unreachable by automation — it is unreachable by *pipe-spawned* automation. An agent driving a terminal pane can use it, and we do: the probe behind this report was run by an automated agent that split a pane, became another user, launched the console and sent the slash command as synthetic keystrokes. It worked.

**The workaround existing makes the ask sharper, not weaker.** A tool whose only automation path is "drive a TUI with synthetic keystrokes" has no first-class automation path. That path requires a terminal multiplexer as a hard dependency, returns no exit code, produces no structured result, gives no way to distinguish "finished" from "still rendering" except scraping a token counter out of a status line, and is racy by construction — the caller must poll the screen to learn what happened. Every one of those is a property of screen-scraping, not of an API.

Concretely: in our deployment the timer that runs verification, the workers that execute queued tasks, and every orchestrator are all spawned on pipes. None can reach the one verb designed for exactly the work they need to do — lightweight tasks that should still carry atomic commits and a state record. The observed consequence was a worker doing off-lifecycle work by moving a git ref directly, producing no lifecycle record at all. Under quick it would have produced a commit and an artifact.

**Requested:** expose quick through `gsd headless quick "<task>"`, or as an MCP tool, or both — returning a structured result (task number, slug, branch, resulting commit sha) and a meaningful exit code.

### Environment

gsd-pi v1.11.0, Node 24, Linux, `claude-opus-4-8`.

### See also

Companion issue against `open-gsd/gsd-core` covering three behaviour defects in the shared `quick.js` (`md5 fc4ae2bc92c971c1117d825b16c7338f`, byte-identical in both packages): the "Quick Tasks Completed" STATE.md table is documented but never written, unrelated working-tree changes are swept into a `chore: auto-commit after quick-task` commit, and the summary artifact cites a commit sha that is not an ancestor of the main branch.

jeremymcs:
## Summary
- Added direct and headless quick-task execution with structured results, verified by 60 focused tests and syntax checks.

## Bugs Addressed
- [x] **CLI silently ignores `quick` and its task arguments**
- [x] **Quick tasks have no non-interactive entry point**

## Verification

## Related Issue
- Closes #1554
- [#1554 quick has no non-interactive entry point, and 'gsd quick <task>' silently launches the console instead of erroring](https://github.com/open-gsd/gsd-pi/issues/1554)

## User
- @proofoftom

## Repo
- `open-gsd/gsd-pi`

## Branch
- `issue/1554-quick-has-no-non-interactive-entry-point-1787319344`

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,6 +179,13 @@ Then use slash commands inside the GSD session:
 /gsd status
 ```
 
+For automation, quick tasks also have a non-interactive entry point with a
+structured result and meaningful exit code:
+
+```bash
+gsd quick --output-format json "Describe the task"
+```
+
 ## What GSD Pi Does
 
 - Plans work into milestones, slices, and tasks.
--- a/docs/user-docs/commands.md
+++ b/docs/user-docs/commands.md
@@ -422,6 +422,7 @@ The following commands are sent directly in your **Telegram chat** to a configur
 | `gsd remove <source>` | Remove a previously installed extension |
 | `gsd list` | List installed extensions |
 | `gsd graph <subcommand>` | Build, query, status, or diff the project knowledge graph built from `.gsd/` artifacts |
+| `gsd quick <task>` | Execute a quick task without a TUI (alias for `gsd headless quick <task>`) |
 | `gsd headless --json` | Structured JSONL event stream to stdout for scripting, CI, and troubleshooting (alias: `--output-format stream-json`) |
 | `gsd headless new-milestone` | Create a new milestone from a context file (headless — no TUI required) |
 
@@ -435,6 +436,9 @@ gsd headless
 
 # Run a single unit
 gsd headless next
+
+# Execute a quick task and return task, branch, artifact, commit, and exit details
+gsd quick --output-format json "fix the login button on mobile"
 
 # Instant JSON snapshot — no LLM, ~50ms
 gsd headless query
--- a/gsd-orchestrator/references/commands.md
+++ b/gsd-orchestrator/references/commands.md
@@ -47,6 +47,20 @@ Step mode — execute exactly one unit (task/slice/milestone step), then exit. R
 gsd headless --output-format json next
 ```
 
+### `quick <task>`
+
+Execute a lightweight task with GSD branch, artifact, and commit guarantees. The
+short form `gsd quick` is an alias for the headless command.
+
+```bash
+gsd quick --output-format json "fix the login button on mobile"
+gsd headless --output-format json quick "update the setup instructions"
+```
+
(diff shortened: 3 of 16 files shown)


jeremymcs:

Re-checked PR #1916 against [#1554 quick has no non-interactive entry point, and 'gsd quick <task>' silently launches the console instead of erroring](https://github.com/open-gsd/gsd-pi/issues/1554).

## Subtasks
- [x] **CLI silently ignores `gsd quick` arguments**
- [x] **Quick has no non-interactive entry point**

PR #1979, opened 2026-08-23
feat(gsd): session-start GitHub Copilot catalog refresh and runtime model activation

pimmink:
## Stacked submission — depends on #1978

**This PR depends on #1978 (live GitHub Copilot model-catalog sync). Do not merge until #1978 lands.**

Because a PR's base branch must exist on `open-gsd/gsd-pi` itself, this PR is opened with `base: main` (the same as #1978) rather than against #1978's fork-only branch. As a direct consequence, the diff below currently shows #1978's 18 commits plus this branch's own 2 commits — this is expected for a stacked PR and will shrink to just this branch's own commits once #1978 merges and this branch is rebased onto the new `main`.

## TL;DR

**What:** Adds an opt-in (default off) automatic refresh of the GitHub Copilot live model catalog at GSD session start, plus a 3-tier runtime model classification (trusted / manual-only / quarantined).

**Why:** #1978 made the live catalog inspectable via an explicit command; this closes the gap where a newly available model goes unnoticed for an entire session because nobody thought to run it.

**How:** A session-scoped coordinator (`copilot-catalog-session-refresh.ts`) reuses #1978's existing fetch/last-known-good pipeline, deduplicates concurrent triggers, never blocks startup, never triggers a login flow, and classifies each live-catalog model from its own `availability` / `execution` / `billing` / `conflicts` fields without ever fabricating capability or pricing data.

## What
- New `src/resources/extensions/gsd/copilot-catalog-session-refresh.ts`: config resolvers (`resolveCopilotCatalogRefreshMode`, `resolveCopilotCatalogNotifyOnChanges`, `resolveCopilotCatalogStaleAfterMs`), the pure `shouldTriggerCopilotCatalogRefresh` decision, the `startCopilotCatalogSessionRefresh` / `awaitCopilotCatalogSessionRefresh` coordinator pair (dedup + bounded timeout), `classifyRemoteCopilotModel` (3-tier classification), and `getLastKnownCopilotCatalogSnapshot`.
- Modified `preferences-types.ts` / `preferences-validation.ts`: new `copilot_catalog` preference block with full validation (enum, boolean, clamped numeric range, unknown-key warnings).
- Modified `bootstrap/register-hooks.ts`: wires the coordinator into `session_start` only (not `session_switch`), fire-and-forget.
[...]

PR #1980, opened 2026-08-23
feat(gsd): cheaper/better same-tier Copilot model suggestions and notifications

pimmink:
## Stacked submission — depends on #1978 and #1979

**This PR depends on #1978 (live catalog sync) and #1979 (session-start refresh). Do not merge until both land.**

Because a PR's base branch must exist on `open-gsd/gsd-pi` itself, this PR is opened with `base: main` rather than against #1979's fork-only branch. As a direct consequence, the diff below currently shows #1978's 18 commits + #1979's 2 commits + this branch's own 1 commit — expected for a stacked PR, and it will shrink to just this branch's own commit once both merge and this branch is rebased onto the new `main`. This is the last of the three; once it merges, all three features are fully upstream.

## TL;DR

**What:** Explicit cheaper-same-tier suggestion text in `/gsd copilot-models pricing`/`why`, plus a real non-blocking, deduplicated notification when the active GitHub Copilot model has a qualifying cheaper (or better-and-cheaper) same-tier alternative.

**Why:** Closes the "user never finds out they're on a needlessly expensive model" gap without any automatic switching or background routing changes.

**How:** A new `copilot-catalog-notifications.ts` reuses the existing `findCheaperSameTierOption` suggestion logic (already used by the `pricing`/`why` text), adds a pure capability-score comparison for "equivalent" vs. "better" messaging, and is wired to two lifecycle triggers: manual model selection (`model_select`, `source: set|cycle` only) and #1979's session-start refresh completion.

## What
- New `src/resources/extensions/gsd/copilot-catalog-notifications.ts`: `maybeNotifyCheaperAlternative` (fingerprint-deduped, provider-gated) and the pure `compareCapabilityScores` helper.
- Modified `commands/handlers/copilot-models.ts`: widens `findCheaperSameTierOption`, `findLocalModel`, and `localCopilotModels` from `ExtensionCommandContext` to `ExtensionContext` (structurally sufficient — both only ever touch `ctx.modelRegistry`), a non-breaking change for all existing callers.
- Modified `copilot-catalog-session-refresh.ts` (#1979): adds `getLastKnownCopilotCatalogSnapshot`.
- Modified `bootstrap/register-hooks.ts`: wires the `model_select` hook (excludes `source: "restore"`) and adds a post-refresh check inside the existing #1979 session-start trigger.
- Modified `docs/dev/ADR-012-provider-id-vs-api-shape.md` / `src/tests/provider-equality-allowlist.test.ts`: allowlists the new transport-specific `github-copilot` gate in `register-hooks.ts`.
- Retained from the original suggestion-text implementation: the bundled `claude-sonnet-5` pricing fallback (independently confirmed authoritative against `packages/pi-ai/src/models.generated.ts`) and its existing focused tests.

## Why
The suggestion-finding logic already existed and was already tested; this PR turns it into an actual proactive signal instead of requiring the user to run an explicit command to notice it.

## How
Extension-first. No new network calls of its own — the notification feature only ever reads the snapshot #1979's coordinator (or the explicit `/gsd copilot-models` command) already fetched; it never fetches independently.

## Architecture boundaries
- GitHub Copilot only. Advisory only: never mutates the active model, routing policy, or any persisted file.
- Never fires for automatic/dynamic routing selections, which do not emit the `model_select` event this feature listens to.
- Never fabricates capability or pricing data: an unknown profile always resolves to the safer "equal-or-lower capability" (cheaper equivalent) messaging, never a guessed "better" claim.
- Deduplicated by a fingerprint of (scope, selected model, suggested model, catalog revision) — a stable pairing is announced once; a material economics/catalog change allows a fresh notification.

## Change-type checklist
- [x] `feat` — New notification feature
- [x] `test` — 10 new tests (81 total for this feature area)
- [x] `docs` — ADR-012 allowlist update
- [ ] `fix`
- [ ] `refactor`
- [ ] `chore`

## Rollback notes
Fully additive: the new module and its two call sites can be removed with no schema/format changes elsewhere. The suggestion-text portion of `pricing`/`why` predates this PR and is unaffected by a rollback of just the notification wiring.

## AI-assistance disclosure
This PR was prepared with AI assistance (GitHub Copilot coding agent). All code was tested to the same standard as human-written code and is understood by the human maintainer of this fork.

### DIFF
--- a/docs/dev/ADR-012-provider-id-vs-api-shape.md
+++ b/docs/dev/ADR-012-provider-id-vs-api-shape.md
@@ -52,6 +52,7 @@ A small set of call sites legitimately keys on `provider`. These are **not** gat
 - **Display labels / onboarding copy** (`onboarding.ts`). Surface-only, no behavior impact.
 - **Live provider-account catalog state** (`copilot-model-catalog.ts`, `copilot-overlay-writer.ts`, `commands/handlers/copilot-models.ts`). GitHub Copilot's live `/models` sync, its remote-only-model quarantine/registration logic, and the `/gsd copilot-models` command all exist specifically to talk to the `github-copilot` transport's authenticated account state — the check is inherently transport-specific, not API-shape-specific.
 - **Session-start catalog refresh coordinator** (`copilot-catalog-session-refresh.ts`). GSD-W018's opt-in automatic refresh reuses the same GitHub Copilot auth-token resolution as the manual `/gsd copilot-models` command, for the same transport-specific reason.
+- **Session-start cheaper-alternative notification gate** (`bootstrap/register-hooks.ts`). GSD-W017's post-refresh notification only ever evaluates the active model when it is on the `github-copilot` transport, since the cheaper-alternative comparison itself is scoped to that account's live catalog and economics.
 
 These sites are enumerated in the allowlist at `src/tests/provider-equality-allowlist.test.ts`.
 
--- a/docs/dev/ADR-012-provider-id-vs-api-shape.md
+++ b/docs/dev/ADR-012-provider-id-vs-api-shape.md
@@ -51,6 +51,7 @@ A small set of call sites legitimately keys on `provider`. These are **not** gat
 - **Default-provider migrations** (`provider-migrations.ts`). Moving persisted settings from the `anthropic` transport to the `claude-code` transport is intentionally provider-id specific.
 - **Display labels / onboarding copy** (`onboarding.ts`). Surface-only, no behavior impact.
 - **Live provider-account catalog state** (`copilot-model-catalog.ts`, `copilot-overlay-writer.ts`, `commands/handlers/copilot-models.ts`). GitHub Copilot's live `/models` sync, its remote-only-model quarantine/registration logic, and the `/gsd copilot-models` command all exist specifically to talk to the `github-copilot` transport's authenticated account state — the check is inherently transport-specific, not API-shape-specific.
+- **Session-start catalog refresh coordinator** (`copilot-catalog-session-refresh.ts`). GSD-W018's opt-in automatic refresh reuses the same GitHub Copilot auth-token resolution as the manual `/gsd copilot-models` command, for the same transport-specific reason.
 
 These sites are enumerated in the allowlist at `src/tests/provider-equality-allowlist.test.ts`.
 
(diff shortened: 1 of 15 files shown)


pimmink:
Rebased onto the latest `feat/copilot-catalog-session-refresh` (PR #1979) tip. Unlike the previous rebases in this stack, this one produced a real merge conflict in `copilot-models.ts` (8 hunks) because #1978's review-driven fixes touched the same functions this branch's feature also touches (`localCopilotModels`, `findLocalModel`, `resolveEconomicsForModel`, `formatPricingRecord`). Resolved by combining both sides: kept #1978's `Model<Api>`/type-safety fixes while preserving this branch's new parameters and call sites (`cheaperSameTierOption`, `snapshot`).

The rebase also silently broke two call sites inside `findCheaperSameTierOption` that called `resolveEconomicsForModel` with #1978's *old* 3-argument signature — no conflict marker, since it was a wholly-new hunk with nothing to conflict against; caught only by `tsc`. Fixing that surfaced a genuine type mismatch: `findCheaperSameTierOption` only ever has `ExtensionContext` available (it's invoked from the `model_select` hook, not a command handler), so it can't pass `ctx` into a parameter typed `ExtensionCommandContext`. Relaxed `readGithubCopilotModelsJsonCost` / `buildUserOverrideEconomics` / `resolveEconomicsForModel`'s `ctx` parameter to `ExtensionContext` — they only ever touch `ctx.modelRegistry`, unchanged on the narrower type, so every existing caller stays valid.

Full-suite verification after the rebase then surfaced two **pre-existing** test-fixture bugs, confirmed present even on the original pre-rebase commit (checked out from a local backup for comparison):
1. `copilot-models-handler.test.ts`'s two cheaper-same-tier tests set `Model<Api>.cost` using $-per-1K units instead of the actual $-per-1M convention (`tokenPricesFromCost` divides by 1000) — they were silently relying on the bare-ID bundled-cost-table implicit fallback instead of the fixture's own intended cost, and broke once #1978's `disableImplicitFallback: true` was inherited.
2. `copilot-catalog-notifications.test.ts`'s `createFakeCtx` never supplied any cost data at all, so 5 of its 10 tests were failing (via early-return `false`, no exception) even before this rebase.

Both fixed with correctly-scaled explicit cost fixtures.

Verified with `tsc --noEmit` (clean) and the full affected test surface: 269/269 passing across `copilot-catalog-notifications`, `copilot-models-handler`, `model-cost-table`, `copilot-catalog-session-refresh`, `provider-equality-allowlist`, `oauth-api-model-routing`, `copilot-overlay-writer`, `model-router`, `copilot-model-catalog`.

Latest commit: `2a936c03`. The full stack (#1978 → #1979 → #1980) is now consistent and green locally.

jeremymcs:
Superseded by #2060 — this slice (cheaper/better same-tier notifications, GSD-W017) landed on main via a fresh cherry-pick on top of the landed session-refresh (#2058). Commit authorship preserved. Thanks for the groundwork.

PR #2011, opened 2026-08-25
fix(issue): [Bug]: Extensions installed via `/gsd extensions install` are registered but never loaded at runtime

### LINKED ISSUE #2005: [Bug]: Extensions installed via `/gsd extensions install` are registered but never loaded at runtime
### Bug type

Incorrect behavior or output

### Summary

Extensions installed via `/gsd extensions install` are correctly registered (registry entry, files on disk) but never actually loaded at the next GSD startup, because the resolver never merges the installed-extensions directory into the loaded extension set.

### Steps to reproduce

1. Inside a `gsd` session, run `/gsd extensions install npm:<any package that exposes an extension-manifest.json>`.
2. Confirm the command reports `Installed "<id>" v<version>. Restart GSD to activate.`, and that `~/.gsd/extensions/<id>/` and a `registry.json` entry with `source:"user"` exist.
3. Restart `gsd` (or start a new session).
4. Try to use a command or tool the extension declares in its manifest `provides` block (e.g. its slash command).

### Expected behavior

After restart, the extension's command/tool is available and usable, exactly as it would be for a bundled extension.

### Actual behavior

The command/tool is never registered. The extension shows as `enabled` in `/gsd extensions list`, but it is functionally inert — nothing it provides is reachable.

### GSD version

1.16.1

### Node.js version

v24.18.0

### Operating system

Debian 13.6

### Install method

npm global install

### Affected area

MCP / tools / extensions

### Model and provider

_No response_

```shell
`DefaultPackageManager.resolve()` (`packages/pi-coding-agent/src/core/package-manager.ts`) builds the extensions list from exactly two sources: `settings.json`'s `packages`/`extensions` arrays, and auto-discovery of `<agentDir>/extensions` (`~/.gsd/agent/extensions/`). The directory where `/gsd extensions install` actually writes files — `~/.gsd/extensions/<id>/`, returned by `getInstalledExtDir()` in `commands-extensions.ts` — is never included in either source.

A dedicated merge helper already exists for exactly this purpose: `mergeExtensionEntryPaths(bundledPaths, installedExtDir)` in `src/extension-discovery.ts`, with a design comment stating: "Installed extensions with the same manifest ID as a bundled extension take precedence (D-14). Loader stays dumb — receives a pre-merged path list (D-15)." It has zero call sites outside its own unit test — `grep -rn "mergeExtensionEntryPaths(" --include="*.ts"` across the repository returns only the definition and the test file.

Reproduced directly by calling `DefaultPackageManager.resolve()` right after a successful `/gsd extensions install`: the installed extension's path is absent from `resolved.extensions`.
```

### Impact

Severity: `/gsd extensions install` — the documented, TUI-native way to install third-party extensions — is non-functional for its core purpose.
Frequency: 100% reproducible, every installation through this path.
Consequence: users get a false "Installed... Restart GSD to activate" success message, then silently get nothing after restart, with no error explaining why.

### Additional information

A workaround exists: installing the same package via the shell command `gsd install npm:<pkg>` does load and activate the extension correctly, because it resolves through a different code path (`settings.json`'s `packages` array via `resolvePackageSources()`). That path has its own, separate bookkeeping gaps.

jeremymcs:
## Summary
- Fixed runtime loading of installed extensions and verified discovery, precedence, registry lookup, sorting, and package boundaries.

## Verification

## Related Issue
- Closes #2005
- [#2005 [Bug]: Extensions installed via `/gsd extensions install` are registered but never loaded at runtime](https://github.com/open-gsd/gsd-pi/issues/2005)

## User
- @efrembaraldo

## Repo
- `open-gsd/gsd-pi`

## Branch
- `issue/2005-bug-extensions-installed-via-gsd-extensi-1787680153`


jeremymcs:

Re-checked PR #2011 against [#2005 [Bug]: Extensions installed via `/gsd extensions install` are registered but never loaded at runtime](https://github.com/open-gsd/gsd-pi/issues/2005).

## Subtasks
- [x] **[Bug]: Extensions installed via `/gsd extensions install` are registered but never loaded at runtime**

PR #2012, opened 2026-08-25
fix(issue): [Bug]: `gsd install` (shell) never writes an extensions registry entry, making installed extensions invisible to `/gsd extensions list`/`update`

### LINKED ISSUE #2006: [Bug]: `gsd install` (shell) never writes an extensions registry entry, making installed extensions invisible to `/gsd extensions list`/`update`
### Bug type

Incorrect behavior or output

### Summary

`gsd install npm:<pkg>` successfully installs and loads a gsd-pi extension package, but never writes an entry to the extensions registry (`~/.gsd/extensions/registry.json`), so the extension is invisible to `/gsd extensions list` and never eligible for `/gsd extensions update`.

### Steps to reproduce

1. From a shell, run `gsd install npm:<package that exposes an extension-manifest.json>`.
2. Confirm the package installs successfully under `~/.gsd/agent/npm/node_modules/<pkg>/` and is added to `settings.json`'s `packages` array.
3. Start a `gsd` session and confirm the extension's command/tool works (it loads correctly through `settings.json`'s `packages` resolution).
4. Inside the session, run `/gsd extensions list`.

### Expected behavior

The extension appears in the list output, tagged with its install source and version, and is eligible for `/gsd extensions update`.

### Actual behavior

The extension does not appear at all. If a later auto-discovery pass happens to register it independently, it appears without `source:"user"`/`version`, and is therefore excluded from `/gsd extensions update`, which filters strictly on `source === "user"`.

### GSD version

1.16.1

### Node.js version

v24.18.0

### Operating system

Debian 13.6

### Install method

npm global install

### Affected area

MCP / tools / extensions

### Model and provider

_No response_

```shell
`gsd install` (`packages/pi-coding-agent/src/core/package-commands.ts` → `package-manager.ts`) has no awareness of the `extension-manifest.json` convention used by gsd-pi extensions — it treats every installed package as a generic npm dependency and only ever writes to `settings.json`'s `packages` array. It never touches `~/.gsd/extensions/registry.json`, which is instead written exclusively by `/gsd extensions install`'s `writeInstalledRegistryEntry()` (`commands-extensions.ts`).
```

### Impact

Severity: moderate — the extension works, but goes silently stale: users who installed it via `gsd install` are never notified of updates and cannot see or manage it via `/gsd extensions`.
Frequency: 100% reproducible for any extension installed through this path.
Consequence: loss of update visibility and management for third-party extensions installed via the shell CLI.

### Additional information

`/gsd extensions install` (the TUI slash-command counterpart) has the mirror-image gap: it writes the registry entry correctly, but the installed extension is never loaded at runtime — see #2007 . A good follow-up improvement would be unifying the two installation mechanisms so a single code path produces both a working load and a correct, mutually-visible registry entry.

jeremymcs:
## Summary
- Fixed registry convergence and runtime loading for both extension installation paths with focused regression coverage.

## Bugs Addressed
- [x] **CLI install omits extensions registry entry**
- [x] **Slash-command-installed extensions do not load at runtime**

## Verification

## Related Issue
- Closes #2006
- [#2006 [Bug]: `gsd install` (shell) never writes an extensions registry entry, making installed extensions invisible to `/gsd extensions list`/`update`](https://github.com/open-gsd/gsd-pi/issues/2006)

## User
- @efrembaraldo

## Repo
- `open-gsd/gsd-pi`

## Branch
- `issue/2006-bug-gsd-install-shell-never-writes-an-ex-1787682194`

### DIFF
--- a/packages/pi-coding-agent/src/core/resource-loader.ts
+++ b/packages/pi-coding-agent/src/core/resource-loader.ts
@@ -156,6 +156,10 @@ export interface DefaultResourceLoaderOptions {
 	};
 	systemPromptOverride?: (base: string | undefined) => string | undefined;
 	appendSystemPromptOverride?: (base: string[]) => string[];
+	extensionPathsTransform?: (paths: string[]) => {
+		paths: string[];
+		diagnostics?: string[];
+	};
 }
 
 export class DefaultResourceLoader implements ResourceLoader {
@@ -194,6 +198,7 @@ export class DefaultResourceLoader implements ResourceLoader {
 	};
 	private systemPromptOverride?: (base: string | undefined) => string | undefined;
 	private appendSystemPromptOverride?: (base: string[]) => string[];
+	private extensionPathsTransform?: DefaultResourceLoaderOptions["extensionPathsTransform"];
 
 	private extensionsResult: LoadExtensionsResult;
 	private skills: Skill[];
@@ -242,6 +247,7 @@ export class DefaultResourceLoader implements ResourceLoader {
 		this.agentsFilesOverride = options.agentsFilesOverride;
 		this.systemPromptOverride = options.systemPromptOverride;
 		this.appendSystemPromptOverride = options.appendSystemPromptOverride;
+		this.extensionPathsTransform = options.extensionPathsTransform;
 
 		this.extensionsResult = { extensions: [], errors: [], runtime: createExtensionRuntime() };
 		this.skills = [];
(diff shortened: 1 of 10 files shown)


jeremymcs:

Re-checked PR #2012 against [#2006 [Bug]: `gsd install` (shell) never writes an extensions registry entry, making installed extensions invisible to `/gsd extensions list`/`update`](https://github.com/open-gsd/gsd-pi/issues/2006).

## Subtasks
- [x] **Shell install does not register extensions**
- [x] **Slash-command install does not load extensions**

PR #2017, opened 2026-08-26
fix(gsd): prevent UAT schema errors from poisoning retries

### LINKED ISSUE #1993: [Bug]: run-uat harnessAbort poisoning blocks gsd_uat_result_save and the promised rerun never dispatches (finalize-retry wedge); intermittent missing workflow tool surface masked by ToolSearch rewrite
## Problem

A `run-uat` unit that completes all its checks can still lose its entire UAT result and wedge auto-mode. Two triggers, one converging tail: (B) a single schema-invalid `gsd_uat_exec` call poisons the unit via `harnessAbort`, after which every `gsd_uat_result_save` is rejected — while the rejection message promises "auto-mode can rerun it", a rerun path that does not exist (finalize-retry only re-verifies the missing artifact and the liveness backstop stops auto); (A) intermittently the run-uat session spawns without the workflow tool surface at all, and the session layer's unknown-tool → `ToolSearch` rewrite hides the known "tool not registered" signature so the existing bounded-retry path never engages. Observed 9 times in 3 days on one project (waves of 4-slice milestones), each costing an operator repair plus a full auto restart (~15–20 min).

## Root Cause (Mode B — deterministic, fully traced)

1. The unit's first `gsd_uat_exec` omits required args (`milestoneId`, `sliceId`, `checkId`). The tool returns a deterministic schema error: `Validation failed for tool "gsd_uat_exec": - milestoneId: must have required properties milestoneId, sliceId, checkId`.
2. `isRetryableHarnessToolError` (`bootstrap/register-hooks.js:521-539`) excludes `isDeterministicPolicyError` — but that only covers `HARD BLOCK:`-style policy texts (`auto-tool-tracking.js:120` `DETERMINISTIC_POLICY_ERROR_RE`). It then falls through to `isToolInvocationError` (`auto-tool-tracking.js:135-139`), whose `TOOL_INVOCATION_ERROR_RE` (`auto-tool-tracking.js:119`) explicitly matches `Validation failed for tool` → classified retryable-harness → `recordCurrentUnitHarnessAbort({kind:"tool-error", ...})` (`bootstrap/register-hooks.js:53-57, 541-547`) writes `harnessAbort` into the unit runtime record.
3. `unit-runtime.js:100-135` (`writeUnitRuntimeRecord`) preserves `harnessAbort` for the whole unit run (`sameRun`). Successful later calls do NOT clear it (only update `lastProgressAt`).
4. `tools/workflow-tool-executors.js:88-101` then rejects **every** subsequent `gsd_uat_result_save` in that run:

   ```text
   Retryable harness abort: run-uat {unit} was truncated by tool-error while calling gsd_uat_exec.
   This gate result was not saved; leave the gate pending so auto-mode can rerun it.
   ```

   (`details.error = "harness_aborted_needs_retry"`, `retryable: true`)
5. The worker complies and retries the save once — rejected identically — and surrenders. In the traced incident the worker had 13+ **successful** `gsd_uat_exec` calls after the one invalid call and produced a complete 10-check PASS result that was thrown away.
6. The promised rerun is unimplemented: finalize re-runs artifact verification on identical inputs; `auto-post-unit.js:982` reports `Artifact verification failed: … ASSESSMENT.md was not found on disk after unit execution`; two identical failures trip the ADR-047 liveness backstop → `finalize-retry` sanctioned exit (`auto/loop-sanctioned-exits.js:36`) whose text tells the **operator** to re-run. `clearUnitRuntimeRecord` is only called by guided-flow self-heal (`guided-flow.js:1331-1337`); a fresh unit dispatch would clear the poison (new `startedAt` → `sameRun=false`, `unit-runtime.js:100-135`) but no code path performs it.

Note: the #1714/#1761 verify-after-write guard (`auto-post-unit.js:1726-1740`, `missingDurableSaveReceipt`) does not cover this shape — it only runs when artifact verification *succeeded* (silent-pass detector). Here the artifact never lands, so the failure surfaces earlier as the generic wedge.

## Root Cause (Mode A — intermittent 3/9, race + masking)

`auto-tool-tracking.js:174-181` documents a known registration race: "the workflow MCP server registers its tool surface asynchronously after session start, so a Unit's first tool call can race the registration. Callers should retry (bounded)". `TOOL_UNAVAILABLE_ERROR_RE` (`auto-tool-tracking.js:130`) matches `No such tool available|TOOL_SURFACE_NOT_READY` to drive that bounded retry. However, in three sessions the missing surface never produced that signature: the model's calls for workflow tools arrived at the session as `ToolSearch` probes (unknown-tool rewrite at the session layer), and the GSD `ToolSearch` stub answered `"ToolSearch is not available in GSD. Call the workflow tool you need directly by name."` — which matches none of the retry signatures. The worker dead-ends (observed: 10–16 `ToolSearch`-only calls, zero workflow calls, honest "UNIT FAILED" final text, no failure signal to the orchestrator), and the unit tails into the same finalize-retry wedge. Root of the intermittency needs upstream reproduction; the observable is the empty workflow-tool surface in exactly those sessions while sibling units in the same waves had the full surface.

## Expected Behavior

1. **Do not poison the unit for model-fixable schema errors.** In `isRetryableHarnessToolError` (`bootstrap/register-hooks.js`), exclude tool-schema validation errors (e.g. `Validation failed for tool …` / `must have required properties …`) from setting `harnessAbort` — same rationale as the existing `isDeterministicPolicyError` exclusion. In every observed trace the model self-corrects on the very next call.
2. **Heal stale aborts.** A successful `gsd_uat_exec`/`gsd_exec` result after a recorded `tool-error` abort should clear (or downgrade) `harnessAbort` in the unit runtime record instead of preserving it for the whole run — the abort is provably stale once the same tool works again.
3. **Implement the promised rerun.** When finalize fails for a `run-uat`/`validate-milestone` unit whose runtime record carries `harnessAbort`, re-dispatch the unit (fresh `startedAt` clears the poison) instead of re-verifying the artifact twice and wedging; the rejection message already advertises this behavior.
4. **Assert the tool surface at dispatch (Mode A).** At run-uat dispatch, assert the required workflow tool names are present in the child session's tool schema; if not, fail the dispatch loudly and retry (the bounded-retry machinery for `TOOL_SURFACE_NOT_READY` exists) rather than letting the worker dead-end on the `ToolSearch` stub. Also consider matching the `ToolSearch` stub reply as a tool-unavailable signal, or disabling the unknown-tool → `ToolSearch` rewrite inside auto-mode units.

## Environment

- GSD: gsd-pi 1.16.1 dist loader + live agent overlay (contains the #1714/#1761 `missingDurableSaveReceipt` guard, so post-that-fix runtime)
- Mode: `gsd headless auto`, milestone waves (research → plan → gates → tasks → complete-slice → run-uat → validate)
- Model: execution routing per GSD preferences (3rd-party OpenAI-compatible provider)
- Units: 9 `run-uat` wedges across 5 milestones over 3 days; 3 further `run-uat` units in the same waves passed normally

## Reproduction Context

- Mode B (deterministic): dispatch `run-uat`; let the worker's **first** `gsd_uat_exec` omit required `checkId`/`milestoneId`/`sliceId`, then complete all checks and call `gsd_uat_result_save` → save rejected with `harness_aborted_needs_retry`; unit ends; finalize verifies the missing `ASSESSMENT.md` twice → liveness backstop → `finalize-retry` wedge.
- Mode A (intermittent, 3/47 run-uat sessions): session starts without workflow tools; all workflow-tool calls surface as `ToolSearch` probes answered by the stub.

## Forensic Evidence

- 9 journal `auto-exit` events, identical tail: `finalize-retry recurred 2x … Artifact verification failed: … ASSESSMENT.md was not found on disk after unit execution (UAT assessment result)` (units/wedges: S01 W-875f30cb, S01 W-0f64e107, S01 W-a4c28613, S03 W-89f73a54, S04 W-b84e1f57, S01 W-ba4b8a72, S02 W-4a79a7c2, S04 W-4f19c039, S04 W-821d8bf0).
- Tool-call census of the failing vs passing `run-uat` activity logs: Mode B failures made 7–16 successful `gsd_uat_exec` calls + 1–2 rejected `gsd_uat_result_save`; passing units in the same waves show the identical shape minus the invalid first call; Mode A failures show only `ToolSearch`×10–16 with the stub reply quoted above.
- Traced rejection payload (Mode B): `details.error="harness_aborted_needs_retry"`, `details.harnessAbortReason="Validation failed for tool \"gsd_uat_exec\": - milestoneId: must have required properties milestoneId, sliceId, checkId"`.

Related: #1714, #1761 (fixed the silent-pass shape via verify-after-write — does not cover save-rejected/never-callable shapes), #1128 / PR #1136 (introduced the harness-abort rejection for truncated evaluations — this issue reports the missing recovery half).

---

pimmink:
## TL;DR

**What:** Prevents model-fixable `gsd_uat_exec` schema validation errors from poisoning a `run-uat` unit with a durable `harnessAbort`, and clears stale `tool-error` abort state after a later successful execution.
**Why:** A single malformed UAT execution could block every subsequent `gsd_uat_result_save`, discard an otherwise complete UAT result, and route the unit into a misleading finalize-retry wedge.
**How:** Separates schema-validation failures from genuine harness truncation signals and adds an explicit, locked runtime-state clear for stale `tool-error` aborts only upon subsequent successful `gsd_exec` or `gsd_uat_exec`.

## What

This PR addresses the deterministic Mode B path from issue #1993:

- schema-invalid `gsd_uat_exec` no longer records `harnessAbort`;
- a successful `gsd_uat_exec` or `gsd_exec` clears a same-run `tool-error` abort (including prior deadline timeouts, as subsequent successful execution supersedes stale failure evidence);
- cross-tool execution success (`gsd_exec` following `gsd_uat_exec` or vice-versa) heals a same-run `tool-error` abort;
- genuine `turn-abort` (e.g. model context truncation or interrupted turns) remains protected and cannot be cleared by tool execution;
- existing tool-surface readiness preflight and finalize retry dispatch behavior remain unchanged.

## Why

Previously, `Validation failed for tool`, `Input validation error`, `Invalid arguments for tool`, and MCP `-32602` failures were included in the generic retryable invocation matcher used by harness-abort recording. A model-correctable schema mistake therefore became indistinguishable from a truncated harness execution. The persisted marker then caused `gsd_uat_result_save` to be rejected for the remainder of the same run.

## How

The shared `isToolInvocationError` classifier remains unchanged for its existing recovery consumers. A narrower `isToolSchemaValidationError` predicate is applied only when deciding whether `gsd_exec`/`gsd_uat_exec` should record a harness abort. The runtime layer exposes an explicit locked clear operation, and hooks invoke it for successful execution of those two tools when the active same-run marker is specifically `kind: "tool-error"`.

A later successful execution proves the tool surface is functioning, superseding prior transient execution or deadline errors. Turn aborts (`kind: "turn-abort"`) are not cleared.

Closes #1993

## Verification

- Rebased onto latest `upstream/main`.
- `pnpm run test:compile` & `register-hooks-loop-guard-auto.test.js`: **16/16 passed** (including new timeout-heal and cross-tool-heal tests).
- `pnpm run verify:fast`: **all checks passed ✓** (strict matrix, package boundaries, actionlint).
- Commit: `53817970`.

AI-assisted contribution: investigation and implementation are AI-assisted; the human contributor remains responsible for the design and implementation.


jeremymcs:
The root-cause fix is well-targeted and I confirmed the poisoning mechanism is real (`blockIfHarnessAbortedUnit` gates `save_uat_result`/`save_gate_result`; one schema mistake durably blocked subsequent saves). One design question has to be settled before this leaves draft:

**The stated invariant doesn't match the clear path.** The description says genuine timeouts / forced-resolution aborts "remain protected", but exec timeouts record `kind: "tool-error"` aborts (the existing deadline test asserts this), and the new clear removes **any** `tool-error` abort after a later successful exec — so a genuine timeout abort *is* healed by a subsequent successful `gsd_uat_exec`. That healing is arguably correct (a later successful exec supersedes stale evidence; saves require fresh results anyway), but we need to pick one:

1. keep the broad clear and correct the description + add an explicit timeout→heal test, or
2. narrow the clear to aborts whose reason matches a tool-surface/schema failure.

Option 1 is my lean. Also: cross-tool healing (`gsd_exec` success clearing an abort recorded for `gsd_uat_exec`) looks intentional but is untested — worth one test either way.

The branch is ~a week behind main and draft; once the above is settled, please rebase and run `pnpm run verify:fast` (the touching areas are covered by the compiled unit suite, so `pnpm run test:unit:compiled` locally before push is the safer gate).

pimmink:
@jeremymcs Thanks for the review!

Following Option 1:
1. **Updated PR description & invariants:** Stated explicitly that subsequent successful execution (`gsd_exec` or `gsd_uat_exec`) supersedes stale `tool-error` aborts (including prior deadline timeouts) because fresh successful execution proves tool-surface viability. Turn aborts (`kind: "turn-abort"`) remain protected.
2. **Added regression tests:**
   - Explicit timeout → heal test (`register-hooks heals an exec deadline timeout abort after subsequent successful execution`)
   - Cross-tool healing test (`register-hooks allows cross-tool execution success (gsd_exec) to heal a gsd_uat_exec abort`)
3. **Rebased & verified:**
   - Rebased onto latest `upstream/main` (`e34e7d7d0`)
   - `register-hooks-loop-guard-auto.test.js`: **16/16 passed**
   - `pnpm run verify:fast`: **all checks passed ✓**

Marked ready for review.

pimmink:
Follow-up to the review comments regarding abort-state integrity:

1. **Atomic conditional clear under `withRecordLock`:**
   - `clearUnitHarnessAbort` now accepts `expectedKind?: UnitHarnessAbortRecord["kind"]` (e.g. `"tool-error"`).
   - The kind verification and `startedAt` check are evaluated atomically inside `withRecordLock` before clearing `harnessAbort`, eliminating the un-locked check window.
2. **`turn-abort` precedence protection:**
   - `recordUnitHarnessAbort` will never overwrite an existing `turn-abort` with a subsequent `tool-error` in the same run.
   - Subsequent `gsd_exec`/`gsd_uat_exec` execution success will only heal `tool-error` aborts and will not clear a genuine `turn-abort`.
3. **Verified:**
   - Added `register-hooks preserves a turn-abort and does not overwrite it with a subsequent tool-error or clear it on tool success` to `register-hooks-loop-guard-auto.test.ts` (**17/17 passed**).
   - `pnpm run verify:fast`: **all checks passed ✓**.

PR #2030, opened 2026-08-28
fix(issue): [Bug]: initResources() never inspects the destination, so a torn ~/.gsd/agent/ copy is never repaired — and hasMissingBundledResourceFiles() has no caller

### LINKED ISSUE #2028: [Bug]: initResources() never inspects the destination, so a torn ~/.gsd/agent/ copy is never repaired — and hasMissingBundledResourceFiles() has no caller
### Bug type

Managed resource sync / extension staleness

### Summary

`initResources()` decides whether to re-sync `~/.gsd/agent/` **entirely from source-side state** — the bundled version plus a fingerprint of `dist/resources/`. It never inspects the destination. Once the managed extension copy under `~/.gsd/agent/extensions/` drifts (a file deleted, truncated, hand-edited, quarantined by AV, or left half-written by an interrupted copy), that drift is **permanent for the lifetime of the release**: every subsequent launch takes the early return and the broken copy is what actually executes.

The package already ships a destination-drift detector for exactly this — and it has no caller.

### Steps to reproduce

1. Launch `gsd` once so `~/.gsd/agent/` is populated and `managed-resources.json` is stamped.
2. Populate a scratch agent dir, then simulate drift in it:

```bash
node --input-type=module -e "
const m = await import('file:///<npm-global-root>/@opengsd/gsd-pi/dist/resource-loader.js');
m.initResources('<tmp>/agent-test', '<tmp>/skills-alt');
"
rm '<tmp>/agent-test/extensions/gsd/db/writers/liveness-backstop.js'
printf 'CORRUPTED\n' > '<tmp>/agent-test/extensions/gsd/auto-liveness-backstop.js'
```

3. Re-run `initResources()` against the same agent dir.
4. Ask the package's own detector about the same directory pair:

```bash
node --input-type=module -e "
const m = await import('file:///<npm-global-root>/@opengsd/gsd-pi/dist/resource-loader.js');
console.log(m.hasMissingBundledResourceFiles('<tmp>/agent-test/extensions',
  '<npm-global-root>/@opengsd/gsd-pi/dist/resources/extensions'));
"
```

### Expected behavior

The second `initResources()` restores the deleted file and overwrites the corrupted one. The managed copy is a faithful mirror of the bundle whenever the CLI starts.

### Actual behavior

Nothing is repaired. The deleted file is still missing and the corrupted file is still 10 bytes. `hasMissingBundledResourceFiles()` returns `true` for the exact directory pair `initResources()` just declared up to date.

The cause, `dist/resource-loader.js:705-711` (line numbers from 1.16.2):

```js
if (manifest && isCurrentPackageManifest(manifest) && manifest.gsdVersion === currentVersion) {
    // Version matches — check content fingerprint for same-version staleness.
    const currentHash = getCurrentResourceFingerprint();
    if (manifest.contentHash && manifest.contentHash === currentHash) {
        return;
    }
}
```

`getCurrentResourceFingerprint()` (`dist/resource-loader.js:149`) reads the precomputed hash shipped inside `dist/resources/`, falling back to hashing that same source tree. Both branches describe the **source**. The manifest it is compared against was written at the last successful sync. So the condition answers "has the bundle changed since we last synced?" — never "does the destination still match the bundle?"

`hasMissingBundledResourceFiles(destDir, sourceDir)` is defined at `dist/resource-loader.js:771` and declared at `dist/resource-loader.d.ts:74`. Searching the entire `dist/` tree for the identifier returns only those two lines: **it is exported and never called.** A destination-drift check exists and is simply not wired in.

### Expected fix

Call the existing detector before taking the early return — it already has the right signature and the right semantics:

```js
if (manifest && isCurrentPackageManifest(manifest) && manifest.gsdVersion === currentVersion) {
    const currentHash = getCurrentResourceFingerprint();
    if (manifest.contentHash && manifest.contentHash === currentHash
        && !hasMissingBundledResourceFiles(join(agentDir, 'extensions'),
                                           join(resourcesDir, 'extensions'))) {
        return;
    }
}
```

A full destination re-hash on every launch would be too costly, but the existence check is cheap and catches the failure mode that actually bites (missing / truncated files). Note it will not catch a *same-size, same-name* hand edit; if that matters, a destination fingerprint stored in the manifest would.

### Why this matters — the failure it produces is very hard to diagnose

The observable symptom is an ESM load error naming a symbol that does not exist in the release, e.g.:

```
Extension load error: ... does not provide an export named 'someSymbol'
```

That reads exactly like an upstream regression, so the natural response is to go looking for a removed export — which does not exist, because the bundle is fine and only the executing copy is torn. This has a documented precedent: a downstream project carried two local patch passes for months on the premise that gsd-pi 1.15.0 had removed `observeExternalMarkdownEdits` / `observeExternalPlanningEdits` / `recordCompatProjectionWrite`; verification against a clean `npm pack @opengsd/gsd-pi@1.15.0` showed all three *did* exist and the installed copies were simply stale. A clean reinstall resolved it. The mechanism above is what allows that state to persist.

### Related, and worth considering together

The sync is strictly one-directional (`dist/resources/` → `~/.gsd/agent/`), and `syncResourceDir` (`dist/resource-loader.js:356`) deletes destination subdirectories before re-copying. That is defensible as a design choice, but combined with the above it means the copy that actually executes is both un-verified and silently replaceable. A user who edits `~/.gsd/agent/extensions/` to debug something gets no warning that their edits are load-bearing until a version bump erases them.

### GSD version

1.16.2 (verified against a clean `npm pack @opengsd/gsd-pi@1.16.2`; also present in 1.16.1)

### Node.js version

22.x

### Operating system

Windows 11

### Install method

### Affected area

resource-loader, extension loading, `gsd update`

jeremymcs:
## Summary
- Repaired torn managed extension copies by checking destination completeness before skipping sync, with regression verification.

## Verification

## Related Issue
- Closes #2028
- [#2028 [Bug]: initResources() never inspects the destination, so a torn ~/.gsd/agent/ copy is never repaired — and hasMissingBundledResourceFiles() has no caller](https://github.com/open-gsd/gsd-pi/issues/2028)

## User
- @bbasketballer75

## Repo
- `open-gsd/gsd-pi`

## Branch
- `issue/2028-bug-initresources-never-inspects-the-des-1787883811`

### DIFF
--- a/src/resource-loader.ts
+++ b/src/resource-loader.ts
@@ -746,7 +746,8 @@ export function initResources(agentDir: string, skillsDir: string = join(agentDi
   if (manifest && isCurrentPackageManifest(manifest) && manifest.gsdVersion === currentVersion) {
     // Version matches — check content fingerprint for same-version staleness.
     const currentHash = getCurrentResourceFingerprint()
-    if (manifest.contentHash && manifest.contentHash === currentHash) {
+    if (manifest.contentHash && manifest.contentHash === currentHash
+      && !hasMissingBundledResourceFiles(extensionsDir, bundledExtensionsDir)) {
       return
     }
   }
--- a/src/tests/resource-loader.test.ts
+++ b/src/tests/resource-loader.test.ts
@@ -586,8 +586,9 @@ test("initResources repairs missing bundled extension files with a current manif
   initResources(fakeAgentDir, skillsDir);
 
   const extensionsDir = join(fakeAgentDir, "extensions", "gsd");
-  const missingFile = join(extensionsDir, "debug-logger.ts");
-  const corruptedFile = join(extensionsDir, "auto.ts");
+  const extensionSuffix = existsSync(join(extensionsDir, "auto.ts")) ? ".ts" : ".js";
+  const missingFile = join(extensionsDir, `debug-logger${extensionSuffix}`);
+  const corruptedFile = join(extensionsDir, `auto${extensionSuffix}`);
   const originalCorruptedContent = readFileSync(corruptedFile, "utf-8");
 
   rmSync(missingFile);
--- a/src/tests/resource-loader.test.ts
+++ b/src/tests/resource-loader.test.ts
@@ -543,7 +543,7 @@ test("initResources syncs top-level shared resources used by extension imports",
   );
 });
 
-test("initResources steady-state hash match returns before recursive drift checks", async (t) => {
+test("initResources steady-state extension match skips refresh for non-extension drift", async (t) => {
   const tmp = mkdtempSync(join(tmpdir(), "gsd-resource-loader-fast-path-"));
   const fakeAgentDir = join(tmp, ".gsd", "agent");
   const skillsDir = join(tmp, "skills");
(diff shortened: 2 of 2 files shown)


jeremymcs:

Re-checked PR #2030 against [#2028 [Bug]: initResources() never inspects the destination, so a torn ~/.gsd/agent/ copy is never repaired — and hasMissingBundledResourceFiles() has no caller](https://github.com/open-gsd/gsd-pi/issues/2028).

## Subtasks
- [x] **[Bug]: initResources() never inspects the destination, so a torn ~/.gsd/agent/ copy is never repaired — and hasMissingBu**

jeremymcs:

PR #2032, opened 2026-08-28
fix(issue): [Bug]: 503 no_available_providers never reaches model fallback — core-retry hand-off keys on a display-only string

### LINKED ISSUE #2031: [Bug]: 503 no_available_providers never reaches model fallback — core-retry hand-off keys on a display-only string
## Summary

On the `turn-abort` path, `no_available_providers` (HTTP 503) never reaches model fallback. Auto-mode retries the same failing model 3 times, then hard-aborts the unit — even though a healthy fallback is configured, available, and correctly resolved.

The root cause is that the hand-off from core retry to model fallback keys on a string prefix (`Retry failed after N attempts:`) that only exists in the TUI display layer and is never written into the message body. So the "core retry is exhausted, now try fallback" signal never arrives.

## Environment

- gsd-pi `1.16.2`
- macOS (darwin 25.5.0), node `v24.13.0`
- Anthropic-compatible custom provider via `models.json`
- Gateway returned a sustained `503 {"type":"no_available_providers"}` for the primary model while a configured fallback on a different provider was returning `200`

## Configuration

```yaml
models:
  execution:
    model: <provider-a>/<primary-model>
    thinking: xhigh
    fallbacks:
      - <provider-b>/<fallback-model>
```

Resolution verified correct via the shipped resolver:

```js
resolveModelWithFallbacksForUnit('execute-task', cwd)
// { primary: '<provider-a>/<primary-model>',
//   fallbacks: [ '<provider-b>/<fallback-model>' ], thinking: 'xhigh' }

getNextFallbackModel('<primary-model>', cfg)  // -> '<provider-b>/<fallback-model>'
isModelBlocked(...)                            // -> false
isModelTemporarilyUnavailable(...)             // -> false
```

The fallback model was present in `--list-models` and served `200` throughout.

## Observed behaviour

Auto-mode emitted the same error four times, then gave up:

```
Error: 503 {"error":{"type":"no_available_providers","message":"No available providers ..."}}
Error: 503 ...
Error: 503 ...
Error: 503 ...
Error: Retry failed after 3 attempts: 503 {"error":{"type":"no_available_providers", ...}}
```

No `fallback_provider_switch` status appeared, and no `fallback_chain_exhausted` error either — the chain was never entered.

Unit record (`.gsd/runtime/units/execute-task-<id>.json`):

```json
{
  "unitType": "execute-task",
  "phase": "recovered",
  "lastProgressKind": "harness-abort:turn-abort",
  "recoveryAttempts": 2,
  "harnessAbort": {
    "kind": "turn-abort",
    "reason": "Agent turn aborted before the unit could complete its gate evaluation. stopReason=error error=503 {\"error\":{\"type\":\"no_available_providers\", ...}}"
  }
}
```

## Expected behaviour

After core retry is exhausted for a `server`-class error, `tryProviderModelFallback` should run and switch to the next available entry in `fallbacks[]` — matching the intent documented at `agent-end-recovery.js` ("Server errors — try fallback model") and the behaviour established by #1364, #1533 and #806.

## Root cause

`classifyError` handles 503 correctly — it is `server`, which `isTransient` accepts:

```js
// error-classifier.js:159
if (SERVER_RE.test(errorMsg)) {
    return { kind: "server", retryAfterMs: retryAfterMs ?? 30_000 };
}
```

`handleAgentEnd` then defers it to core retry before the fallback branch is reachable:

```js
// bootstrap/agent-end-recovery.js:627
if (shouldDeferTransientErrorToCoreRetry(cls, rawErrorMsg, rawErrorMsg || displayMsg)) {
    return;    // <-- returns here; fallback lives at :702
}
```

```js
// bootstrap/agent-end-recovery.js:84-93
export function shouldDeferTransientErrorToCoreRetry(cls, rawErrorMsg, deferCheckMsg = rawErrorMsg) {
    if (!isTransient(cls) || cls.kind === "rate-limit") return false;
    if (!rawErrorMsg) return false;
    return !/retry failed after \d+ attempts:/i.test(deferCheckMsg);
}
```

The design implies a two-pass hand-off: pass 1 defers to core retry; once core retry is exhausted the message is expected to carry the `Retry failed after N attempts:` prefix, so pass 2 falls through to fallback. Verified with the shipped functions:

| `errorMessage` content | `kind` | `shouldDefer...` |
|---|---|---|
| `503 {"type":"no_available_providers", ...}` | `server` | `true` |
| `Retry failed after 3 attempts: 503 {...}` | `server` | `false` |

That prefix is never present in the message body. It is applied only when rendering:

```js
// gsd-agent-modes/.../chat-controller.js:395
host.showError(`Retry failed after ${event.attempt} attempts: ${event.finalError || "Unknown error"}`);
```

and the event carries the untouched original text:

```js
// gsd-agent-core/dist/session/agent-session-prompt.js:44-50
if (msg.stopReason === "error" && this.host._retryAttempt > 0) {
    this.host.emit({
        type: "auto_retry_end",
        success: false,
        attempt: this.host._retryAttempt,
        finalError: msg.errorMessage,   // <-- raw 503, no prefix
    });
}
```

`grep -rn "Retry failed after"` across the package matches only `chat-controller.js:395`, confirming nothing writes it back into `msg.errorMessage`.

So when core retry is exhausted and the turn ends as `stopReason=error`, `handleAgentEnd` re-reads the same unprefixed 503, `shouldDeferTransientErrorToCoreRetry` returns `true` again, and it is deferred to a retry handler that has already terminated. `recordRetryableTurnAbort` records the abort but performs no model switch, so `tryProviderModelFallback` at `:702` is unreachable for this class of error.

## Suggested fix directions

1. Have the retry handler record exhaustion in the message/event state (e.g. an `attemptsExhausted` flag, or writing the prefix into `errorMessage`) and key `shouldDeferTransientErrorToCoreRetry` off that instead of a display string.
2. Alternatively, pass the retry-exhaustion state through the `agent_end` event so recovery can distinguish "core is mid-retry" from "core is done".

A structural signal seems preferable to the string match either way, since the current predicate silently regresses if the display wording changes.

## Repro sketch

1. Configure any phase with a `fallbacks[]` entry on a second, healthy provider.
2. Make the primary provider return a sustained `503` whose body has no network keywords (so `NETWORK_RE` misses and it lands in `server`).
3. Run auto-mode and dispatch a unit for that phase.

Observed: 3 same-model retries, then `Retry failed after 3 attempts:` and a `turn-abort` on the unit.
Expected: a switch to the configured fallback after retries are exhausted.

Happy to test a patch against this setup — the failing provider reproduces it on demand.e.g. `{"error":{"type":"no_available_providers"}}`).
3. Run auto-mode.
4. Observe 3 same-model retries, `Retry failed after 3 attempts:`, and a `turn-abort` — with no `fallback_provider_switch`.

Happy to test a patch against this gateway setup, since the 503 reproduces reliably here.

jeremymcs:
## Summary
- Exposed core retry exhaustion to model fallback with regression coverage and clean static verification.

## Verification

## Related Issue
- Closes #2031
- [#2031 [Bug]: 503 no_available_providers never reaches model fallback — core-retry hand-off keys on a display-only string](https://github.com/open-gsd/gsd-pi/issues/2031)

## User
- @Jack11111eee

## Repo
- `open-gsd/gsd-pi`

## Branch
- `issue/2031-bug-503-no-available-providers-never-rea-1787897480`

### DIFF
--- a/docs/extension-sdk/api-reference.md
+++ b/docs/extension-sdk/api-reference.md
@@ -345,7 +345,7 @@ Agent lifecycle events carry optional correlation metadata when the current prov
 |-------|------|-------------|-------------|
 | `before_agent_start` | `BeforeAgentStartEvent` | `BeforeAgentStartEventResult` | After user submits prompt, before agent loop. Can inject system prompt or message. |
 | `agent_start` | `AgentStartEvent` | — | Agent loop started. May include `sessionId` and `turnId`. |
-| `agent_end` | `AgentEndEvent` | — | Agent loop ended. Includes `messages`; may include `sessionId`, `turnId`, and `abortOrigin`. |
+| `agent_end` | `AgentEndEvent` | — | Agent loop ended. Includes `messages`; may include `willRetry`, `sessionId`, `turnId`, and `abortOrigin`. |
 | `stop` | `StopEvent` | — | Agent is truly idle with no follow-up or steering pending. May include `sessionId`, `turnId`, and `abortOrigin`. |
 | `turn_start` | `TurnStartEvent` | — | Start of each turn. Includes `turnIndex` and `timestamp`; may include `sessionId` and `turnId`. |
 | `turn_end` | `TurnEndEvent` | — | End of each turn. Includes `turnIndex`, `message`, and `toolResults`; may include `sessionId` and `turnId`. |
--- a/packages/gsd-agent-core/src/session/agent-session-events.ts
+++ b/packages/gsd-agent-core/src/session/agent-session-events.ts
@@ -54,11 +54,13 @@ export class AgentSessionEventsModule {
 			}
 		}
 
+		const agentEndWillRetry = event.type === "agent_end" ? this.willRetryAfterAgentEnd(event) : undefined;
+
 		// Emit to extensions first
-		await this.emitExtensionEvent(event);
+		await this.emitExtensionEvent(event, agentEndWillRetry);
 
 		// Notify all listeners
-		this.emit(event.type === "agent_end" ? { ...event, willRetry: this.willRetryAfterAgentEnd(event) } : event);
+		this.emit(event.type === "agent_end" ? { ...event, willRetry: agentEndWillRetry ?? false } : event);
 
 		// Handle session persistence
 		if (event.type === "message_end") {
@@ -155,12 +157,16 @@ export class AgentSessionEventsModule {
 		Object.assign(targetRecord, replacement);
 	}
 
-	async emitExtensionEvent(event: AgentEvent): Promise<void> {
+	async emitExtensionEvent(event: AgentEvent, agentEndWillRetry?: boolean): Promise<void> {
 		if (event.type === "agent_start") {
 			this.host._turnIndex = 0;
 			await this.host._extensionRunner.emit({ type: "agent_start" });
 		} else if (event.type === "agent_end") {
-			await this.host._extensionRunner.emit({ type: "agent_end", messages: event.messages });
+			await this.host._extensionRunner.emit({
+				type: "agent_end",
+				messages: event.messages,
+				willRetry: agentEndWillRetry ?? false,
+			});
 		} else if (event.type === "turn_start") {
 			const extensionEvent: TurnStartEvent = {
 				type: "turn_start",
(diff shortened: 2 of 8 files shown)


jeremymcs:

Re-checked PR #2032 against [#2031 [Bug]: 503 no_available_providers never reaches model fallback — core-retry hand-off keys on a display-only string](https://github.com/open-gsd/gsd-pi/issues/2031).

## Subtasks
- [x] **[Bug]: 503 no_available_providers never reaches model fallback — core-retry hand-off keys on a display-only string**

PR #2035, opened 2026-08-28
fix(issue): [bug] model router ignores dotted Copilot model IDs and pins all phases to the ceiling model

### LINKED ISSUE #2034: [bug] model router ignores dotted Copilot model IDs and pins all phases to the ceiling model
## Summary

The model router resolves tiering and capability checks against a hyphenated alias map (), while the actual GitHub Copilot dispatch IDs are provider-qualified dotted strings ().

As a result, the internal router misses tier matches and falls through the unknown-model bypass, so light/standard tasks can end up pinned to the ceiling model (Opus) instead of downshifting to Luna/Sonnet.

## Observed behavior

-  /  keys are hyphenated
-  /  /  do not normalize dotted vs hyphenated variant names
-  is valid and in the catalog, but it is not recognized by the router map
- dynamic phases therefore skip the normal light/standard/heavy downgrade path and stay at Opus

## Reproduction

A light task or standard task with the canonical Copilot ID  still routes to the ceiling model because the router never recognizes the dotted ID as a valid alias of .

## Why this matters

This defeats the phase routing policy and makes the dynamic router effectively ignore the intended light/standard/heavy cap. This is especially visible when a project depends on  and explicit downgrade behavior.

## Expected behavior

- canonical provider-qualified Copilot model IDs should resolve cleanly to the correct tier/cost/capability map
- dotted and hyphenated aliases should normalize to the same internal model key
- light/standard tasks should downgrade correctly; only heavy tasks should be pinned to the ceiling model

## Proposed fix

Normalize model IDs before tier lookup and capability checks, and add dotted aliases to the model/router tables used by  and .

## Notes

This was reproduced while validating the GSD Pi dynamic-routing behavior in the local fork; local patching has already been applied in the fork branch  and the issue is being tracked here for upstream visibility.

jeremymcs:
## Summary
- Normalized dotted Copilot Claude model aliases across routing tables and verified correct tier downgrades and ceiling preservation.

## Verification

## Related Issue
- Closes #2034
- [#2034 [bug] model router ignores dotted Copilot model IDs and pins all phases to the ceiling model](https://github.com/open-gsd/gsd-pi/issues/2034)

## User
- @pimmink

## Repo
- `open-gsd/gsd-pi`

## Branch
- `issue/2034-bug-model-router-ignores-dotted-copilot--1787911165`

### DIFF
--- a/docs/user-docs/dynamic-model-routing.md
+++ b/docs/user-docs/dynamic-model-routing.md
@@ -69,10 +69,27 @@ Keep `cross_provider: false` when routing inside a flat-rate subscription unless
 Override which model is used for each tier. When omitted, the router uses a built-in capability mapping that knows common model families:
 
 - **Light:** `claude-haiku-4-5`, `gpt-4o-mini`, `gpt-4.1-mini`, `gpt-4.1-nano`, `gpt-5-mini`, `gpt-5-nano`, `gpt-5.1-codex-mini`, `gpt-5.3-codex-spark`, `gpt-5.4-mini`, `gemini-2.0-flash`
-- **Standard:** `claude-sonnet-4-6`, `gpt-4o`, `gpt-4.1`, `gpt-5.1-codex-max`, `gemini-2.5-pro`, `deepseek-chat`
-- **Heavy:** `claude-opus-4-6`, `claude-opus-4-7`, `claude-opus-4-8`, `gpt-5`, `gpt-5-pro`, `gpt-5.1`, `gpt-5.2`, `gpt-5.2-codex`, `gpt-5.3-codex`, `gpt-5.4`, `gpt-5.5`, `gpt-5.6-sol`, `gpt-5.6-terra`, `gpt-5.6-luna`, `o1`, `o3`, `o4-mini`
+- **Standard:** `claude-sonnet-4`, `claude-sonnet-4-5`, `claude-sonnet-4-6`, `gpt-4o`, `gpt-4.1`, `gpt-5.1-codex-max`, `gemini-2.5-pro`, `deepseek-chat`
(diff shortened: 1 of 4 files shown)


jeremymcs:

Re-checked PR #2035 against [#2034 [bug] model router ignores dotted Copilot model IDs and pins all phases to the ceiling model](https://github.com/open-gsd/gsd-pi/issues/2034).

## Subtasks
- [x] **[bug] model router ignores dotted Copilot model IDs and pins all phases to the ceiling model**

pimmink:
I agree this fixes the immediate bug, but I’m concerned we’re encoding model-name variants in a hardcoded way that won’t scale.

Right now the failure mode seems structural: model IDs can vary by separator/casing (e.g. `.`, `_`, `-`), and when they don’t match exactly we fall through and effectively pin phases to the ceiling model. Adding more explicit aliases in routing tables solves this instance, but we’ll likely keep hitting the same class of bug.

Could we refactor toward a canonical matching path instead?

### Suggested direction
1. Introduce a single `canonicalizeModelId()` step (e.g. trim + lowercase + normalize separators).
2. Store/compare routing keys in canonical form only.
3. Keep a small explicit alias map only for real semantic aliases (not formatting variants).
4. On unknown model IDs, use explicit fallback behavior + warning log (instead of silent mismatch).
5. Add table-driven tests proving dotted/underscored/hyphenated variants resolve to the same routing target and preserve downgrade/ceiling semantics.

This would make the router resilient for future Copilot model ID format changes and reduce ongoing maintenance churn from per-model hardcoding.

@copilot review this

jeremymcs:
**Verifying** — checking the applied changes.
**Resolved** — addressed in commit `6c5f971`.

jeremymcs:

jeremymcs:
Addressed in commit `6c5f971`.

> Responding to [comment by @pimmink](https://github.com/open-gsd/gsd-pi/pull/2035#issuecomment-5452378111):
> I agree this fixes the immediate bug, but I’m concerned we’re encoding model-name variants in a hardcoded way that won’t…

Refactored model routing around a single canonical ID path that normalizes provider prefixes, whitespace, casing, and `.`, `_`, and `-` separators while keeping only a semantic Gemini alias. Added explicit deduplicated warnings and safe fallback behavior for unknown IDs plus table-driven tests preserving downgrade and ceiling semantics across formatting variants.

jeremymcs:

PR #2065, opened 2026-08-29
release: v1.17.0

jeremymcs:
Version bump for v1.17.0. Direct push to main was rejected by branch rules, so the release commit lands through the merge queue. npm, the `v1.17.0` tag and the GitHub release are already published.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,69 @@ This changelog starts from the `open-gsd/gsd-pi` ownership baseline. Earlier pro
 
 ## [Unreleased]
 
+## [1.17.0] - 2026-08-29
+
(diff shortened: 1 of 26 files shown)


jeremymcs:
Duplicate of #2066 — and #2066 is the correct one to keep. This PR's lockfile fold raced the engine publishes: `@opengsd/engine-win32-x64-msvc@1.17.0` is missing from pnpm-lock.yaml (the fold's `pnpm install --lockfile-only` ran before the win32 engine was queryable on npm), so `pnpm install --frozen-lockfile` fails here. #2066 was regenerated after all five engines were live and carries the complete lockfile. Tracked the fold race separately.

Issue #2061, opened 2026-08-29
npm install -g --ignore-scripts produces a broken CLI: loader statically imports @gsd/pi-coding-agent that only the postinstall link creates


jeremymcs:
Found while validating #2059 (the docker e2e `runtime-local` failure there reproduces this exactly).

**Repro** (tarball packed from current main, clean environment):
```
npm install -g --ignore-scripts ./opengsd-gsd-pi-1.16.2.tgz   # succeeds
gsd --version
# Error [ERR_MODULE_NOT_FOUND]: Cannot find package '@gsd/pi-coding-agent'
#   imported from .../dist/extension-discovery.js
```
The same install **without** `--ignore-scripts` works (postinstall runs `scripts/postinstall.js`/`link-workspace-packages.cjs`, which create `node_modules/@gsd/*` links).

**Impact:** any hardened environment that installs with `--ignore-scripts` (common in CI, containers, air-gapped setups — and our own docker `runtime-local` e2e) gets a CLI that fails on first invocation, after a successful-looking install. This is the same 'silent success, broken runtime' class as the extensions registry bugs.

**Possible directions** (needs an ADR-level decision, packaging is release-critical):
1. Ship the workspace packages as real bundled deps in the tarball (bundledDependencies or flattening into the package's own node_modules), making the loader's resolution work without scripts; or
2. Make the dist loader resolve `@gsd/*` from a vendored fallback path when the link is absent, with a loud doctor-style warning; or
3. Keep the postinstall link but fail the install loudly when scripts are skipped (worst option — still broken, but honest).

**Acceptance:** `npm install -g --ignore-scripts <tarball> && gsd --version` exits 0 (option 1/2), or the install itself fails loudly (option 3). The docker `runtime-local` e2e in `tests/e2e/docker/runtime.e2e.test.ts` covers this scenario.

jeremymcs:
### Issue evaluation - needs review

Issue: [npm install -g --ignore-scripts produces a broken CLI: loader statically imports @gsd/pi-coding-agent that only the postinstall link creates](https://github.com/open-gsd/gsd-pi/issues/2061)

Needs maintainer review because it touches a privileged area.

Confidence: high (75%)

**Safety flags**
- privileged-area

**Labels applied**
- `needs-maintainer-review`

PR #2059, opened 2026-08-29
chore(deps): bump undici from 7.28.0 to 7.29.0

jeremymcs:
Same bump as #1603, re-cut from current main — the original dependabot branch had accumulated stale merge commits and a lockfile conflict with #1631. Single dependency change in `packages/pi-coding-agent` + regenerated lockfile.

Supersedes #1603.

PR #2066, opened 2026-08-29
release: v1.17.0

jeremymcs:
Lands the 1.17.0 version bump on main to match the published release (tag `v1.17.0`, npm `latest`). The NPM Publish workflow created the release commit + tag off-main and published `@opengsd/engine-*@1.17.0` first, so the regenerated lockfile resolves cleanly — this is the same post-release sync as #2013 for 1.16.2.

### DIFF
--- a/extensions/google-search/package.json
+++ b/extensions/google-search/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@gsd-extensions/google-search",
-  "version": "1.16.2",
+  "version": "1.17.0",
   "description": "Web search via Google with AI-synthesized answers and source citations",
   "type": "module",
   "keywords": [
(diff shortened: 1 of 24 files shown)


jeremymcs:
Blocked by the same pre-existing packaging bug that hit #2059: the docker `runtime-local` e2e installs with `--ignore-scripts` and the loader's static `@gsd/pi-coding-agent` import fails without the postinstall link — tracked in #2061 (required ci-gate has no bypass). The bump content here is exactly what shipped in the tagged `v1.17.0` release commit; keeping this open as the ready-to-merge main-sync once #2061 lands. npm `latest` is already correct (1.17.0); only main's `package.json`/lockfile lag.

PR #2079, opened 2026-08-30
fix(issue): [Bug]: auto-mode setModel persist:false is ignored — every unit dispatch rewrites the global settings.json default model

### LINKED ISSUE #2076: [Bug]: auto-mode setModel persist:false is ignored — every unit dispatch rewrites the global settings.json default model
## Bug type

Incorrect behavior or output

## Summary

In auto-mode, `setModel(model, { persist: false })` silently writes the selected model to the global `~/.gsd/agent/settings.json` on every unit dispatch, because the `persist` option is never threaded through the setModel call chain — so per-dispatch model selection pollutes the shared default model for every project on the machine.

## Steps to reproduce

1. Configure a project (via its `.gsd/PREFERENCES.md` `models:` block) so that some auto-mode unit resolves to a model **different** from the global default in `~/.gsd/agent/settings.json` (e.g. global default = `ccswitch-anthropic-k3/kimi-k3`, unit model = `ccswitch-grok/grok-4.6`).
2. Ensure the project has **no** project-scoped `.gsd/settings.json`, so it relies on the global default.
3. Run `gsd auto` and let it dispatch a unit (e.g. `execute-task`).
4. Inspect `~/.gsd/agent/settings.json` after the dispatch.

## Expected behavior

`selectAndApplyModel` calls `pi.setModel(model, { persist: false })` (see `resources/extensions/gsd/auto-model-selection.js`), which should switch the model **in memory only** for the dispatched unit. The global `settings.json` `defaultProvider`/`defaultModel` should remain whatever the user configured — auto-mode's per-unit model choice must not be persisted.

## Actual behavior

`~/.gsd/agent/settings.json` is rewritten with the auto-selected model (e.g. it flips from `ccswitch-anthropic-k3/kimi-k3` to `ccswitch-grok/grok-4.6`). Every project that relies on the global default then silently runs on the wrong model.

## Root cause (located, with stack trace)

The `persist` option is accepted at the extension boundary but dropped before the write happens:

`packages/gsd-agent-core/dist/session/agent-session-extensions.js` — `setModel`:
```js
setModel: async (model, options) => {
    if (!this.host.modelRegistry.hasConfiguredAuth(model))
        return false;
    await this.host.setModel(model);        // ← writes settings.json unconditionally
    if (options?.persist === false)          // ← checked AFTER the write; dead branch
        return true;
    return true;                             //    both branches return true
},
```

`this.host.setModel(model)` → `AgentSession.setModel(model)` → `AgentSessionModelModule.setModel(model)` (`packages/gsd-agent-core/dist/session/agent-session-model.js`), whose signature takes **no options** and unconditionally calls:
```js
this.host.settingsManager.setDefaultModelAndProvider(model.provider, model.id);
```

So `persist: false` is evaluated only after the global write already happened, and the underlying model module has no way to skip persistence. Both ends of the chain must be fixed together: the extension must forward `options`, and `AgentSessionModelModule.setModel` must accept an options bag and skip `setDefaultModelAndProvider` when `persist === false`.

Observed stack (captured live via instrumentation on the real dispatch):
```
runUnitPhase (auto/unit-phase.js)
→ selectAndApplyModel (auto-model-selection.js)
→ pi.setModel (loader.js)
→ Object.setModel (agent-session-extensions.js)
→ AgentSession.setModel (agent-session.js)
→ AgentSessionModelModule.setModel (agent-session-model.js)
→ SettingsManager.setDefaultModelAndProvider → writes ~/.gsd/agent/settings.json
```

## GSD version

1.16.2

## Node.js version

v24.13.0

## Operating system

macOS 26.5.2

## Install method

npm global install (via pnpm global)

## Affected area

Auto-mode / dispatch loop  (also: State / persistence, AI provider / model integration)

## Model and provider

Any pair where the auto-dispatched unit model differs from the global default. Reproduced with global `ccswitch-anthropic-k3/kimi-k3` vs unit `ccswitch-grok/grok-4.6`.

## Impact

- Severity: high — silently changes the user's configured default model machine-wide.
- Frequency: 100% repro on every auto-mode unit dispatch whose selected model differs from the global default.
- Consequence: shared global `settings.json` is polluted; projects without a project-scoped override run on the wrong model; the corruption is attributed to "the user changed it", making it very hard to diagnose. With multiple `gsd` processes across different projects, whichever project dispatches last wins the global default.

## Additional information

- Note that `cycleScopedModel` / `cycleAvailableModel` in the same file also call `setDefaultModelAndProvider`, but those are user-initiated model cycles where persisting is arguably intended; the defect is specifically the `setModel(model, { persist: false })` path used by auto-mode dispatch.
- A local workaround (threading `options.persist` through the three files and skipping `setDefaultModelAndProvider` when `persist === false`) resolves it; interactive `/model` selection (no options passed) still persists as before.
- Possibly related fallback-selection issue (separate root cause): #2031.

jeremymcs:
## Summary
- Fixed non-persistent model selection and verified it with builds, tests, type-checking, and fast gates.

## Verification

## Related Issue
- Closes #2076
- [#2076 [Bug]: auto-mode setModel persist:false is ignored — every unit dispatch rewrites the global settings.json default model](https://github.com/open-gsd/gsd-pi/issues/2076)

## User
- @Jack11111eee

## Repo
- `open-gsd/gsd-pi`

## Branch
- `issue/2076-bug-auto-mode-setmodel-persist-false-is--1788053619`

### DIFF
--- a/packages/gsd-agent-core/src/agent-session.test.ts
+++ b/packages/gsd-agent-core/src/agent-session.test.ts
@@ -4,6 +4,7 @@ import { SessionManager } from "@gsd/pi-coding-agent/core/session-manager.js";
 import { resolvePath } from "@gsd/pi-coding-agent/utils/paths.js";
 import { parseSkillBlock } from "./agent-session.ts";
 import { AgentSessionExtensionsModule } from "./session/agent-session-extensions.ts";
+import { AgentSessionModelModule } from "./session/agent-session-model.ts";
 import { AgentSessionNavigationModule } from "./session/agent-session-navigation.ts";
 import { AgentSessionPromptModule } from "./session/agent-session-prompt.ts";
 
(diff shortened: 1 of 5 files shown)


jeremymcs:

Re-checked PR #2079 against [#2076 [Bug]: auto-mode setModel persist:false is ignored — every unit dispatch rewrites the global settings.json default model](https://github.com/open-gsd/gsd-pi/issues/2076).

## Subtasks
- [x] **[Bug]: auto-mode setModel persist:false is ignored — every unit dispatch rewrites the global settings.json default model**

PR #2093, opened 2026-08-30
fix(gsd): make Copilot suggestions fail closed

### LINKED ISSUE #2090: [Bug]: Make GitHub Copilot savings suggestions fail closed
## Summary

Post-merge correction for GitHub Copilot cheaper-model suggestions.

## Problem

Average capability scoring could describe a cheaper model as better despite a lower required capability dimension. Candidate selection could also choose the cheapest candidate before applying the safety comparison. Missing live tool-call metadata was displayed optimistically in diagnostics.

## Scope

- Use deterministic dominance: every required capability dimension must be equal or higher; at least one higher means better-and-cheaper; exact equality permits cheaper-equivalent; any lower or unknown dimension means no suggestion.
- Apply the dominance gate before pricing tie-breaks so an inferior cheapest candidate cannot suppress a safe alternative.
- Keep suggestions provider-scoped and never auto-switch models.
- Preserve unknown live tool support as `unknown`.

## Validation

Tests cover equivalent, superior, inferior, and missing capability profiles; a real same-tier inferior pairing produces no notification; positive equivalent suggestions remain deduplicated and secret-free.

Remote full gate will be linked from the PR.

## Follow-up boundary

Explicit manual versus automatic selection origin requires the separate runtime-catalog/origin RFC; this bug does not introduce a fragile suppression flag or infer that every `source: set` event is manual.

pimmink:
## Linked issue

Closes #2090

---

## TL;DR

**What:** Makes GitHub Copilot savings suggestions require deterministic capability dominance before a model can be described as equivalent or better.
**Why:** Average scoring could hide a lower required capability dimension, while choosing the cheapest candidate before safety filtering could suppress an eligible alternative.
**How:** Filter candidates by per-dimension dominance before ranking by economics, preserve unknown facts as incomparable, resolve inherited capability profiles through the same router path, and retain advisory-only notifications.

## What

- A candidate must be equal or higher on every required capability dimension.
- One strictly higher dimension produces a better-and-cheaper notice.
- Exact equality produces a cheaper-equivalent notice.
- Any lower, missing, non-finite, or unknown dimension suppresses the suggestion.
- Inherited capability profiles are resolved through the router's capability-profile resolver instead of direct table lookup.
- Dominance filtering happens before price-based ranking, allowing a safe second candidate when the cheapest candidate is inferior or incomparable.
- `/gsd copilot-models why` retains `unknown` live tool-call status rather than claiming support.
- Notifications remain advisory only and never change the active model.

## Why

A cheaper model is not a safe alternative when it loses a required capability. The former average score could obscure that regression. This change preserves user control and avoids presenting incomparable economics/capabilities as a recommendation.

## How

The implementation shares deterministic dominance at the routing layer and applies it to candidate selection before economics tie-breaking. Regression tests cover equal, higher, lower, malformed, inherited/unknown profile, and three-model ordering cases.

## Evidence

- Exact branch head: `00f9ee0fdad9d97d12bb88e42bc123c5f465fe11`.
- Rebased onto current `upstream/main`: `0 behind / 5 ahead` before push.
- Focused notification/handler/router suites: `pnpm exec tsx --test src/resources/extensions/gsd/tests/copilot-catalog-notifications.test.ts src/resources/extensions/gsd/tests/copilot-models-handler.test.ts src/resources/extensions/gsd/tests/model-router.test.ts` passed **204/204**.
- `pnpm exec tsc --noEmit --project tsconfig.extensions.json` passed.
- `git diff --check` passed.
- Post-push upstream checks are running; Socket checks are green, GitHub Actions jobs were queued at the time of this update.

## Change type

- [x] `fix` — Bug fix

## Scope

## Breaking changes

- [x] No breaking changes

## Test plan

- [x] Focused local tests pass
- [x] Extension typecheck passes
- [ ] Upstream CI completes after latest push

## AI disclosure

- [x] This PR includes AI-assisted code. GitHub Copilot and Codex were used for adversarial review; findings were inspected, corrected, and verified.


jeremymcs (approved):
Validated: dominance comparison is fail-closed in the right direction and the test fixtures match the real HEAD capability tables (targeted suites 75/75 on the branch). I merged current main into the branch to resolve the single conflict on the tool-calls display line in favor of your toolCallsSummary() helper (109/109 across both PRs' tests post-resolution, CI green on the merge commit). Merging.

Issue #2158, opened 2026-09-04
Windows: readProjectProgressViaBridge read recreates a deleted gsd.db — 'a read must not recreate gsd.db' fails windows-portability (first surfaced by #2156)


jeremymcs:
## Summary

`packages/mcp-server/src/workflow-tools.test.ts` → `readProjectProgressViaBridge` → **"returns null without recreating a DB removed at the open boundary"** fails deterministically on the `windows-portability` CI job (failed the original run and a clean `--failed` rerun), with:

```
AssertionError [ERR_ASSERTION]: a read must not recreate gsd.db
true !== false
at dist-test\packages\mcp-server\src\workflow-tools.test.js:3408
```

374/377 tests in that suite pass on the same job; this one assertion fails.

## Provenance

- The test arrives with #2105 (`4f9c72dff`, 2026-08-31, "make progress reads database-authoritative").
- At current `main` (`e34e7d7d0`) the `windows-portability` job has only ever been **skipped** (light diffs), so this assertion appears to have never executed on Windows in CI until PR #2156's heavy gate ran it.
- PR #2156's diff (`src/resources/extensions/gsd/{publication,journal,worktree-lifecycle,auto-worktree-merge-already-merged}.ts`) has **no import path** into `packages/mcp-server` (verified: no `journal.js`/`publication.js`/`worktree-lifecycle.js` imports anywhere under `packages/mcp-server/src`), and its Linux `build` job passes with 14,787+ tests green after one unrelated test-fixture fix. The Windows failure is unrelated to that diff.

## Expected

A progress read through the bridge must not (re)create `gsd.db` after it was removed at the open boundary — on any platform.

## Actual (Windows)

The read recreates `gsd.db`.

## Suspected area

The open-boundary guard exercised by `readProjectProgressViaBridge` (`packages/mcp-server/src/workflow-tools.ts` → mcp-bridge → progress read seam added in #2105). Windows-only file semantics around unlink/open are the likely differentiator.

## Evidence

- Original failure: https://github.com/open-gsd/gsd-pi/actions/runs/33863300539/job/100999489240
- Clean rerun (failed again): same run via `--failed` rerun, `windows-portability completed failure`.

jeremymcs:
[...]

PR #2163, opened 2026-09-04
fix(gsd): detect same-version resource drift via live fingerprint

### LINKED ISSUE #2106: [Bug]: precomputed resource fingerprint prevents initResources() from detecting the same-version changes it exists to catch
### Bug type

Incorrect behavior or output

### Summary

`initResources()` skips the resource sync when the content fingerprint matches, and the comment above that check says the fingerprint exists specifically to catch **same-version content changes**:

`dist/resource-loader.js:699-701`
```js
// Skip the full copy when both version AND content fingerprint match.
// Version-only checks miss same-version content changes (npm link dev workflow,
// hotfixes within a release). The content hash catches those at ~1ms cost.
```

But `getCurrentResourceFingerprint()` does not hash anything — it returns a **precomputed file shipped inside the package**:

`dist/resource-loader.js:149-159`
```js
function getCurrentResourceFingerprint() {
    try {
        const precomputed = readFileSync(join(resourcesDir, resourceFingerprintFileName), 'utf-8').trim();
        if (/^[a-f0-9]{16}$/i.test(precomputed)) {
            return precomputed;
        }
    }
    catch {
        // Source-tree and partial-build workflows may not have a precomputed hash.
    }
    return computeResourceFingerprint();
}
```

`resourceFingerprintFileName` is `.managed-resources-content-hash` (line 27), it lives under `dist/resources/`, and `dist` is in the package's `files` array — so it ships to every install.

Consequence: editing any file under `dist/resources/**` in an installed package does not change the value this function returns, the fingerprint still matches the manifest, `initResources()` returns early, and the edit is **never** copied into `~/.gsd/agent/`. The two workflows the comment names — `npm link` dev and same-release hotfixes — are exactly the two the precomputed hash prevents it from detecting.

The failure is silent. Files change, nothing happens, no warning.

### Steps to reproduce

On any normal global install (no `npm link` needed):

1. `cat "$(npm root -g)/@opengsd/gsd-pi/dist/resources/.managed-resources-content-hash"`
2. `cat ~/.gsd/agent/managed-resources.json` — note `contentHash` matches it
3. Edit any file under `dist/resources/extensions/gsd/` (change a string; a same-size edit is fine)
4. Launch `gsd`
5. Diff the edited file against its counterpart in `~/.gsd/agent/extensions/gsd/`

### Expected behavior

Per the comment at :699-701, a same-version content change is detected and the resources re-sync, so the edited file reaches `~/.gsd/agent/`.

### Actual behavior

The fingerprint is unchanged (it was read from the shipped file, not computed), so `initResources()` returns at :705 and the edited file never reaches `~/.gsd/agent/`. The runtime keeps loading the old copy.

Verified on this host: the shipped `.managed-resources-content-hash` and the `contentHash` in `~/.gsd/agent/managed-resources.json` were both `5b8cd91e97767b5b`, so the early return was taken on every launch. Deleting the precomputed file made `getCurrentResourceFingerprint()` fall through to `computeResourceFingerprint()`, which returned a live hash of the edited tree (`2584c4b7ea4e6855`) — that differed from the manifest, and the full sync ran as intended.

(Note `collectFileEntries` at :270 already skips `.managed-resources-content-hash` itself, so a live hash is stable regardless of whether that file is present.)

### GSD version

1.17.0

### Node.js version

22.22.0

### Operating system

Windows 11 Pro 26100

### Install method

npm global install

### Affected area

MCP / tools / extensions

### Impact

This cost a full debugging cycle while shipping the fix in #2104. The natural way to validate a fix against a real install is to patch the installed package and run it — and here that silently does nothing, while every obvious check says success: the files on disk *are* patched, and a grep of `dist/` confirms it. The only thing that reveals the problem is diffing `~/.gsd/agent/`, which you'd have no reason to suspect.

It also makes third-party patch instructions actively misleading. The handoff that started #2103 specified patching `dist/…` and offered "grep dist/ for the change" as its acceptance criterion — both of which pass while the running code is untouched.

Low severity for end users (a version bump forces a resync regardless), but it is a sharp edge for anyone developing against or hotfixing an installed copy.

Relevant code, all in `dist/resource-loader.js` of 1.17.0:

- `:27` — `const resourceFingerprintFileName = '.managed-resources-content-hash';`
- `:149-159` — `getCurrentResourceFingerprint()` prefers the shipped file
- `:270` — `collectFileEntries` skips the fingerprint file when computing
- `:699-706` — the early return and the comment stating the intent

I checked for an existing override and found none: no `GSD_FORCE*`, `FORCE_RESOURCE*`, or `GSD_SKIP_RESOURCE*` handling anywhere in `dist/`.

### Additional information

**Possible fixes**, roughly in increasing cost — I have no strong preference and you'll know which fits your constraints:

1. **An escape hatch.** `GSD_FORCE_RESOURCE_SYNC=1` bypasses the early return. Smallest change; makes the trap escapable once you know it exists, but doesn't help anyone who doesn't.
2. **Surface it in `gsd doctor`.** A check that compares the shipped fingerprint against a live `computeResourceFingerprint()` and reports drift, with `--fix` forcing a resync. `dist/resources/extensions/gsd/doctor.js` already exists as a home for this, and doctor is where someone confused by this would plausibly look.
3. **Verify rather than trust.** Keep the precomputed file as the fast path, but when it matches, also confirm it against a live hash — the comment at :701 already prices a full computation at ~1ms, which suggests the saving being protected is small. This is the only option that makes the comment at :699-701 true as written.

I'd lean toward 2 or 3, since 1 only helps people who already know. Happy to open a PR for whichever you prefer.

Adjacent closed issues I checked, neither of which covers this: **#2028** (initResources never inspects the destination, so a torn `~/.gsd/agent/` copy is never repaired — the `hasMissingBundledResourceFiles` guard now in the early-return condition came from that) and **#1853** (boot sweep deleting user-created extensions). This is a different failure: the destination is intact and complete, but the *source* fingerprint can't change.

Context: found while fixing #2103 / #2104.

_Investigated with AI assistance (Claude); every code reference and hash value above was read from a real 1.17.0 install and the reproduction was run before filing._

jeremymcs:
## TL;DR

**What:** `initResources()` now detects same-version resource drift by hashing the bundled tree live, and the manifest stamps that same live fingerprint.
**Why:** The comparison read the precomputed shipped `.managed-resources-content-hash` file — immutable post-install — so same-version changes to `dist/resources` (npm link, hotfix edits) could never trigger the resync the hash was added for (#2106).
**How:** Both the comparison and the manifest stamp use `computeResourceFingerprint()` (once per launch, ~25–30 ms measured), with Windows path-separator normalization keeping the live hash byte-identical to the build's; drift resyncs exactly once, then converges.

## What

- `src/resource-loader.ts` — comparison and manifest stamp share the live fingerprint; `collectFileEntries` normalizes `\` → `/` so the hash is platform-stable; the precomputed-file reader is removed (both callers migrated); a `setBundledResourcesDirForTests` seam (mirroring the existing `setGsdBrowserPackageSkillPathForTests` convention) enables the shipped-hash scenario; the stale "~1–2 ms" cost doc corrected to the measured ~25–30 ms.
- `src/tests/resource-loader-content-hash.test.ts` — drift test (shipped-hash manifest + post-install edit → resync runs, then a third call early-returns: exactly-once convergence, proven with a negative-control mutation), plus the unmodified-tree early-return control.

## Why

The precomputed file made the drift check a no-op by construction. An isolated validator proved the intermediate state ("compare live, stamp shipped") caused a mirror-image bug — a full resync on *every* launch, forever — which is why the manifest stamp shares the live source and a third-call convergence assertion pins it. A TOCTOU window (bundle advancing during the ~128 ms copy) is closed by stamping the *pre-sync* fingerprint.

Closes #2106

## How

Determinism verified (sorted entries, no mtime/size leakage, byte-identical to an independent reimplementation of the build hashing across 3 runs); Windows normalization complete (drive letters unreachable in relative paths, one normalization covers all entry kinds); behavior exercised at the real seam (`app-smoke`'s `initResources` tests green). Accepted boundaries, per review: an ABA race (concurrent mutation during the copy window that reverts) can leave one stale file undetected — snapshot-copy semantics out of scope for this convenience sync, documented in code; deletions are detected but `syncResourceDir`'s pre-existing merge semantics don't prune installed extras (candidate follow-up).

> This PR is AI-assisted.

## Testing

The author-supplied baseline had already established the pre-fix drift failure plus clean typecheck/fast verification; locally, an initial missing-`chalk` setup failure was repaired with the locked install, after which the content-hash battery passed 5/5, supporting resource-loader/conflict/staleness coverage passed 34/34, end-to-end filesystem and simulated-Windows checks passed with saved evidence, and the worktree remained clean.

<details>
<summary>Evidence: Issue #2106 end-to-end resource-sync evidence</summary>

`Same-version bundle drift triggered resync despite an unchanged shipped hash. A B→C mid-sync advance left manifest=B/live=C, remained detectable, converged to C on the next launch, then early-returned when unchanged. Native and simulated-Windows fingerprints matched.`

```text
Issue #2106 end-to-end resource sync evidence

Executable interface: initResources(agentDir, skillsDir)
Scenario: installed package has an immutable shipped content hash; bundled resource content changes at the same version; the bundle advances again during the copy/stamp window; subsequent launches must resync once and converge, then early-return.

{"stage":"initial install","shippedHash":"e7eafa605738cdfb","manifestHash":"e7eafa605738cdfb","installed":"bundle A"}
{"stage":"post-install edit B copied; source advances to C before stamp","shippedHashStill":"e7eafa605738cdfb","installed":"bundle B","manifestHash":"a9b677d89400f2c4","expectedPreSyncHash":"a9b677d89400f2c4","liveHash":"bb87e4039ebf7d3d","driftRemainsDetectable":true}
{"stage":"next launch converges to C","installed":"bundle C","manifestHash":"bb87e4039ebf7d3d","liveHash":"bb87e4039ebf7d3d","converged":true}
{"stage":"unchanged launch early-return","markerPreserved":true}

Result: the stale shipped hash remained unchanged, the manifest stamped the pre-sync B fingerprint rather than the post-copy C fingerprint, the next launch detected the remaining drift and installed C, and the following unchanged launch skipped the destructive resync.

Cross-platform executable check

The real computeResourceFingerprint() export was run once with the native path.relative implementation and once after Node's builtin path.relative binding was changed to emit Windows-style backslashes for the same nested resource tree.

{"nativeHash":"ac78edeeb49deaee","simulatedWindowsHash":"ac78edeeb49deaee","byteIdentical":true}

Result: separator normalization kept the live fingerprint byte-identical under simulated Windows path output.
```
</details>

## Pipeline

Updates from [git push no-mistakes](https://github.com/kunchenguid/no-mistakes)

<details>
<summary>✅ **intent** - passed</summary>

✅ No issues found.
</details>

<details>
<summary>✅ **Rebase** - passed</summary>

✅ No issues found.
</details>

<details>
<summary>✅ **Review** - passed</summary>

✅ No issues found.
</details>

<details>
<summary>✅ **Test** - passed</summary>

✅ No issues found.
- <code>`git status --short --branch`, `git rev-parse HEAD`, and `git merge-base --is-ancestor 304916e77e82d76357a01e03a719939a60f78152 origin/main`</code>
- <code>`pnpm install --frozen-lockfile --ignore-scripts` to repair missing isolated-worktree dependencies</code>
- `node --import ./src/resources/extensions/gsd/tests/resolve-ts.mjs --experimental-strip-types --test src/tests/resource-loader-content-hash.test.ts`
- `node --import ./src/resources/extensions/gsd/tests/resolve-ts.mjs --experimental-strip-types --test src/tests/resource-loader.test.ts src/tests/resource-loader-conflicts.test.ts`
- <code>Executable `initResources()` filesystem scenario covering immutable shipped-hash drift, B→C mid-sync mutation, next-launch convergence, and unchanged-launch early return</code>
- <code>Executable `computeResourceFingerprint()` comparison using native and simulated-Windows `path.relative` output</code>
- `Final worktree cleanliness and evidence-file integrity check`
</details>

<details>
<summary>✅ **Document** - passed</summary>

✅ No issues found.
</details>

<details>
<summary>✅ **Lint** - passed</summary>

✅ No issues found.
</details>

<details>
<summary>✅ **Push** - passed</summary>

✅ No issues found.
</details>

### DIFF
--- a/src/resource-loader.ts
+++ b/src/resource-loader.ts
@@ -781,6 +781,7 @@ export function initResources(agentDir: string, skillsDir: string = join(agentDi
   makeTreeWritable(agentDir)
 
   afterResourceSyncForTests?.()
+  // ABA note: a concurrent mutation of the bundled tree during the copy window that reverts before the next launch can leave one stale installed file undetected; snapshot-copy semantics are out of scope for this convenience sync.
   writeManagedResourceManifest(agentDir, currentHash)
   ensureRegistryEntries(join(agentDir, 'extensions'))
 }
--- a/src/resource-loader.ts
+++ b/src/resource-loader.ts
@@ -27,8 +27,8 @@ function loadPiCodingAgentModule(): Promise<PiCodingAgentModule> {
 // dist/resources/ is populated by the build step (`npm run copy-resources`) and
 // reflects the built state, not the currently checked-out branch.
 const packageRoot = resolve(dirname(fileURLToPath(import.meta.url)), '..')
-const resourcesDir = resolveBundledResourcesDirFromPackageRoot(packageRoot)
-const bundledExtensionsDir = join(resourcesDir, 'extensions')
+let resourcesDir = resolveBundledResourcesDirFromPackageRoot(packageRoot)
+let bundledExtensionsDir = join(resourcesDir, 'extensions')
 const resourceVersionManifestName = 'managed-resources.json'
 const resourceFingerprintFileName = '.managed-resources-content-hash'
 const gsdBrowserSkillName = 'gsd-browser'
@@ -125,7 +125,11 @@ function writeManagedResourceManifest(agentDir: string): void {
     gsdVersion: getBundledGsdVersion(),
     packageName: getBundledPackageName(),
     syncedAt: Date.now(),
-    contentHash: getCurrentResourceFingerprint(),
+    // Record the LIVE fingerprint of the tree that was just synced. Stamping
+    // the shipped precomputed value here would leave the manifest permanently
+    // mismatched with the live tree after a same-version edit, forcing a full
+    // resync on every launch (#2106).
+    contentHash: computeResourceFingerprint(),
     installedExtensionRootFiles,
     installedExtensionDirs,
   }
@@ -166,9 +170,10 @@ function readManagedResourceManifest(agentDir: string): ManagedResourceManifest 
  * triggers a full resync in `initResources`. The old path+size approach
  * silently cached stale prompts across upgrades.
  *
- * Cost is ~1-2ms for a typical resources tree (~100 small .md files) —
- * still negligible at startup. Files are streamed via `readFileSync` but
- * bundled prompts are tiny so this is fine.
+ * Measured ~25-30ms on a 2025 laptop for the full shipped tree (~1500 files,
+ * 17MB) — larger than early estimates but still well under the ~128ms
+ * synchronous cpSync it gates at startup. Files are streamed via
+ * `readFileSync` but bundled resources are small so this is fine.
  *
  * Exported for unit tests and for callers that want to check a different
  * directory (e.g. pre-install verification).
@@ -178,18 +183,6 @@ export function computeResourceFingerprint(rootDir: string = resourcesDir): stri
   collectFileEntries(rootDir, rootDir, entries)
   entries.sort()
   return createHash('sha256').update(entries.join('\n')).digest('hex').slice(0, 16)
-}
-
-function getCurrentResourceFingerprint(): string {
-  try {
-    const precomputed = readFileSync(join(resourcesDir, resourceFingerprintFileName), 'utf-8').trim()
-    if (/^[a-f0-9]{16}$/i.test(precomputed)) {
-      return precomputed
-    }
-  } catch {
-    // Source-tree and partial-build workflows may not have a precomputed hash.
-  }
-  return computeResourceFingerprint()
 }
 
 function resolveGsdBrowserPackageSkillPath(): string | null {
@@ -205,6 +198,16 @@ function resolveGsdBrowserPackageSkillPath(): string | null {
 
 export function setGsdBrowserPackageSkillPathForTests(skillPath: string | null | undefined): void {
   gsdBrowserPackageSkillPathForTests = skillPath
+}
+
+/**
+ * Points the resource loader at a temp bundle for tests, so the shipped-hash
+ * scenario (#2106) can be reproduced without touching the real resources tree.
+ * Pass undefined/null to restore the default resolution.
+ */
+export function setBundledResourcesDirForTests(dir: string | null | undefined): void {
+  resourcesDir = dir ?? resolveBundledResourcesDirFromPackageRoot(packageRoot)
+  bundledExtensionsDir = join(resourcesDir, 'extensions')
 }
 
 export function collectGsdBrowserPackageSkillReferences(content: string): string[] {
@@ -316,7 +319,10 @@ function collectFileEntries(dir: string, root: string, out: string[]): void {
     if (entry.isDirectory()) {
       collectFileEntries(fullPath, root, out)
     } else {
-      const rel = relative(root, fullPath)
+      // Normalize separators to match scripts/copy-resources.cjs and
+      // scripts/watch-resources.js, so the live hash equals the shipped
+      // precomputed hash on Windows too (#2106).
+      const rel = relative(root, fullPath).replaceAll('\\', '/')
       // Hash the file contents — see function doc for #4787 rationale.
       let contentHash: string
       try {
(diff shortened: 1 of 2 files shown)


jeremymcs:
Note: the `ci-gate` red is `windows-portability` failing in `packages/mcp-server` (`readProjectProgressViaBridge` recreating a deleted `gsd.db`) — the pre-existing #2158 failure from #2105's progress-read seam, tracked there with full evidence. This diff touches `src/resource-loader.ts`, which has no import path into that test, and the Linux `build` gate (14k+ tests) is green. Unrelated to this change.

PR #2182, opened 2026-09-06
test(diag): probe Windows open-boundary deletion (#2158)

jeremymcs:
Diagnostic-only, do not merge. Logs whether the recreate-boundary test's deletion hook succeeds on the windows-portability runner, discriminating #2158 hypothesis 1 (rmSync EBUSY from a leaked handle — file never deleted) vs hypothesis 2 (a second recreation path). Output lands in the windows-portability job log; the branch is deleted after evidence collection.


jeremymcs:
Evidence collected — closing the diagnostic branch. Windows logs on the final run confirmed the root cause and the fix: pre-fix `hook DELETE FAILED` (EBUSY-style failure under the identity lock), post-fix `hook deleted ... existsSync=false` → `after read: result=null existsSync=false` with the full suite green. The clean fix is #2183.

Issue #2267, opened 2026-09-11
Regression of #1593: manual blocker route omits recoveryActionId and leaves task recovery non-actionable


robert1112:
## Summary

#1593 was fixed by #1597 and is included in released versions from v1.13.0 onward. However, on v1.19.0 the manual blocker path still leaves recovery non-actionable:

```text
gsd_task_complete({ blockerDiscovered: true })
  -> { attemptId, resultId, nextStage: "route" }
  -> "Recorded blocker for task T18; awaiting recovery routing."
```

The response does not include the `recoveryActionId` required by `gsd_task_recovery_resume`.

## Reproduction

1. Run an `execute-task` unit and finish it with `gsd_task_complete(..., blockerDiscovered: true)`.
2. Observe that the successful tool result returns `attemptId`, `resultId`, and `nextStage: "route"`, but no `recoveryActionId`.
3. Attempt recovery through the documented `gsd_task_recovery_resume` API.
4. There is no supported read tool that resolves the required recovery action ID from the task or returned attempt ID. Passing the returned `attemptId` produces:

```text
Task recovery resume rejected by recovery-action-exists guard:
no Recovery Action exists with this recoveryActionId
```

5. No successor canonical attempt can be created or completed until recovery is authorized, so the task remains blocked.

## Expected

The manual blocker route should provide an actionable continuation without requiring direct database access or guessing opaque IDs. One of the following should hold:

- Return the exact `recoveryActionId` in the blocker result details and routing prompt; or
- provide a supported read tool resolving the current recovery action by milestone/slice/task or attempt; or
- have the host reliably route and authorize recovery, then dispatch the successor attempt without exposing this identifier to the worker.

The journal should make the complete progression observable:

```text
blocker -> route/recovery -> authorized successor -> execute-task -> closeout
```

## Why this is distinct from #1593

#1593/#1597 addressed the verify-timeout abort route. This report is the `gsd_task_complete({ blockerDiscovered: true })` manual blocker route, observed on the current v1.19.0 release. Its output contract still omits the identifier required by the public recovery-resume contract.

## Acceptance criteria

- Add an end-to-end regression test for manual blocker -> repair -> recovery authorization -> successor attempt -> completion.
- Workers never need to guess or query the database for a `recoveryActionId`.
- A failed resume error identifies a supported next action rather than only rejecting an unknown opaque ID.

jeremymcs:
[...]

Issue #2148, opened 2026-09-04
execute-task blockerDiscovered staging deadlocks host verification: runPostUnitVerification has no blocker-discovered branch (1.17.0)


DragonSigh:
# execute-task blockerDiscovered staging deadlocks host verification: runPostUnitVerification has no blocker-discovered branch (1.17.0)

## Problem

When an execute-task unit honestly reports a blocker via `gsd_task_complete` with `blockerDiscovered: true` (the documented sanctioned exit for tasks whose DONE gate cannot be met — see #1693/#1787), auto-mode deadlocks instead of routing the blocker: post-unit host verification throws `Host verification requires the latest succeeded canonical Attempt at the verify stage`, the orchestrator abandons and re-dispatches, the re-dispatch hits the recovery guard, the next session re-stages the identical honest blocker, and the ADR-047 liveness backstop trips at recurrence 2 ("iteration-error recurred 2x with unchanged inputs") with no in-engine exit. Every occurrence burns two blind full unit re-executions before a human intervenes.

## Root Cause

A four-leg contradiction in the durable execute-task path (line numbers from the compiled extension tree, v1.17.0):

**Leg 1 — the tool deliberately stages blockers as failed results.**
`extensions/gsd/task-completion-compatibility-adapter.js:233-238` (`stageTaskCompletion`):
```js
const blocked = input.completion.blockerDiscovered;
const settlement = settleTaskAttempt({
    invocation: input.invocation,
    attemptId,
    outcome: blocked ? "failed" : "succeeded",
    failureClass: blocked ? "blocker-discovered" : "none",
```
and `:260-263` suppresses the SUMMARY projection by design (#1726). So after an honest blocker report, `readLatestTaskAttempt` returns `state="settled", outcome="failed", nextStage="route", resultFailureClass="blocker-discovered"`, and no SUMMARY exists.

**Leg 2 — the cutover deliberately passes blockers through as "completed".**
`extensions/gsd/auto/task-execution-cutover.js:268-272` defines the predicate:
```js
function isNewlyRecordedBlocker(attempt) {
    return attempt?.state === "settled" &&
        attempt.outcome === "failed" &&
        attempt.nextStage === "route" &&
        attempt.resultFailureClass === "blocker-discovered";
}
```
and `reconcileNext` at `:283-284` short-circuits:
```js
if (isNewlyRecordedBlocker(attempt))
    return result;
```
[...]

Issue #1973, opened 2026-08-23
[Bug]: 'Request timed out.' misclassified as permanent — attempt cancelled while session survives, gsd_task_complete permanently rejected


DragonSigh:
## Problem

A mid-unit provider request timeout (`Request timed out.`) is misclassified as a **permanent** error, so GSD pauses auto-mode and settles the running Task Attempt as failed — while the host session, which transparently retried and survived, keeps executing. When the agent then calls `gsd_task_complete` with fully verified work, the call is permanently rejected with `Canonical Task completion requires a running or replay-matched Attempt` (even with `blockerDiscovered: true`), leaving verified work stranded on disk with no DB record and no agent-side write path to recover.

## Root Cause

Four linked defects, all confirmed in source:

**1. Error classifier has no pattern for the host's timeout phrasing — `error-classifier.js:36`**

```js
const NETWORK_RE = /network|ECONNRESET|ETIMEDOUT|ECONNREFUSED|socket hang up|web ?socket|fetch failed|connection.*reset|dns|unexpected eof|stream idle timeout|partial response received|stream ended without finish_reason/i;
```

`classifyError("Request timed out.")` (called at `bootstrap/agent-end-recovery.js:541`) matches **none** of the regexes (`NETWORK_RE` covers `ETIMEDOUT` and `stream idle timeout` but not `timed out`), so it falls through to `{ kind: "unknown" }` → `isTransient()` returns `false`. Note the inconsistency: `auto/run-unit.js:124` already treats `"Session creation timed out"` as `isTransient: true` — the mid-turn variant of the same condition is classified permanent.

**2. `unknown` routes to the indefinite-pause branch — `bootstrap/agent-end-recovery.js:709-718`**

```js
// --- Permanent / unknown: pause indefinitely ---
await pauseAutoForProviderError(ctx.ui, errorDetail, () => pauseAuto(ctx, pi, {
    message: `Provider error: ${errorDetail}`,
    category: "provider",
    isTransient: false,
}), { isRateLimit: false, isTransient: false, retryAfterMs: 0 });
```

A timeout — the canonical transient failure — takes the permanent path instead of the same-model in-session retry at `bootstrap/agent-end-recovery.js:655-673`.

**3. `pauseAuto` cancels the unit/attempt but leaves the live turn running — `auto.js:1644,1663`**

```js
if (options.abortActiveTurn && ctx && !ctx.isIdle())   // line 1644 — NOT passed by the provider-error path
[...]

PR #2271, opened 2026-09-11
fix(gsd): surface the recovery action id on blocker receipts

jeremymcs:
## TL;DR

**What:** a canonical blocker receipt (`gsd_task_complete` with `blockerDiscovered`) now surfaces `recoveryActionId`, `action`, and `resumeEligible` in its result details when a recovery action is recorded, and the routing notification names the id and the sanctioned `gsd_task_recovery_resume` call.
**Why:** the manual blocker route omitted the identifier the public recovery-resume contract requires, while the #1593/#1597 abort route surfaced it — leaving workers to guess IDs or read the database (#2267).
**How:** best-effort `readTaskRecoveryRoute(attemptId)` on the `nextStage: "route"` receipt, with the routing wording extracted verbatim from `latestAttemptRecoveryContext` into a shared helper; the tool description now names the field.

Addresses #2267 (does not close it — the issue's full repair → authorize → successor → completion e2e test and the failed-resume error-message rework remain open).

## What

- `workflow-tool-executors.ts`: on a blocker receipt, best-effort route lookup (try/catch → absent, never throws); `details` gains `recoveryActionId`/`action`/`resumeEligible`; the notification appends the eligibility lever. Non-blocker (`nextStage: "verify"`) results are byte-identical.
- `task-completion-compatibility-adapter.ts`: `latestAttemptRecoveryContext`'s three-case wording extracted into exported `recoveryRouteLever(route)` (single source of truth; adapter output unchanged, 44 adapter tests green).
- `db-tools.ts`: the tool's `promptSnippet`/`promptGuidelines` now name `recoveryActionId` — qualified with "when a recovery action is recorded", since a fresh blocker intentionally mints none (#2148).
- New suite: replay-after-routing receipt names the recorded action and resume call (failing-first), fresh no-route blocker keeps the plain notice, verify-stage unchanged.

## Why

The recovery action is recorded during blocker settlement, and `readTaskRecoveryRoute(attemptId)` resolves it — the data existed at receipt time but wasn't surfaced, so the documented resume API was unusable without database access. Root-cause finding worth noting: a *fresh* blocker mints no recovery action at all (auto-mode deliberately pauses rather than routing fresh blockers, #2148); the id exists in the replayed-after-routing state (#1973 surviving-worker retry), which is exactly where this fix surfaces it.

## How

Wording parity is proven byte-for-byte against the pre-extraction implementation (the adapter suite asserts the #1973 messages). Blast radius: `latestAttemptRecoveryContext` keeps its two gate-rejection call sites unchanged; `recoveryRouteLever` gains one new caller; verify-stage and non-blocker results are pinned by exact-string assertions; the tool layer passes `details` through unvalidated. Open remainder on #2267, unchanged by this patch: the full blocker → repair → authorize → successor → completion e2e chain, and reworking the failed-resume error to name a supported next action.

### DIFF
--- a/src/resources/extensions/gsd/bootstrap/db-tools.ts
+++ b/src/resources/extensions/gsd/bootstrap/db-tools.ts
@@ -1429,13 +1429,13 @@ export function registerDbTools(pi: ExtensionAPI): void {
 		description:
 			"Record a Task execution result and verification input in SQLite. Canonical Tasks advance to host verification or recovery and publish completion only after a current passing Technical Verdict; legacy Tasks complete directly and refresh readable projections.",
 		promptSnippet:
-			"Record a GSD Task result and advance verification or recovery",
+			"Record a GSD Task result and advance verification or recovery; a blocker receipt includes recoveryActionId for gsd_task_recovery_resume when a recovery action is recorded",
 		promptGuidelines: [
 			"Use gsd_task_complete (or gsd_complete_task) when a task is finished and needs to be recorded.",
 			"Include verification whenever possible. If verification is omitted, the executor derives it from verificationEvidence when possible.",
 			"verificationEvidence is an array of objects with command, exitCode, verdict, durationMs.",
 			"The tool validates required fields and returns an error message if verification cannot be derived.",
-			"Canonical success returns attemptId, resultId, nextStage, and summaryPath while completion awaits host verification; a blocker routes to recovery.",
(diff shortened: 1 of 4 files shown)


jeremymcs:
CI note for reviewers: the first build run failed on an unrelated flaky integration test — `doctor-environment-async.test.ts` "runEnvironmentChecksAsync returns the same results as runEnvironmentChecks" compares live free-disk-space strings between two sequential executions, so runner disk churn between the calls (71.3GB vs 71.2GB) broke the equality assertion. The re-run passed clean; the diff touches only the task-completion/tool-description files. A stabilization (compare keys+statuses, or normalize the free-space value) would remove the flake for every PR.

PR #2287, opened 2026-09-12
feat(hermes): notify on a stalled running session (#2210)

proofoftom:
## Linked issue

Closes #2210

- [x] I have linked an issue above. I understand that PRs without a linked issue will be closed without review.

---

## TL;DR

**What:** Adds a stall watchdog to the Hermes `SupervisorFsm` — one notification when a `running` session stops making progress.
**Why:** A finalize-retry wedge (#2046 class) looks like a healthy `running` session forever, so the operator is never told.
**How:** Track `last_change_at` on `SupervisorContext`, reset it on any state change or progress diff, and fire once past `gsd.stall_minutes` while the session is RUNNING.

## What

Six files under `integrations/hermes/` — Python only, no TS/Rust touched.

| Decision (from your 2026-09-10 approval) | Where it lives |
| --- | --- |
| Configurable `gsd:` stall key, default ~20 min | `open_gsd_hermes/config.py:24` (default), `:42` (parse), `plugin.yaml:43` (schema) |
| Progress = any state change or progress diff resets the timer; no tool-activity granularity | `open_gsd_hermes/supervisor.py:117` (`changed` flag), `:155-157` (reset), `:246-263` (`_diff_progress` now returns whether anything moved) |
| One fire-once `notify_stall` per stall, cleared on resume, carrying milestone/slice/task | `supervisor.py:49-50` (`last_change_at`, `stall_notified`), `:169` (mark), `:195-226` (`_notify_stall`), `notifications.py:67-75` |
| Optional `.gsd/stall-log.jsonl` line | `supervisor.py:212-226` — best-effort, `OSError` swallowed |
| `gsd doctor` output only if cheap | `supervisor.py:228-244` — skipped entirely when no `cli_path` is configured, `timeout=30`, `check=False`, catches `OSError`/`SubprocessError`, returns `""` on any fault. Head is 3 lines, capped at 300 chars |
| **Addition from triage: suppress the stall notification while a blocked/terminal notification is already active** | `supervisor.py:158-168` — the guard requires `ctx.state == SupervisorState.RUNNING`, with a comment naming this as the reason |

No auto-repair.

## Why

`SupervisorFsm._tick` consumed only `status.status`, `status.pending_blocker`, and the `ProgressSnapshot.active_{milestone,slice,task}` diff. Nothing tracked *when* the last change happened, so `running` with no movement was treated as healthy indefinitely — exactly the shape a wedged session presents to the Hermes operator.

## How

[...]

PR #2305, opened 2026-09-13
release: v1.20.0

jeremymcs:
Version bump for v1.20.0. Direct push to main was rejected by branch rules, so the release commit lands through the merge queue. npm, the `v1.20.0` tag and the GitHub release are already published.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,56 @@ This changelog starts from the `open-gsd/gsd-pi` ownership baseline. Earlier pro
 
 ## [Unreleased]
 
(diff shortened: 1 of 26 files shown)


jeremymcs:
The contract-tests failure is a cross-PR integration, not a release defect: #2293 made supervisor notify calls pass a per-run `target` keyword, and #2287 then added `test_supervisor_stall.py` whose `FakeNotifications` predates that signature. Repair PR #2306 widens the four fake signatures (test-only, 4 lines); merging it into main and pulling it into `release/v1.20.0` should turn contract-tests green.

PR #2306, opened 2026-09-13
test(hermes): widen stall-test notification fakes for the target parameter

jeremymcs:
## TL;DR

**What:** the `FakeNotifications` doubles in `test_supervisor_stall.py` accept the `target` keyword that supervisor notify calls now pass.
**Why:** #2293 made supervisor notify calls pass the per-run delivery target; #2287 then added `test_supervisor_stall.py` against the pre-target signature, so the v1.20.0 release branch's contract tests fail with `TypeError: ... unexpected keyword argument 'target'` on every stall/blocker/transition notification.
**How:** widen the four fake signatures with `target=None`, mirroring the widening already applied to the other notification test doubles in #2293.

## What

Four one-line signature widenings in `FakeNotifications` (`notify_stall`, `notify_transition`, `notify_blocker`, `notify_terminal`). No production code, no assertions changed.

## Why

Unblocks the v1.20.0 release PR (#2305), whose contract-tests job fails on exactly this cross-PR integration (my #2293 test-doubles widening vs #2287's new tests written against the older API).

## How

Full hermes suite green: 111 passed, 1 skipped (including #2287's stall tests). Release maintainers should merge main (or cherry-pick) into `release/v1.20.0` after this lands.

### DIFF
--- a/integrations/hermes/tests/test_supervisor_stall.py
+++ b/integrations/hermes/tests/test_supervisor_stall.py
@@ -38,16 +38,16 @@ class FakeNotifications:
         self.blockers: list[SessionStatus] = []
         self.terminals: list[tuple[str, str | None]] = []
 
-    def notify_stall(self, where: str, idle_minutes: int, doctor_head: str = "") -> None:
+    def notify_stall(self, where: str, idle_minutes: int, doctor_head: str = "", target=None) -> None:
         self.stalls.append((where, idle_minutes, doctor_head))
 
-    def notify_transition(self, message: str) -> None:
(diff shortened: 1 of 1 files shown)


jeremymcs:
Urgency update — this is no longer release-branch-only: **main's hermes contract tests are currently red** (verified on synced main: `test_supervisor_stall.py` → 4 failed with `TypeError: ... unexpected keyword argument 'target'`), and the v1.20.0 tag shipped without this fix (the release commit still has the unwidened fakes). This 4-line test-only PR is green and ready; merging it turns main's hermes workflow green again.

PR #2311, opened 2026-09-13
fix(gsd): make gate-evaluate completed-no-advance wedges clearable

### LINKED ISSUE #2310: recheckWedge never clears completed-no-advance wedges for gate-evaluate units — --resume-wedge fails forever
Follow-up to #2309, distinct root cause. #2309 explains how a `gate-evaluate` unit trips the `completed-no-advance` liveness backstop; this issue is about the *recovery path* being structurally broken for that same unit type, independent of whether #2309's dispatch bug is ever fixed.

## Repro

1. A `gate-evaluate <M>/<S>/gates+<Q...>` unit trips the liveness backstop (e.g. via #2309, or any other reason a gate-evaluate unit completes without advancing state) → wedge `W-xxxxxxxx` recorded.
2. Manually resolve the underlying issue — evaluate every gate in scope and persist verdicts via `gsd_save_gate_result` for all of them.
3. Run `/gsd auto --resume-wedge W-xxxxxxxx`.
4. Get: `Error: Cannot acknowledge wedge W-xxxxxxxx: originating guard completed-no-advance still blocks: state did not advance for gate-evaluate <M>/<S>/gates+<Q...>` — every time, regardless of how correctly the gates were evaluated.

## Root cause

`AutoOrchestrator.recheckWedge()` (`src/resources/extensions/gsd/auto/orchestrator.ts:995-1003`) only un-blocks a `completed-no-advance` wedge when `snapshotUnitTargetRows()` (`src/resources/extensions/gsd/auto-liveness-backstop.ts:317-364`) returns a hash that differs from the one stored on the wedge.

`snapshotUnitTargetRows()` builds that hash from `slices`/`tasks` table rows, and *only* additionally folds in `quality_gates`/`assessments` rows when `unitType === 'run-uat'` (lines 340-356):

```ts
} else if (slice) {
  collect('SELECT * FROM slices WHERE milestone_id = :m AND id = :s ORDER BY id', { ':m': milestone, ':s': slice });
  collect('SELECT * FROM tasks WHERE milestone_id = :m AND slice_id = :s ORDER BY id', { ':m': milestone, ':s': slice });
  if (unitType === 'run-uat') {
    collect(/* assessments ... */);
    collect(/* quality_gates WHERE gate_id = 'UAT' ... */);
  }
}
```

For `unitType === 'gate-evaluate'`, gate verdicts are **never** part of the hash. But a gate-evaluate unit's entire job is to write `quality_gates` rows — it doesn't touch `slices`/`tasks` at all. So the hash is structurally incapable of changing no matter how correctly Q3/Q4 (or any gate) get evaluated and persisted. `recheckWedge()` will report `blocking: true` forever, and `--resume-wedge` — the single sanctioned re-entry path per ADR-047 §5 — can never succeed for this unit type.

Confirmed against 1.20.0 (latest published, no update available).

## Impact

Any project that hits a `completed-no-advance` wedge on a `gate-evaluate` unit (e.g. via #2309's background-dispatch bug, or any other way that unit type stalls) is permanently stuck in the standalone `/gsd auto` CLI/TUI loop — there is no way to acknowledge the wedge short of a DB hand-edit. The MCP-driven (`gsd-workflow` server) path is unaffected since it never consults the wedge mechanism, but that's a workaround, not a fix — anyone relying on the standalone `gsd auto` loop has no recovery path at all.

## Proposed fix

Extend the `unitType === 'run-uat'` special case in `snapshotUnitTargetRows()` to also include `unitType === 'gate-evaluate'`, folding in the relevant `quality_gates` rows for the gate IDs in scope (parsed from the unit id's `gates+Q3,Q4` suffix, or simply all gate rows for that milestone/slice) so a newly-persisted verdict changes the hash and the wedge can be legitimately acknowledged.

## Reference

Observed on M033/S02 (`gate-evaluate M033/S02/gates+Q3,Q4`), immediately after manually persisting both Q3 and Q4 via `gsd_save_gate_result` per the #2309 workaround.

jeremymcs:
## TL;DR

**What:** `snapshotUnitTargetRows` now folds the slice's `quality_gates` rows into the wedge hash for `gate-evaluate` units, so persisting a gate verdict advances the hash and a `completed-no-advance` wedge becomes legitimately clearable.
**Why:** a gate-evaluate unit's target rows are quality_gates — it never touches slices/tasks — and the hash previously included them only for `run-uat`, so its wedge hash was frozen forever and the sanctioned `--resume-wedge` re-entry could never succeed (#2310).
**How:** normalize the `gates+` task segment (parseUnitId parks the gate list there, so the unit hashed a nonexistent-task query) and add a gate-evaluate special case alongside the existing run-uat one.

Closes #2310

## What

- `auto-liveness-backstop.ts`: in `snapshotUnitTargetRows`, a `gates+`-prefixed final segment normalizes to "no task" for gate-evaluate (3 lines + comment; parseUnitId itself is untouched — other callers depend on it), and a new `else if (unitType === 'gate-evaluate')` folds all `quality_gates` rows for the milestone/slice into the hash.
- `auto-liveness-backstop-1655.test.ts`: three tests — persisting a gate verdict advances the hash (failing-first: byte-identical hash before the fix), unchanged rows keep it stable, run-uat behavior pinned unchanged.

## Why

The wedge's whole premise is "hash unchanged = nothing the unit was responsible for advanced". For gate-evaluate the responsibility IS the gate rows, so excluding them made the wedge a life sentence: operators who manually persisted every verdict per the recovery instructions still got `--resume-wedge` rejected.

## How

Verified `parseUnitId` parks the gate list in `task` (the frozen hash actually came from a nonexistent-task query, not the slice branch the issue named) and that the `gates+` shape is constructed exclusively for gate-evaluate — execute-task ids keep their task-specific branch (probe-confirmed an unrelated task's change doesn't move a task-unit's hash). Known semantics: the gate-evaluate hash rides the slice branch, so it also reflects the slice's task rows — strictly more permissive for clearing, never less; and wedges tripped under the old frozen hash become acknowledgeable after upgrade, which is the intended recovery. Blast radius: all four `snapshotUnitTargetRows` callers use it symmetrically for wedge creation and recheck; run-uat byte-identical; 22/22 + 74/74 + 136/136 across the liveness/orchestrator/loop suites.


jeremymcs:
Closing as superseded: the stabilization Wave-1 PR #2316 landed its own implementation of the #2310 fix (a gate-evaluate special case in snapshotUnitTargetRows folding the slice's quality_gates rows into the wedge hash — same semantics, different shape). This PR's validation round surfaced two details worth noting for that landed code: (1) parseUnitId parks the gate list in the task slot, so pre-fix the unit hashed a nonexistent-task query (Wave-1's special case correctly targets the gates+ segment); (2) the gate-evaluate hash rides the slice branch and also reflects slice task rows — strictly more permissive for clearing, never less. #2310 is fixed on main either way.

PR #2315, opened 2026-09-13
fix(gsd): classify Anthropic extra-usage 400s as transient rate-limit

### LINKED ISSUE #2314: auto-mode pauses instead of failing over on Anthropic 400 "extra usage" error
## Summary
When the `anthropic` provider is authenticated with a Claude subscription OAuth token
(`sk-ant-oat…`, still accepted by `packages/pi-ai` at v1.20.0) and an Anthropic-backed
model returns `400 invalid_request_error` with the subscription "extra usage" message,
GSD auto-mode pauses indefinitely instead of failing over to a configured `fallbacks:`
model. The error string matches none of `classifyError()`'s patterns, so it is classed
`unknown`, and `unknown` is excluded from the fallback path — so a per-phase `fallbacks`
list that exists specifically to survive this situation is never consulted.

The exact provider message (Anthropic API):

```
400 {"type":"error","error":{"type":"invalid_request_error","message":"Third-party
apps now draw from your extra usage, not your plan limits. Add more at
claude.ai/settings/usage and keep going."}}
```

## Steps to reproduce
1. Authenticate the `anthropic` provider via subscription OAuth, then configure a
   per-phase model in `~/.gsd/PREFERENCES.md` with an Anthropic primary and a
   non-Anthropic fallback, e.g.:
   ```yaml
   models:
     execution:
       model: anthropic/claude-sonnet-5
       fallbacks:
         - openai/gpt-5        # any non-Anthropic fallback; the provider is irrelevant
   ```
2. Exhaust the Anthropic subscription's third-party "extra usage" allowance (so the
   API returns the 400 above), then run `gsd auto` / `/gsd auto` on any project.
3. When an execution unit dispatches to the Anthropic model, the 400 is returned.

## Expected behavior
GSD classifies the error as a spend/limit condition (the same bucket as
"out of extra usage" / "credits exhausted"), tries `tryProviderModelFallback()`, and
switches to the configured fallback — continuing auto-mode.
If no fallback is configured, it should pause with auto-resume (rate-limit path),
not an indefinite hard pause.

## Actual behavior
Auto-mode pauses indefinitely with:
```
Auto-mode paused due to provider error: 400 {"type":"error","error":{"type":
"invalid_request_error","message":"Third-party apps now draw from your extra usage…"}}
```
The configured fallback is never attempted.

### Root cause (observed in 1.11.0; re-traced at v1.20.0 — line numbers are v1.20.0)
- `src/resources/extensions/gsd/error-classifier.ts`
  - `RATE_LIMIT_RE` (line 55) includes `out of extra usage` but NOT the actual
    wording `from your extra usage` nor the `claude.ai/settings/usage` marker.
  - `AFFORDABILITY_RE` (line 58: `requires more credits|can only afford|insufficient
    credits|…`) also does not match.
  - So `isRateLimit` (line 140) is false; `PERMANENT_RE`, `UNSUPPORTED_MODEL_*`, and
    `MODEL_ERROR_RE` (the only 400-aware branch — requires "invalid argument/params")
    also miss → `classifyError()` returns `{ kind: "unknown" }`.
- `src/resources/extensions/gsd/bootstrap/agent-end-recovery.ts`
  - Fallback is attempted only for `rate-limit | network | server | connection |
    stream` (line 863) plus the `model-error` / `unsupported-model` / `tool-schema`
    branches. `unknown` matches none.
  - `unknown` falls to the terminal branch (line ~889, "Permanent / unknown: pause
    indefinitely") → `pauseAutoForProviderError(...)` with no auto-resume.

There is no config knob to make this error fallback-eligible; classification is
hardcoded regexes.

### Suggested fix (one line)
Route the message to the existing `rate-limit` class (semantically consistent with
the "out of extra usage" / spend-limit precedent, which already fails over then
temporarily blocks the exhausted model). In `error-classifier.ts`, broaden
`RATE_LIMIT_RE` (line 55):
```diff
- …|out of extra usage|quota (?:reached|hit)|…
+ …|(?:out of|from your) extra usage|claude\.ai\/settings\/usage|quota (?:reached|hit)|…
```
Verified locally: this matches the target message and a JSON-wrapped
`invalid_request_error` variant; it does not match benign strings like
"extra usage stats" or "usage settings"; and `PERMANENT_RE` does not shadow it, so it
cleanly reaches the rate-limit → fallback branch. (`AFFORDABILITY_RE` on line 58 is an
equally valid home if maintainers prefer to keep "credits/affordability" separate from
"throttle" semantics.)

## Related prior art
Same error-classification gap as **#1364** *"Model fallback not triggered for 'abort' /
'unknown' provider errors"* (closed via **PR #1365**), which classified the `"abort"` stop-reason
as a transient connection failure so fallbacks fire. That fix was string-specific and did **not**
cover the Anthropic subscription "extra usage" `400` — this is the sibling case on the same
`unknown` → no-fallback → indefinite-pause path. A broader structural fix (letting a configured
`fallbacks:` list be tried before an `unknown` error hard-pauses) would subsume both; the
one-line regex change above is the minimal fix for this specific, common wall.

Also related: **#1709** (Anthropic `400 cache_control.ttl` ordering error — previously
classified `unknown`, since routed) is the same "specific Anthropic 400 string → `unknown`
→ bypasses fallback" shape, fixed one string at a time.

## GSD version
Observed 1.11.0; confirmed still present by source at v1.20.0.

## Node version
(fill from `node -v` at filing time)

## OS
macOS (Darwin)

## Install method

## Affected area
auto-mode / provider error handling & model fallback

## Model
Primary `anthropic/claude-sonnet-5` (authenticated via subscription OAuth); any
non-Anthropic fallback (the fallback provider is irrelevant — the bug is upstream of
fallback selection, at classification time).

## Impact
Any user relying on a per-phase `fallbacks:` list to survive Anthropic
subscription-usage exhaustion loses the safety net: auto-mode halts and must be
manually resumed/re-pointed, defeating the "walk away" premise of auto-mode.

jeremymcs:
## TL;DR

**What:** Anthropic's subscription extra-usage 400 ("Third-party apps now draw from your extra usage… claude.ai/settings/usage") now classifies as transient rate-limit instead of `unknown`.
**Why:** the wording matched no classifier pattern, so auto-mode hard-paused indefinitely and never consulted the configured `fallbacks:` list — while the sibling wording "out of extra usage" was already treated as transient rate-limit (#2314).
**How:** extend `RATE_LIMIT_RE` and the zero-tool provider-error gate with the current wording; the settings/usage URL counts as an error signal only alongside an explicit error carrier, so prose merely linking the page can't pause auto-mode.

Closes #2314

## What

- `error-classifier.ts`: `RATE_LIMIT_RE` gains `from your extra usage` and `claude.ai/settings/usage`; comment updated. All other patterns untouched.
- `auto/unit-phase.ts`: the zero-tool provider-error path — the prefix gate gains `from your extra usage`, and the `ZERO_TOOL_PROVIDER_ERROR_SIGNAL_RE` signal set gains the settings/usage URL only when carried with `400`/`invalid_request_error` on the same line.
- `provider-errors.test.ts`: four tests — exact JSON and bare message classify as transient rate-limit via `classifyError` (failing-first); generic `invalid_request_error` 400 without the markers stays `unknown`; the zero-tool path recognizes each marker independently; prose that merely links the usage page does NOT classify.

## Why

Any user relying on per-phase `fallbacks:` to survive Anthropic subscription exhaustion lost that safety net: `unknown` is excluded from the fallback path and hard-pauses, defeating auto-mode's walk-away premise. The fix routes the condition into the existing rate-limit bucket — the same bucket the repo already uses for the older "out of extra usage" wording of the same condition.

## How

Verified end-to-end that `rate-limit` routes to `tryProviderModelFallback` (and `unknown` does not), so the classification change genuinely flips hard-pause into fallback consultation. Over-match probed adversarially: auth/billing errors that merely mention the usage URL would misroute to transient — this is the classifier's pre-existing precedence design (rate-limit over permanent) and the markers are specific to Anthropic's wording; the zero-tool gate additionally requires an error carrier alongside the URL, with a negative prose test pinning that a normal sentence linking the page does not classify. Known pre-existing edges, unchanged and noted: `RATE_LIMIT_RE` itself keeps the bare URL alternative (its input is error-text by contract), and the zero-tool path pauses with auto-resume rather than consulting fallbacks — a limitation that predates this fix for every transient class on that path (follow-up candidate).


jeremymcs:
Mergeability note: GitHub reports CONFLICTING/DIRTY, but a real `git merge main` into this branch succeeds cleanly with zero conflicts (verified against current main `154aac33c` — post-wave-8, post-#2306). The flag appears to be a stale mergeability computation; the branch merges cleanly. Close-reopen did not clear it.

jeremymcs:
@copilot resolve the merge conflicts on this branch.

jeremymcs:
Closing as superseded: the complete fix landed on main via #2316 (Wave 1, commit 7941e8c5e).

Verification against current main (`b1c706dfd`):
- `error-classifier.ts` RATE_LIMIT_RE now contains the compact equivalent `(?:out of|from your) extra usage|claude\.ai/settings/usage` — functionally identical coverage to this branch's regex (both "out of extra usage" and "from your extra usage" phrasings plus the settings/usage marker).
- `provider-errors.test.ts` on main already carries both the canonical payload test (`classifyError treats Anthropic subscription extra-usage 400 as transient rate-limit (#2314)`) and the negative prose test — 100/100 pass with `GSD_NATIVE_DISABLE=1 node --import ./src/resources/extensions/gsd/tests/resolve-ts.mjs --experimental-strip-types --test src/resources/extensions/gsd/tests/provider-errors.test.ts`.

Nothing of value is lost by dropping this branch; the rebased version only re-added duplicate tests.

